PR #19, opened 2021-10-11
Minor patch while computing residual

avik-pal:


avik-pal:
Actually I am not sure if this is the right thing to do. The main issue was that Krylov.jl uses `mul!` everywhere and if `A` is a JacVecOperator then DiffEqOperators.jl makes the following assumptions:
1. If `A * x` is called the function `f` **must** take 3 arguments
2. If `mul!(z, A, x)` is called then `f` **must** take 4 arguments.

PR #31, opened 2021-11-25
branch krylov solver call on KrylovAlg

vpuri3:
branch krylov solver call on KrylovAlg, instead of KrylovSolver (stored in cache.cacheval) as KrylovSolver is known only at runtime

### DIFF
--- a/src/wrappers.jl
+++ b/src/wrappers.jl
@@ -145,18 +145,18 @@ function SciMLBase.solve(cache::LinearCache, alg::KrylovJL; kwargs...)
     kwargs = (atol=abstol, rtol=reltol, itmax=maxiter, verbose=verbose,
               alg.kwargs...)
 
-    if cache.cacheval isa Krylov.CgSolver
+    if alg.KrylovAlg === Krylov.cg!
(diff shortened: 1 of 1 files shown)


src/wrappers.jl
```diff
@@ -145,18 +145,18 @@ function SciMLBase.solve(cache::LinearCache, alg::KrylovJL; kwargs...)
     kwargs = (atol=abstol, rtol=reltol, itmax=maxiter, verbose=verbose,
               alg.kwargs...)
 
-    if cache.cacheval isa Krylov.CgSolver
+    if alg.KrylovAlg === Krylov.cg!
```
ChrisRackauckas, line 148: isn't this computed at compile time already?

vpuri3: cache.cacheval is set at runtime in init_cacheval().

since KrylovAlg is known before and is used to set KrylovSolver, branching on KrylovAlg would be faster. what do you think?

ChrisRackauckas: what does inference say? I assume this is fully inferred at compile time.

Issue #26, opened 2021-11-24
Insert default preconditioners


ChrisRackauckas:
The default preconditioners are then composed with any preconditioner placed in there, like is done in https://github.com/SciML/DiffEqBase.jl/blob/v6.76.0/src/linear_nonlinear.jl#L219-L226 . This is shown to almost always speed up the convergence.

vpuri3:
how do we want to go about this? cache is initialized with default scalevector preconditioner and then we compose that with alg.Pl,Pr? what should we initialize ScaleVector with - norm(b) or something?

ChrisRackauckas:
Yeah, see how it's done in the other repo. ComposePreconditioner make it easy to combine preconditioners, and then this one is added directly to whatever is put in, or if nothing this is the only one that is ran. That's a strategy taken from SUNDIALS because it generally gives a pretty good convergence boost, though it might be good to have a way to disable it just in case.

Ping @YingboMa if you need help on it.

PR #33, opened 2021-11-25
default preconditioners

vpuri3:
TODO

- [x] scaling preconditioner, compose preconditioner
- [x] add warnings to wrappers, flag for disabling default preconditioner
- [x] write tests

Fix https://github.com/SciML/LinearSolve.jl/issues/26

### DIFF
--- a/src/common.jl
+++ b/src/common.jl
@@ -60,19 +60,19 @@ function SciMLBase.init(prob::LinearProblem, alg::Union{SciMLLinearSolveAlgorith
                         reltol=√eps(eltype(prob.A)),
                         maxiters=length(prob.b),
                         verbose=false,
-                        scale=one(eltype(prob.A)),
+                        Pl = nothing,
+                        Pr = nothing,
                         kwargs...,
                        )
     @unpack A, b, u0, p = prob
 
-    u0 = (u0 === nothing) ? zero(b) : u0
+    u0 = (u0 !== nothing) ? u0 : zero(b)   
+    Pl = (Pl !== nothing) ? Pl : Identity()
+    Pr = (Pr !== nothing) ? Pr : Identity()
 
     cacheval = init_cacheval(alg, A, b, u0)
     isfresh = cacheval === nothing
     Tc = isfresh ? Any : typeof(cacheval)
-
-    Pl = scaling_preconditioner(scale, true)
-    Pr = scaling_preconditioner(scale, false)
 
     A = alias_A ? A : deepcopy(A)
     b = alias_b ? b : deepcopy(b)
--- a/src/common.jl
+++ b/src/common.jl
@@ -60,7 +60,7 @@ function SciMLBase.init(prob::LinearProblem, alg::Union{SciMLLinearSolveAlgorith
                         reltol=√eps(eltype(prob.A)),
                         maxiters=length(prob.b),
                         verbose=false,
-                        preconditioner_scale=one(eltype(prob.A)),
+                        scale=one(eltype(prob.A)),
                         kwargs...,
                        )
     @unpack A, b, u0, p = prob
@@ -71,8 +71,8 @@ function SciMLBase.init(prob::LinearProblem, alg::Union{SciMLLinearSolveAlgorith
     isfresh = cacheval === nothing
     Tc = isfresh ? Any : typeof(cacheval)
 
-    Pl = default_preconditioner(preconditioner_scale, true)
-    Pr = default_preconditioner(preconditioner_scale, false)
+    Pl = scaling_preconditioner(scale, true)
+    Pr = scaling_preconditioner(scale, false)
 
     A = alias_A ? A : deepcopy(A)
     b = alias_b ? b : deepcopy(b)
--- a/src/common.jl
+++ b/src/common.jl
@@ -60,6 +60,7 @@ function SciMLBase.init(prob::LinearProblem, alg::Union{SciMLLinearSolveAlgorith
                         reltol=√eps(eltype(prob.A)),
                         maxiters=length(prob.b),
                         verbose=false,
+                        preconditioner_scale=one(eltype(prob.A)),
                         kwargs...,
                        )
     @unpack A, b, u0, p = prob
@@ -70,8 +71,8 @@ function SciMLBase.init(prob::LinearProblem, alg::Union{SciMLLinearSolveAlgorith
     isfresh = cacheval === nothing
     Tc = isfresh ? Any : typeof(cacheval)
 
-    Pl = default_preconditioner(one(eltype(A)), true)
-    Pr = default_preconditioner(one(eltype(A)), false)
+    Pl = default_preconditioner(preconditioner_scale, true)
+    Pr = default_preconditioner(preconditioner_scale, false)
 
     A = alias_A ? A : deepcopy(A)
     b = alias_b ? b : deepcopy(b)
--- a/src/common.jl
+++ b/src/common.jl
@@ -70,8 +70,8 @@ function SciMLBase.init(prob::LinearProblem, alg::Union{SciMLLinearSolveAlgorith
     isfresh = cacheval === nothing
     Tc = isfresh ? Any : typeof(cacheval)
 
-    Pl = ScaleVector(one(eltype(A)), true)
-    Pr = ScaleVector(one(eltype(A)), false)
+    Pl = default_preconditioner(one(eltype(A)), true)
+    Pr = default_preconditioner(one(eltype(A)), false)
 
     A = alias_A ? A : deepcopy(A)
     b = alias_b ? b : deepcopy(b)
--- a/src/common.jl
+++ b/src/common.jl
@@ -44,6 +44,12 @@ function set_cacheval(cache::LinearCache, alg_cache)
     return cache
 end
 
+function set_prec(cache, Pl, Pr)
+    @set! cache.Pl = Pl
+    @set! cache.Pr = Pr
+    return cache
+end
+
 init_cacheval(alg::Union{SciMLLinearSolveAlgorithm,Nothing}, A, b, u) = nothing
 
 SciMLBase.init(prob::LinearProblem, args...; kwargs...) = SciMLBase.init(prob,nothing,args...;kwargs...)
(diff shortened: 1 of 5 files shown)


vpuri3:
thoughts on broadcast vs fastbroadcast for scalevector?

```julia
julia> using FastBroadcast, BenchmarkTools

julia> n = 100; x = rand(n); s=rand(n); f(x,s) = @. x * s; fastf(x,s) = @.. x * s;

julia> @btime f(x,s);
  119.609 ns (1 allocation: 896 bytes)

julia> @btime fastf(x,s);
  124.664 ns (1 allocation: 896 bytes)
```

src/wrappers.jl
```diff
@@ -1,32 +1,90 @@
...
+    else
+        @. x = x * A.s
+    end
+end
+
+# A B α + C β
+function LinearAlgebra.mul!(C, A::ScaleVector, B, α, β)
+    
+    tmp = zero(B)
+    C = β * C + α * mul!(tmp, A, B)
```
ChrisRackauckas, line 29: This is both allocating and shadows C so it's incorrect. Split the operations to first use the tmp and then mutate C.

vpuri3: would this be better?

```julia
# A B α + C β
function LinearAlgebra.mul!(C, A::ScaleVector, B, α, β)
    A.s == one(eltype(A.s)) && return @. C = α * B + β * C

    if A.isleft
        @. C = β * C + (α / A.s) * B
    else
        @. C = β * C + (α * A.s) * B
    end
end
```

vpuri3: i think this would be faster:
```julia
# A B α + C β
function LinearAlgebra.mul!(C, A::ScaleVector, B, α, β)
    A.s == one(eltype(A.s)) && return @. C = α * B + β * C

    s = A.isleft ? 1/A.s : A.s
    mul!(C, s, B, α, β)
end 
```

ChrisRackauckas:
>thoughts on broadcast vs fastbroadcast for scalevector?

It doesn't make sense for this case. There aren't enough operations.

vpuri3:
TODO

1. write tests
2. add other operations on ScaleVector, ComposeVector for Krylov.jl
3. add warnings, flag for default preconditioner

vpuri3:
random thought: instead of writing a struct for `ScaleVector`, can we just

```julia
scalevector(s, isleft) = isleft ? I * s : I * (1/s)
```

src/wrappers.jl
```diff
@@ -1,32 +1,80 @@
...
 end
 
-function LinearAlgebra.ldiv!(v::ScaleVector, x)
+Base.eltype(A::ComposePreconditioner) = Float64 #eltype(A.inner)
+Base.adjoint(A::ComposePreconditioner) = ComposePreconditioner(A.outer', A.inner')
+Base.inv(A::ComposePreconditioner) = InvComposePreconditioner(A)
+
+function LinearAlgebra.mul!(y, A::ComposePreconditioner, x)
+    @unpack inner, outer = A
+    tmp = similar(y)
```
ChrisRackauckas, line 21: don't allocate in the `mul!` implementation. How is this avoided in the other one?

src/wrappers.jl
```diff
@@ -1,32 +1,80 @@
...
+    P::Tp
 end
 
-function LinearAlgebra.ldiv!(v::ComposePreconditioner, x)
-    @unpack inner, outer, isleft = v
+InvComposePreconditioner(inner, outer) = InvComposePreconditioner(ComposePreconditioner(inner, outer))
+
+Base.eltype(A::InvComposePreconditioner) = Base.eltype(A.P)
+Base.adjoint(A::InvComposePreconditioner) = InvComposePreconditioner(A.P')
+Base.inv(A::InvComposePreconditioner) = deepcopy(A.P)
```
ChrisRackauckas, line 55: what's this for?

src/wrappers.jl
```diff
@@ -132,6 +180,9 @@ function SciMLBase.solve(cache::LinearCache, alg::KrylovJL; kwargs...)
         cache = set_cacheval(cache, solver)
     end
 
+    M = InvComposePreconditioner(alg.Pl, cache.Pl) # left precond
+    N = InvComposePreconditioner(alg.Pr, cache.Pr) # right
```
ChrisRackauckas, line 184: use informative names please.

vpuri3:
So for IterativeSolvers.jl, we pass preconditioner `P` with inverse op described by `ldiv!`. But for Krylov.jl, we pass in the inverse of P, `Pinv` with op described by `mul!`. That's why I wrote a lazy inverse to `ComposePreconditioner`, `InvComposePreconditioner`

```julia
struct InvComposePreconditioner{Tp <: ComposePreconditioner}
    P::Tp
end
```

who's `mul!` just calls the `ldiv!` implementation of the `ComposePreconditioner`:

```julia
function LinearAlgebra.mul!(y, A::InvComposePreconditioner, x)
    @unpack P = A
    ldiv!(y, P, x)
end
```

```julia
function LinearAlgebra.ldiv!(A::ComposePreconditioner, x)
    @unpack inner, outer = A

    ldiv!(inner, x)
    ldiv!(outer, x)
end
```

So any preconditioner passed in by the user only needs to have methods `ldiv!(P, y)`, and `ldiv!(y,A,x)` defined. 

Here are the necessary methods for both types:

`ComposePreconditioner`: 2 and 3 arg `ldiv!`, `adjoint`, `eltype` (todo)
`InvComposePreconditioner`: 3 arg `mul!`, `adjoint`, `eltype`

For now, I will comment out the unused `mul!` and `ldiv!` methods for some of which i need help with

ChrisRackauckas:
Okay yeah, if the only allocating operation is also one that isn't supposed to be used it's best to not include it so that no future changes accidentally fall back to allocating.

vpuri3:
i do need help with implementing this
```julia
struct ComposePreconditioner{Ti,To}
    inner::Ti
    outer::To
end

Base.eltype(A::ComposePreconditioner) = Float64 #eltype(A.inner)
```

ChrisRackauckas:
Does it have an eltype in general? I don't think it does. Why is this necessary? For some promotion checks?

vpuri3:
Krylov.jl checks for type consistency
https://github.com/JuliaSmoothOptimizers/Krylov.jl/blob/main/src/gmres.jl#L53-L57

ChrisRackauckas:
That should get changed. It's not quite right because a preconditioner would be fine if its eltype could be promoted, i.e. if you have a preconditioner in Float32 then promoting its values to Float64 would work, so it should be something more like `is_concrete(promote_type(eltype(P),eltype(A))`. If that were the case, then `eltype(P) == Bool` would work, because `Bool` up-promotes to any number type. For now make it `eltype(Float64)` but open an upstream issue.

amontoison:
I can change
```
MisI || (eltype(M) == T) || error("eltype(M) ≠ $T") 
NisI || (eltype(N) == T) || error("eltype(N) ≠ $T")
```
for 
```
MisI || (promote_type(eltype(M), T) == T) || error("... ≠ $T") 
NisI || (promote_type(eltype(N), T) == T) || error("... ≠ $T")
```
if you want.

I did these tests to ensure that `mul!(y, M, x)` or `mul!(y, N, x)` don't allocate.

src/wrappers.jl
```diff
@@ -142,20 +161,20 @@ function SciMLBase.solve(cache::LinearCache, alg::KrylovJL; kwargs...)
               alg.kwargs...)
 
     if cache.cacheval isa Krylov.CgSolver
-        alg.Pr != LinearAlgebra.I  &&
+        N != LinearAlgebra.I  &&
             @warn "$(alg.KrylovAlg) doesn't support right preconditioning."
```
amontoison, line 165: Symmetric Krylov methods, such as CG or MINRES are only using split preconditioning.
Instead of solving `Ax = b`, you solve `M^{1/2} A M^{1/2} x = M^{1/2} b` because you need to keep a symmetric linear system. `M` must be is SPD matrix for that reason.

vpuri3:
Hi @amontoison, yes it would make sense to allow the user to pass in preconditioners that can be promoted up to eltype(A).

amontoison:
> Hi @amontoison, yes it would make sense to allow the user to pass in preconditioners that can be promoted up to eltype(A).

@vpuri3, I released the version 0.7.9 of Krylov.jl. It fixes your issue with `eltype(P)`.

vpuri3:
@ChrisRackauckas @YingboMa good to review

vpuri3:
retriggering CI

ChrisRackauckas (approved):
LGTM. @YingboMa can you review this one ASAP, today or tomorrow?

src/common.jl
```diff
@@ -54,6 +60,7 @@ function SciMLBase.init(prob::LinearProblem, alg::Union{SciMLLinearSolveAlgorith
                         reltol=√eps(eltype(prob.A)),
                         maxiters=length(prob.b),
                         verbose=false,
+                        scale=one(eltype(prob.A)),
```
YingboMa, line 63: You should use `Identity` when there's no predictioning to avoid overhead.  https://github.com/JuliaLinearAlgebra/IterativeSolvers.jl/blob/ae01dfe228138d0db9164eb343252150f6c0bfc6/src/common.jl#L28-L32

vpuri3:
@YingboMa `IterativeSolvers.Indentity` is not recognized by other libraries like `Krylov.jl` whereas 
```julia
julia> one(Float32)I == one(Float64)I == one(Bool)I == I
true
```
is recognized everywhere. Further, I have checks in each wrapper to pass in that library's "no preconditioner argumnet" (`I` for `Krylov.jl`, `Identity` for `IterativeSolvers.jl`) so there won't be any overhead in the solve process.

https://github.com/SciML/LinearSolve.jl/blob/c6c75aeebdd93222678d857aa26c575c9ec0f5d0/src/wrappers.jl#L147-L156

if you are talking about overhead in setting up cache, i can replace that single Float64 allocation to static, or to a `Bool`?

ChrisRackauckas:
holding it in type-information and then lowering to just `one(eltype(...))` for the pieces that can't handle it would make sense.

vpuri3:
from conversation on slack: avoid checks like `Pl == I` because too slow. ref - https://github.com/JuliaSmoothOptimizers/Krylov.jl/issues/462

vpuri3:
check now @YingboMa , @ChrisRackauckas

ChrisRackauckas:
Awesome thanks!

PR #34, opened 2021-11-28
Change default to use KrylovJL_GMRES on non Array

ChrisRackauckas:
Because this is GPU-compatible. IterativeSolvers.jl does some low rank Q updating which might make it more efficient on CPU but precludes it from being used on the GPU, so if that's the case then we use IterativeSolvers for the speed but more generally fall back to Krylov.jl

### DIFF
--- a/src/default.jl
+++ b/src/default.jl
@@ -7,6 +7,9 @@ function SciMLBase.solve(cache::LinearCache, alg::Nothing,
         A = A.A
     end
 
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
@avik-pal the `nothing` choice hits the default factorization which would hit this method and automatically then "do the right thing" between CPU and GPU. More generally, I would want as much downstream defaults to just hit the default linear solver and make this robust enough to handle any details optimally.

PR #39, opened 2021-12-11
Add a return type wrapper

ChrisRackauckas:


ChrisRackauckas:
@vpuri3 do you know why this would make v1.6 fail? 😅

vpuri3:
Screenshot_20211211-212634
it's svd ! i would see if the factorization algorithm overwrites the matrix A which would cause the test A*u \approx b to not be appropriate

vpuri3:
so the in place solution is not the same as the one that is emitted by `ldiv!(y fact, b) --> y`. am i seeing this correctly?
```julia
julia> VERSION
v"1.6.4"

julia> using LinearAlgebra

julia> n=8; A=rand(n,n); u=rand(n); b=A*u; v=zero(u); F=svd(A); ldiv!(v,F,b) ≈ u, v ≈ u
(true, false)
```

vpuri3:
is this expected behavior or should this be an issue in LinearAlgebra?

ChrisRackauckas:
It looks like it was a bug that was fixed in v1.7 but should probably be backported to LTS.

vpuri3:
ok created an issue in julialang https://github.com/JuliaLang/julia/issues/43404

PR #43, opened 2021-12-13
Initial pass at storing symbolic factorization

Wimmerer:
This implementation dives a little deep into the guts of UMFPACK, but I suppose that's probably inevitable. I don't believe there's a good way to check that pattern(A1) == pattern(A2), so it just throws right now.

### DIFF
--- a/src/factorization.jl
+++ b/src/factorization.jl
@@ -36,7 +36,7 @@ end
 
 # This could be a GenericFactorization perhaps?
 Base.@kwdef struct UMFPACKFactorization <: AbstractFactorization
-    reuse_symbolic::Bool = false
+    reuse_symbolic::Bool = true
 end
 
 function init_cacheval(::UMFPACKFactorization, A, b, u)
--- a/src/factorization.jl
+++ b/src/factorization.jl
@@ -35,7 +35,8 @@ function init_cacheval(alg::LUFactorization, A, b, u)
 end
 
 # This could be a GenericFactorization perhaps?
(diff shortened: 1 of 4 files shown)


src/factorization.jl
```diff
@@ -38,7 +38,7 @@ end
 struct UMFPACKFactorization <: AbstractFactorization
```
ChrisRackauckas, line 38: can we add a bool for whether to redo the symbolic factorization?

ChrisRackauckas: Default to no redoing it.

Wimmerer: Where do you want that? In the algorithm? So ` UMFPACKFactorization(reuse_symbolic=false)`?

src/factorization.jl
```diff
@@ -35,10 +35,11 @@ function init_cacheval(alg::LUFactorization, A, b, u)
 end
 
 # This could be a GenericFactorization perhaps?
-struct UMFPACKFactorization <: AbstractFactorization
+Base.@kwdef struct UMFPACKFactorization <: AbstractFactorization
+    reuse_symbolic::Bool = false
```
ChrisRackauckas, line 39:
```suggestion
    reuse_symbolic::Bool = true
```

ChrisRackauckas (approved):

ChrisRackauckas:
CI has a shit ton on it right now, so I ran it locally and it passed.

PR #59, opened 2021-12-15
setter for scaling_preconditioner

vpuri3:


ChrisRackauckas:
We should make sure this matches a more general specification of preconditioners. See https://github.com/SciML/LinearSolve.jl/pull/62

ChrisRackauckas:
Supersceded by https://github.com/SciML/LinearSolve.jl/pull/63

PR #65, opened 2021-12-17
Update Preconditioners.md

vpuri3:
Preconditioners don't need Base.inv, and LinearAlgebra.mul! to be defined. For Krylov.jl, we pass in the lazy `InvComposePreconditioner` whose `mul!` just calls the `ldiv!` of the `ComposePreconditioner`.


ChrisRackauckas:
Doesn't that require knowing the inv of the preconditioner, or does Krylov just want to call mul! on the non-inv preconditioner (that would be bizarre?)

PR #74, opened 2021-12-19
Specialize UMFPack initialization

ChrisRackauckas:


ChrisRackauckas:
@Wimmerer could you look into the KLU one? This one might finally finish the https://github.com/SciML/OrdinaryDiffEq.jl/pull/1543 PR.

Wimmerer:
Yes, but why are we constructing the UMFPACK factorization directly? So that we don't immediately run the symbolic?

ChrisRackauckas:
Yes, we just want the initialization of the cache without doing any of the work in order to get the types right.

Wimmerer:
I'll also see if I can make a PR to SuiteSparse.jl, it seems really bad that we should be handling finalizers at this level just to defer factorization.

PR #75, opened 2021-12-19
specialize init_cacheval

Wimmerer:
Fulfills KLU side of #74


Wimmerer:
Will need to wait for stopwatch on that patch

PR #81, opened 2021-12-21
Use 3-arg `ldiv!`

simonbyrne:
Unless you need it to operate in-place, it would probably make more sense to use the 3-arg `ldiv!` everywhere

### DIFF
--- a/src/factorization.jl
+++ b/src/factorization.jl
@@ -1,11 +1,20 @@
+function _ldiv!(x, A, b)
+    # work around https://github.com/JuliaLang/julia/issues/43507
+    if @which(ldiv!(x,A,b)) == which(ldiv!,Tuple{LU{Float64, Matrix{Float64}},Vector{Float64}}) 
+        copyto!(x, b)
+        ldiv!(A, x)
+    else
+        ldiv!(x, A, b)
+    end
+end
+
+
 function SciMLBase.solve(cache::LinearCache, alg::AbstractFactorization; kwargs...)
     if cache.isfresh
         fact = do_factorization(alg, cache.A, cache.b, cache.u)
         cache = set_cacheval(cache, fact)
     end
-
-    copyto!(cache.u,cache.b)
-    y = ldiv!(cache.u, cache.cacheval, cache.b)
+    y = _ldiv!(cache.u, cache.cacheval, cache.b)
     SciMLBase.build_linear_solution(alg,y,nothing,cache)
 end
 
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
No, we cannot do this because using 3-argument ldiv breaks downstream. This is why DiffEqBase was using it. See https://github.com/JuliaLang/julia/issues/33143#issuecomment-997485855 for details. This was already changed in https://github.com/SciML/LinearSolve.jl/pull/76 and I don't think we'd change the default here given how breaking the 3-argument one is due to the GC bugs.

simonbyrne:
I seem to recall that for our Schur solve, which we treat as a factorization, couldn't generally be solved in-place. What if you defined your own 3-arg `_ldiv!` function that we could intercept?

simonbyrne:
Do you ever need to solve non-square matrices?

simonbyrne:
Okay, how about his (it's not pretty, I will concede that).

simonbyrne:
Also, there is an upstream issue: https://github.com/JuliaLang/julia/issues/43507

src/factorization.jl
```diff
@@ -1,11 +1,20 @@
+function _ldiv!(x, A, b)
+    # work around https://github.com/JuliaLang/julia/issues/43507
+    if @which(ldiv!(x,A,b)) == which(ldiv!,Tuple{LU{Float64, Matrix{Float64}},Vector{Float64}})
```
ChrisRackauckas, line 3: Does this resolve at compile time, or compile away?

simonbyrne: It does look like a runtime thing, so not ideal.

simonbyrne: we could just make it
```julia
function _ldiv!(x, A, b)
        copyto!(x, b)
        ldiv!(A, x)
end
```
and then it could be monkey-patched if need be.

ChrisRackauckas: Or just

```julia
_ldiv!(x, A, b) = ldiv!(x, A, b)
function _ldiv!(x::Vector, A::Matrix, b::Vector)
        copyto!(x, b)
        ldiv!(A, x)
end
```

?

simonbyrne: `Factorization` instead of `Matrix`, but yeah

src/factorization.jl
```diff
@@ -1,11 +1,20 @@
+function _ldiv!(x, A, b)
+    # work around https://github.com/JuliaLang/julia/issues/43507
+    if @which(ldiv!(x,A,b)) == which(ldiv!,Tuple{LU{Float64, Matrix{Float64}},Vector{Float64}}) 
+        copyto!(x, b)
+        ldiv!(A, x)
+    else
+        ldiv!(x, A, b)
+    end
+end
```
ChrisRackauckas, line 9:
```suggestion
_ldiv!(x, A, b) = ldiv!(x, A, b)
function _ldiv!(x::Vector, A::Factorization, b::Vector)
        copyto!(x, b)
        ldiv!(A, x)
end
```

ChrisRackauckas (approved):

PR #84, opened 2022-01-02
Automatic bail dispatches for compose preconditioner

ChrisRackauckas:


vpuri3:
what happened?

ChrisRackauckas:
This isn't a necessarily safe dispatch and it's not used anywhere so the interface is cut down.

Issue #101, opened 2022-01-29
Nonsquare Matrices throw Bounds Error


AlCap23:
It seems like `LinearSolve` can not handle non-square matrices. A MWE would be

```Julia
using LinearAlgebra
using LinearSolve

A = randn(8, 10)
x = randn(10)
b = A*x

ldiv!(x, svd(A), b) # works

prob = LinearProblem(A, b[:,1])
sol = solve(prob, SVDFactorization()) # Bounds error
```

This is repeatable for at least `QR` and `LU`.

Wimmerer:
Was non-square `svd` added in 1.7? I get a `DimensionMismatch` on 1.6.

Wimmerer:
The problem is (I think) https://github.com/SciML/LinearSolve.jl/blob/3ba3bc3d94ed307a16adabbbe079af8228fe79ea/src/common.jl#L90

This is not the right size.

PR #102, opened 2022-01-29
Nonsquare matrix fix

Wimmerer:
This should fix https://github.com/SciML/LinearSolve.jl/issues/101. I'm not sure if the `zero.` is necessary here, it adds an allocation but there could be some solver sensitive to garbage in the result vector?

### DIFF
--- a/src/common.jl
+++ b/src/common.jl
@@ -87,7 +87,12 @@ function SciMLBase.init(prob::LinearProblem, alg::Union{SciMLLinearSolveAlgorith
                        )
     @unpack A, b, u0, p = prob
 
-    u0 = (u0 !== nothing) ? u0 : zero.(similar(b, size(A, 2)))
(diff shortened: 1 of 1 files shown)


src/common.jl
```diff
@@ -87,7 +87,7 @@ function SciMLBase.init(prob::LinearProblem, alg::Union{SciMLLinearSolveAlgorith
                        )
     @unpack A, b, u0, p = prob
 
-    u0 = (u0 !== nothing) ? u0 : zero(b)
+    u0 = (u0 !== nothing) ? u0 : zero.(similar(b, size(A, 2)))
```
ChrisRackauckas, line 90:
```suggestion
    u0 = if u0 !== nothing
        u0 
    else
        u0 = similar(b, size(A, 2))
        fill!(u0,false)
    end
```

PR #104, opened 2022-02-11
Only fill the nonzeros

ChrisRackauckas:
https://github.com/JuliaLang/julia/issues/17670


ChrisRackauckas:
Part of the https://github.com/JuliaGPU/CUDA.jl/pull/1367 fixes

Issue #123, opened 2022-04-13
CuQR is not present in CUDA.jl for julia 1.8+


avik-pal:
https://github.com/JuliaGPU/CUDA.jl/blob/f52201036272591b74b994dd7d9df1496ba1f88a/lib/cusolver/linalg.jl#L72 removes CuQR

ChrisRackauckas:
Wait, why @maleadt ? What are we supposed to use here instead in v1.8?

maleadt:
`Base.QR`. See https://github.com/JuliaLang/julia/pull/42594 for https://github.com/JuliaGPU/CUDA.jl/pull/1449 replacing https://github.com/JuliaGPU/CUDA.jl/pull/1194 to fix https://github.com/JuliaGPU/CUDA.jl/issues/1193. This whole change was because you requested support for more factorizations, so I'm surprised about you being surprised 😛

ChrisRackauckas:
Oh I see. Yes, then https://github.com/SciML/LinearSolve.jl/pull/130/files#diff-4c8f9e26491b28a6897e9cf2c7f8d59c2f4c9b5fc66e1067a4d4fd8e3fd98226 solves this issue because enough has been upstreamed that we don't need to carry around those overloads. I forgot everything got upstreamed 😅

ChrisRackauckas:
Solved in https://github.com/SciML/LinearSolve.jl/pull/130

PR #127, opened 2022-04-19
cuqr

vpuri3:
closes https://github.com/SciML/LinearSolve.jl/issues/123


ChrisRackauckas:
This doesn't solve it, because we still don't know what to do for later versions.

PR #153, opened 2022-07-13
FastLAPACK

Wimmerer:
This is just a basic set of functions I was able/willing to develop on my laptop at sea. I've only tested `A*x \approx b` for those factorizations right now. 

I'll add tests shortly. A few questions:

1. Should we default to this workspace system for QR and LU? Rather than having "FastQRFactorization"
2. Should the workspace be resized when a user uses `set_A` with a different size matrix? Problematically I triggered a segfault once when playing with the sizing of A this evening, so I'll have to track that down. It's supposed to just error (and mostly does). FastLAPACKInterface.jl doesn't provide a method to resize workspaces as far as I can tell.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -6,6 +6,7 @@ version = "1.20.0"
 [deps]
 ArrayInterfaceCore = "30b0a656-2188-435a-8636-2ec0e6a096e2"
 DocStringExtensions = "ffbed154-4ef7-542d-bbb7-c09d3a79fcae"
+FastLapackInterface = "29a986be-02c6-4525-aec4-84b980013641"
 GPUArraysCore = "46192b85-c4d5-4398-a991-12ede77f4527"
 IterativeSolvers = "42fd0dbc-a981-5370-80f2-aaf504508153"
 KLU = "ef3ab10e-7fda-4108-b977-705223b18434"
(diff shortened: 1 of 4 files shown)


ChrisRackauckas:
> Should we default to this workspace system for QR and LU? Rather than having "FastQRFactorization"

I think the high level types is probably easiest. 

> Should the workspace be resized when a user uses set_A with a different size matrix? Problematically I triggered a segfault once when playing with the sizing of A this evening, so I'll have to track that down. It's supposed to just error (and mostly does). FastLAPACKInterface.jl doesn't provide a method to resize workspaces as far as I can tell.

That would be cool if possible.

ChrisRackauckas:
Now for the real test: does it actually end up faster than direct OpenBLAS/MKL?

Wimmerer:
Well it just avoids allocations right? That's the only change AFAICT. So it should be fast for resolves

ChrisRackauckas:
Run the formatter?

ChrisRackauckas:
Actually, I'll run the formatter. I want to update the docs anyways.

Wimmerer:
Oh oops. @ChrisRackauckas just formatted and added a simple set of tests.

ChrisRackauckas:
oh that's more than fine.

PR #155, opened 2022-07-14
save ipiv for rf lu

chriselrod:


chriselrod:
The other `solve` methods don't have a `; kwargs...)`. Are we sure they're getting hit?

ChrisRackauckas:
See https://github.com/SciML/LinearSolve.jl/pull/156

PR #156, opened 2022-07-14
Add args... and kwargs... for FastFactorizations

YingboMa:


Wimmerer:
Oh I suppose I should also be doing that for other solvers like `function SciMLBase.solve(cache::LinearCache, alg::UMFPACKFactorization)`?

ChrisRackauckas:
Can we add `abstol=1e-6` or something to the `solve` call in the tests to catch this?

Wimmerer:
In `check_interface` perhaps?

ChrisRackauckas:
Okay yeah cool, it should be catching this in the tests now.

PR #157, opened 2022-07-16
update values before klu_factor

ChrisRackauckas:
Otherwise it's just using the values from before!

BTW, why do we need to klu_factor before klu!, instead of just klu!? Don't that do the factorization as well?

Fixes https://github.com/SciML/OrdinaryDiffEq.jl/issues/1701

### DIFF
--- a/src/factorization.jl
+++ b/src/factorization.jl
@@ -143,6 +143,7 @@ function SciMLBase.solve(cache::LinearCache, alg::KLUFactorization; kwargs...)
             # If we have a cacheval already, run umfpack_symbolic to ensure the symbolic factorization exists
             # This won't recompute if it does.
             KLU.klu_analyze!(cache.cacheval)
+            copyto!(cache.cacheval.nzval,A.nzval)
             if cache.cacheval._numeric === C_NULL # We MUST have a numeric factorization for reuse, unlike UMFPACK.
                 KLU.klu_factor!(cache.cacheval)
             end
--- a/src/factorization.jl
+++ b/src/factorization.jl
@@ -143,7 +143,7 @@ function SciMLBase.solve(cache::LinearCache, alg::KLUFactorization; kwargs...)
             # If we have a cacheval already, run umfpack_symbolic to ensure the symbolic factorization exists
             # This won't recompute if it does.
             KLU.klu_analyze!(cache.cacheval)
-            copyto!(cache.cacheval.nzval,A.nzval)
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
@Wimmerer do you know why after that change we'd suddenly get https://github.com/SciML/ModelingToolkit.jl/runs/7370823848?check_suite_focus=true#step:6:739 ? 

https://github.com/SciML/LinearSolve.jl/pull/156

It was never dispatching to that `solve` before.

Wimmerer:
I'm going to have to step through this code to be sure this is what we want. The `refactor` occurs in *this* line: https://github.com/SciML/LinearSolve.jl/blob/cd582d98f3334300c8f54d4d90bb287080e69261/src/factorization.jl#L150, which is also where `fact.nzval` is updated to be the same as `A`. 

Why exactly are we hitting the singular error? Is `A` being changed between `init_cacheval` and `solve`? 

First thing I want to try here is actually running `klu` in `init_cacheval` rather than just constructing the factorization. I'll make a PR to do that. That way we *know* we have a factorization to refactor with if necessary.

Wimmerer:
For UMFPACK I recall there were some changes recently which fixed some things in UMFPACK refactors. We may just disable resolves temporarily until 1.9 if I'm correct. I'll try to figure out a way to test that manually.

Wimmerer:
Or rather, the question is: is `A` being created at `init_cacheval` time with a singular matrix, then `set_A` is being used to make the actual matrix?

ChrisRackauckas:
> Or rather, the question is: is A being created at init_cacheval time with a singular matrix, then set_A is being used to make the actual matrix?

Yes, it's creating in OrdinaryDiffEq.jl with a zero matrix matching the same sparsity pattern at the initialization time, and then when solving it updates the value to the Jacobian and factorizes. So I think https://github.com/SciML/LinearSolve.jl/pull/158 would error (and we should try to capture this in tests).

> The refactor occurs in this line:

Then what does `klu_factor` do? Sounds like we can just cut that line and directly go to the `klu!` call?

ChrisRackauckas:
I'm going to merge this so downstream is fixed, but we should continue to discuss.

Wimmerer:
Okay that's the issue then. In order to refactor KLU requires an existing factorization (unlike UMFPACK AFAIK). So we need to detect that, or set some flag to ignore the first matrix. 

Or perhaps `klu!(fact, A)` should solve with `A` even if `fact` wasn't solved originally.

PR #158, opened 2022-07-16
Perform `klu` in `init_cacheval`

Wimmerer:
Let's see if this solves our problem. It simplifies things, and reaches into internal functions less. I'll investigate UMFPACK as well, I don't remember why we're reaching so deep into that package instead of just running the factorization in `init_cacheval`.


Wimmerer:
Uh what's happening with codecov? There's no way these changes should do this.

PR #161, opened 2022-07-17
Tests for zero init safety

ChrisRackauckas:


ChrisRackauckas:
The change here ended up being unnecessary (was due to https://github.com/SciML/OrdinaryDiffEq.jl/pull/1702), but the tests are good so those will get added anyways.

PR #168, opened 2022-07-28
Improve default sparse solver choice

ChrisRackauckas:
Given the benchmarks with @chriselrod and @YingboMa, it seems there is some point where a cutoff to UMFPACK makes sense. I'm placing it at 10_000 for now as a reasonable guess because we're very sure that <10_000 is bad for UMFPACK, but this should get a bit more tuning in the future. It should also depend on whether someone is using OpenBLAS or not.


Wimmerer:
Interesting. I'll work on tuning this, and maybe ask Tim if he knows of any good heuristics.

PR #169, opened 2022-08-01
Remove unnecessary doc bits after unification

ArnoStrouwen:


ArnoStrouwen:
Should the basic installation code example be in all package docs? 
Currently it is in some but not others.
I will make it uniform in whichever preferred direction.

Some packages have a home, others have an overview. I prefer overview.

ChrisRackauckas:
I think these bits should stay. Every package should have an intro IMO.

docs/src/index.md
```diff
@@ -7,9 +7,33 @@ control algorithm swapping. It also interfaces with the
...
+```
+
+## Contributing
+
+- Please refer to the
+  [SciML ColPrac: Contributor's Guide on Collaborative Practices for Community Packages](https://github.com/SciML/ColPrac/blob/master/README.md)
+  for guidance on PRs, issues, and other matters relating to contributing to ModelingToolkit.
+- There are a few community forums:
+    - the #diffeq-bridged channel in the [Julia Slack](https://julialang.org/slack/)
+    - [JuliaDiffEq](https://gitter.im/JuliaDiffEq/Lobby) on Gitter
```
ChrisRackauckas, line 30: can you use the one that I showed? It had the updated Zulip form.

docs/pages.jl
```diff
@@ -1,7 +1,6 @@
 # Put in a separate page so it can be used by SciMLDocs.jl
 
-pages = [
-    "Home" => "index.md",
```
fredrikekre, line 4: This change makes the sidebar pretty ugly since it will use `LinearSolve.jl: High-Performance Unified Linear Solvers` (the title on the `index.md` page) instead of `Home`. (Compare https://linearsolve.sciml.ai/dev/ with https://linearsolve.sciml.ai/stable/)

ChrisRackauckas: Yes. For now I think having the results show up properly in Google is worth more than a cleaner sidebar. When we can unlink the two then it will get updated.

PR #178, opened 2022-08-18
Is it a typo?

Arpit-Babbar:
I see `IterativeSolvers_GMRES` written at places in docs, while what works for me is `IterativeSolversJL_GMRES`. Is that a typo?


ChrisRackauckas:
yes typo thanks

Issue #177, opened 2022-08-17
LinearSolve default handling for non-square


alex-s-gardner:
If I want to solve a 12 parameter model:

```
A = rand(100,12)
b = rand(100,1)
x = A\b
```

works just fine in Julia or matlab. But the following does not work in LinearSolve:

```
using LinearSolve
prob = LinearProblem(A,b)
sol = solve(prob)
```
as it requires A to be a square matrix. 

I'm just curious why this is.

vpuri3:
what \ is doing is a linear least squares solve.

ChrisRackauckas:
The default algorithm choice is just forgetting to check for square before choosing lu. We should add that bit.

ChrisRackauckas:
`sol = solve(prob, QRFactorization)`

PR #187, opened 2022-08-27
Add an assumptions mechanism for type-stable default help, and nonsquare

ChrisRackauckas:
Fixes https://github.com/SciML/LinearSolve.jl/issues/177

### DIFF
--- a/src/default.jl
+++ b/src/default.jl
@@ -71,7 +71,7 @@ function defaultalg(A::GPUArraysCore.AbstractGPUArray, b::GPUArraysCore.Abstract
 end
 
 # Allows A === nothing as a stand-in for dense matrix
-function defaultalg(A, b, ::Assumptions{true})
+function defaultalg(A, b, ::OperatorAssumptions{true})
     # Special case on Arrays: avoid BLAS for RecursiveFactorization.jl when
     # it makes sense according to the benchmarks, which is dependent on
     # whether MKL or OpenBLAS is being used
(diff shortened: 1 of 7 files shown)


src/default.jl
```diff
@@ -1,209 +1,136 @@
...
+    QRFactorization()
+end
+
+## Catch high level interface
+
+function SciMLBase.solve(cache::LinearCache, alg::Nothing,
+                         args...; assumptions::OperatorAssumptions = OperatorAssumptions(),
+                         kwargs...)
+    @unpack A, b = cache
+    SciMLBase.solve(cache, default_alg(A, b, assumptions), args...; kwargs...)
```
YingboMa, line 128: Do you mean `defaultalg`?

ChrisRackauckas: yes

src/default.jl
```diff
@@ -1,209 +1,136 @@
...
+    defaultalg(A.A, b, assumptions)
+end
 
-        # This catches the cases where a factorization overload could exist
-        # For example, BlockBandedMatrix
-    elseif A !== nothing && ArrayInterfaceCore.isstructured(A)
-        alg = GenericFactorization()
+function defaultalg(A, b, ::OperatorAssumptions{nothing})
+    issquare = size(A, 1) == size(A, 2)
+    defaultalg(A, b, OperatorAssumptions(Val(issquare)))
```
YingboMa, line 12: This seems type unstable. Is that intentional, or does Julia always union split here?

ChrisRackauckas: We need to have some point where a type choice is made, because then it sets the algorithm type. Things like the ODE solver would just set the assumptions at the type level.

PR #201, opened 2022-09-23
[WIP] add warning to `KLUFactorization`

vpuri3:


vpuri3:
@Wimmerer

src/factorization.jl
```diff
@@ -308,7 +308,13 @@ end
 function init_cacheval(alg::KLUFactorization, A, b, u, Pl, Pr, maxiters::Int, abstol,
                        reltol,
                        verbose::Bool, assumptions::OperatorAssumptions)
-    return KLU.KLUFactorization(convert(AbstractMatrix, A)) # this takes care of the copy internally.
+    M = if issparse(A)
+        A
+    else
+        @warn "nonsparse operator pased to KLUFactorization(). converting to sparse"
```
ChrisRackauckas, line 314: It should just error

ChrisRackauckas (changes requested):

src/factorization.jl
Wimmerer: I agree, I'm surprised it doesn't already in KLU itself?

vpuri3: it does. but that's just breaking tests here: https://github.com/SciML/SciMLBase.jl/pull/210.

https://github.com/SciML/SciMLBase.jl/actions/runs/3109373156/jobs/5039530974

This PR is the quickest fix

vpuri3: This PR does warn the user if a dense arrays is being passed, and is a no-op otherwise. So this should be harmless

ChrisRackauckas: > but that's just breaking tests here

No, that test is doing something wrong. This is not a fix.

PR #202, opened 2022-10-03
Fix UMFPACK no symbolic factorization branch

ChrisRackauckas:
Needs tests


src/factorization.jl
```diff
@@ -292,7 +292,7 @@ function SciMLBase.solve(cache::LinearCache, alg::UMFPACKFactorization; kwargs..
             # Caches the symbolic factorization: https://github.com/JuliaLang/julia/pull/33738
             fact = lu!(cache.cacheval, A)
         else
-            fact = do_factorization(alg, A, cache.b, cache.u)
+            fact = lu!(A)
```
fredrikekre, line 295:
```suggestion
            fact = lu(A)
```
I guess? There is no method `lu!(::SparseMatrixCSC)`

ChrisRackauckas: Good point yes

PR #207, opened 2022-10-08
Simple quick fix for refactor issue

Wimmerer:
@ChrisRackauckas acceptable? I'll add a check option to UMFPACK for 1.9, and figure out how I want to do this in KLU. But this should suffice for now, we can always set `check = false` default later once those checks are integrated upstream.

### DIFF
--- a/src/factorization.jl
+++ b/src/factorization.jl
@@ -267,6 +267,7 @@ end
 
 Base.@kwdef struct UMFPACKFactorization <: AbstractFactorization
     reuse_symbolic::Bool = true
+    check::Bool = true # Check factorization re-use
 end
 
 function init_cacheval(alg::UMFPACKFactorization, A, b, u, Pl, Pr, maxiters::Int, abstol,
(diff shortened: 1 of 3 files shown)


Wimmerer:
Okay I'm not sure I understand how switching a test from `@test_throws` to a legitimate test is decreasing tcov.

src/factorization.jl
```diff
@@ -303,6 +310,7 @@ end
 
 Base.@kwdef struct KLUFactorization <: AbstractFactorization
     reuse_symbolic::Bool = true
+    check::Bool = true
```
ChrisRackauckas, line 313:
```suggestion
    check::Bool = false
```

We don't want to check by default, since downstream algorithms want retcodes not errors.

src/factorization.jl
```diff
@@ -267,6 +267,7 @@ end
 
 Base.@kwdef struct UMFPACKFactorization <: AbstractFactorization
     reuse_symbolic::Bool = true
+    check::Bool = true # Check factorization re-use
```
ChrisRackauckas, line 270:
```suggestion
    check::Bool = false # Check factorization re-use
```

src/factorization.jl
ChrisRackauckas: oh wait, this is different from singularity checking? Maybe `check_pattern` then?

Wimmerer: This is checking that pattern is the same yes.

Wimmerer:
What exactly do you want KLU and UMFPACK to do on a SingularException?

Wimmerer:
Am I good to merge this, we can solve the SingularException question in a different PR?

ChrisRackauckas:
Yes, I want to solve the other issue with retcodes, but we need enum retcodes first https://github.com/SciML/SciMLBase.jl/pull/274

PR #214, opened 2022-10-30
prototype reproducibility docs

ArnoStrouwen:


ChrisRackauckas:
Is there any way we can give someone the manifest to download?

fredrikekre:
Why not just commit `docs/Manifest.toml`?

ChrisRackauckas:
Because then it doesn't update and it's not a good doc test.

ArnoStrouwen:
If you don't want to commit the manifest, we can maybe copy it to assets in `make.jl` and then link it like:
https://github.com/SciML/LinearSolve.jl/tree/gh-pages/v1.19.0/assets
which will at least work for all the stable docs.

You can go further and for dev link things like:
https://github.com/SciML/LinearSolve.jl/tree/ceafea5e1291d6902c781abf43bc877cd6c16d42/dev/assets

ChrisRackauckas:
I think putting it in the assets is a nice idea.

PR #221, opened 2022-11-04
switch to `SciMLBase`s version of `_unwrap_val`

oscardssmith:
part of the migration in https://github.com/SciML/NonlinearSolve.jl/pull/94 and https://github.com/SciML/OrdinaryDiffEq.jl/pull/1790.


YingboMa (approved):

devmotion:
Maybe it would be less breaking to not change the type parameter but only change `_unwrap_val` and the constructor of `OperatorAssumptions` where it is used?

YingboMa:
Ah, yeah, I guess we can always normalize `OperatorAssumptions{nothing}` to `OperatorAssumptions{Nothing}` in construction.

oscardssmith:
Is that better? If so, I can switch it.

YingboMa:
Yeah, then it won't be breaking at all.

PR #232, opened 2022-11-19
correct mistake in constructor of KrylovKitJL

rveltz:
This tries to remove the error on KrylovKit iterative solver. 

I am not sure why but this does not work because of a Rosetta issue on my computer.

Basically, the way to trigger CG is to pass `isposdef = true` to KrylovKit.


rveltz:
this passes on my computer, let's cross fingers

rveltz:
I would say it worked https://github.com/rveltz/LinearSolve.jl/actions/runs/3516639502/jobs/5893454606

ChrisRackauckas:
Formatter.

PR #234, opened 2022-11-25
Update the wrapper of Krylov.jl

amontoison:
I released Krylov.jl v0.9 yesterday.
I fixed some issues on GPUs, updated stopping tolerances and added the documentation of all keyword arguments.


src/iterative_wrappers.jl
```diff
@@ -17,6 +17,7 @@ end
 
 KrylovJL_CG(args...; kwargs...) = KrylovJL(args...; KrylovAlg = Krylov.cg!, kwargs...)
 KrylovJL_GMRES(args...; kwargs...) = KrylovJL(args...; KrylovAlg = Krylov.gmres!, kwargs...)
+KrylovJL_FGMRES(args...; kwargs...) = KrylovJL(args...; KrylovAlg = Krylov.fgmres!, kwargs...)
```
ChrisRackauckas, line 20: add to the docs?

amontoison: I think that we don't need both `KrylovJL_GMRES` and `KrylovJL_FGMRES` after all.
I just added `KrylovJL_CRAIGMR` for least-norm problems and updated the documentation.

ChrisRackauckas:
That looks like a real test failure.

vpuri3:
@amontoison unrelated, but it would be a good idea to add downstream tests for `LinearSolve.jl` in `Krylov.jl`.

amontoison:
@ChrisRackauckas Do you have a direct "fallback" for linear least-norm problems here, like a `LQFactorization`?
I will update the test asap.

@vpuri3 It's `LinearSolve.jl` that have `Krylov.Jl` as a dependency and not the contrary.
It depends what you want to interface here.
What kind of tests do you have in mind?

amontoison:
> That looks like a real test failure.

`A \ b` doesn't return the solution of the least-norm problem when `A` is sparse... :-1: 
It only returns a solution of `Ax = b` but it's not the minimum norm one.
I need to convert the matrix in a dense format to recover the good solution.

vpuri3:
> @vpuri3 It's `LinearSolve.jl` that have `Krylov.Jl` as a dependency and not the contrary. It depends what you want to interface here. What kind of tests do you have in mind?

`Krylov.jl` is upstream of `LinearSolve.jl`. I suggest we add downstream tests in `Krylov.jl` so for every PR in `Krylov.jl`, the CI would ensure that nothing breaks in `LinearSolve.jl`

amontoison:
better late than never: https://github.com/JuliaSmoothOptimizers/Krylov.jl/pull/737

PR #238, opened 2022-11-28
Use `Preferences` to toggle precompilation of code that requires GPL dependencies

### LINKED ISSUE #224: Avoid (unconditional) precompiling of code that has GPL dependecies
If a user would like to run with a non-GPL version of Julia then precompilation would probably fail because of https://github.com/SciML/LinearSolve.jl/blob/0a3d5ba2f6db5894a13ae8cce717714bdcadf2e4/src/LinearSolve.jl#L64-L68 so it would be useful to make those lines conditional on inclusion of the sparse libraries.

DilumAluthge:
Fixes #224 

We use `Preferences`. The user can toggle the behavior by specifying the `include_sparse` preference. If the user does not specify the preference, we default to the value of `Base.USE_GPL_LIBS`.


andreasnoack:
Ref https://github.com/SciML/LinearSolve.jl/issues/224

ChrisRackauckas:
What about `KLUFactorization` and `UMFPACKFactorization`? Won't those need to be avoided? Then won't the default algorithm choice need to pick some other algorithm for sparse? What algorithm would it choose, does Julia have a fallback here?

DilumAluthge:
In the case where Julia was built with `USE_GPL_LIBS=0`, I think it's fine for us to say that this package doesn't support any functionality on sparse arrays.

Wimmerer:
Nothing in SparseArrays for sure.

ChrisRackauckas:
So then what do we do with the default algorithm choice?

DilumAluthge:
We could check if `LinearSolve.INCLUDE_SPARSE` is `false`, and if it is `false`, if `A` is a sparse matrix, then `defaultalg(A, ...)` throws an error saying that sparse matrices are not supported?

ChrisRackauckas:
But then how do we make that dispatch on A?

DilumAluthge:
The `SparseMatrixCSC` type still exists in Julia even if `Base.USE_GPL_LIBS` is false. So we can still dispatch on it.

ChrisRackauckas:
And does a generic fallback implementation exist on `\`?

ChrisRackauckas:
@Wimmerer weren't you going to do a pure Julia KLU sometime?

ChrisRackauckas:
I guess we can default it to KrylovJL_GMRES

j-fu:
> And does a generic fallback implementation exist on `\`?

https://github.com/PetrKryslUCSD/Sparspak.jl

DilumAluthge:
> And does a generic fallback implementation exist on `\`?

Yeah:

```julia
julia> Base.USE_GPL_LIBS
false

julia> using SuiteSparse, SparseArrays, Random, LinearAlgebra

julia> A = sprand(4, 4, 0.3) + I
4×4 SparseMatrixCSC{Float64, Int64} with 6 stored entries:
 1.24017    ⋅    ⋅         ⋅
  ⋅        1.0   ⋅         ⋅
 0.984142   ⋅   1.0        ⋅
  ⋅         ⋅   0.824643  1.0

julia> b = rand(4)
4-element Vector{Float64}:
 0.9655062442597853
 0.15175107617153183
 0.004248022435501819
 0.4659030670094594

julia> typeof(A)
SparseMatrixCSC{Float64, Int64}

julia> typeof(b)
Vector{Float64} (alias for Array{Float64, 1})

julia> A\b
4-element Vector{Float64}:
  0.7785278622862858
  0.15175107617153183
 -0.7619340985421428
  1.0942264516593845
```

Wimmerer:
It still has to be LGPL. Yeah I'll get it out the door this week @ChrisRackauckas. It just defaults to the dense case as far as I can tell.

DilumAluthge:
> I guess we can default it to KrylovJL_GMRES

Alright, I just pushed b576e4032ef147db454a21123c875e8dfc8537e6 - if `INCLUDE_SPARSE` is `false` and `A` is a `SparseMatrixCSC`, we default to `KrylovJL_GMRES()`.

src/default.jl
```diff
@@ -51,11 +57,7 @@ function defaultalg(A::GPUArraysCore.AbstractGPUArray, b, ::OperatorAssumptions{
 end
 
 function defaultalg(A, b::GPUArraysCore.AbstractGPUArray, ::OperatorAssumptions{true})
-    if VERSION >= v"1.8-"
-        LUFactorization()
-    else
-        QRFactorization()
-    end
+    KrylovJL_GMRES()
```
ChrisRackauckas, line 60: This doesn't seem right?

DilumAluthge: Ooops.

ChrisRackauckas:
> Once glue deps arrive I can add stuff like Sparspak to LinearSolve? Or I can just go ahead and do that as a direct dep.

I think it's fine to add it as a direct dep

src/default.jl
DilumAluthge: Okay I think I've reverted that change.

j-fu:
> > Once glue deps arrive I can add stuff like Sparspak to LinearSolve? Or I can just go ahead and do that as a direct dep.
> 
> I think it's fine to add it as a direct dep

@PetrKryslUCSD

PR #240, opened 2022-12-11
doc cleanup prototype

ArnoStrouwen:
- preview cleanup CI taken from MuladdMacro.jl
- enable preview
- remove analytics

### DIFF
--- a/.github/workflows/DocsPreviewCleanup.yml
+++ b/.github/workflows/DocsPreviewCleanup.yml
@@ -1,0 +1,26 @@
+name: DocsPreviewCleanup
+
+on:
+  pull_request:
+    types: [closed]
+
+jobs:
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
Where would one see the preview?

fredrikekre:
Documenter will post a commit status with the link. It only works on PRs from branches in the same repository, because otherwise the SSH key is not available (so it won't post any preview for this PR). Here is an example: https://github.com/fredrikekre/Literate.jl/pull/195 (open the commit status, click the link under `documenter/deploy` > `Details`).

ChrisRackauckas:
I don't see that same preview link here?

ArnoStrouwen:
As @fredrikekre said, no preview links from forks.

PR #241, opened 2022-12-12
Add Sparspak

Wimmerer:
No re-solve capability immediately.

I will try to get that added upstream.


Wimmerer:
I'm going to use this code: https://github.com/j-fu/ExtendableSparse.jl/blob/master/src/SparspakCSCInterface.jl to enable updates etc.

ChrisRackauckas:
open an issue about that.

PR #253, opened 2023-01-09
Update sparspak

j-fu:


PetrKryslUCSD:
This looks good. Agreed on the precompilation.

ChrisRackauckas:
Precompiling for Float64, which is by far the most common case, is probably useful though.

src/factorization.jl
```diff
@@ -503,30 +503,28 @@ end
...
 function init_cacheval(::SparspakFactorization, A, b, u, Pl, Pr, maxiters::Int, abstol,
                        reltol,
                        verbose::Bool, assumptions::OperatorAssumptions)
     A = convert(AbstractMatrix, A)
-    p = Sparspak.Problem.Problem(size(A)...)
-    Sparspak.Problem.insparse!(p, A)
-    Sparspak.Problem.infullrhs!(p, b)
-    s = Sparspak.SparseSolver.SparseSolver(p)
-    return s
+    return sparspaklu(A)
```
ChrisRackauckas, line 514: this performs an extra lu-factorization though? Can we avoid this and just build a blank/empty LU object to fill later?

j-fu: I see. Assuming I have all info on A this is possible but not in the API. I think I could  add a kwarg to sparspaklu.

ChrisRackauckas: I think a kwarg would be a nice solution.

ChrisRackauckas (approved):

ChrisRackauckas:
I'll run the formatter, thanks!

PR #254, opened 2023-01-12
HYPRE.jl solvers and preconditioners

### LINKED ISSUE #106: Hypre/BoomerAMG wrapper?
https://github.com/hypre-space/hypre

In particular, [BoomerAMG](https://hypre.readthedocs.io/en/latest/solvers-boomeramg.html) only needs a general sparse matrix, without extra physical knowledge.

### LINKED ISSUE #167: Hypre wrapper
https://github.com/fredrikekre/HYPRE.jl

fredrikekre:
This patch adds support to LinearSolve.jl for HYPRE.jl solvers and preconditioners. This is implemented as a package extension and thus requires Julia version 1.9 or higher.

Fixes #106, fixes #167.


ext/LinearSolveHYPRE.jl
```diff
@@ -0,0 +1,193 @@
...
+                                   verbose::Bool, assumptions::OperatorAssumptions)
+    return HYPRECache(nothing, nothing, nothing, UInt(0), nothing)
+end
+
+# Note:
+# SciMLBase.init is overloaded here instead of just LinearSolve.init_cacheval for two
+# reasons:
+# - HYPREArrays can't really be `deepcopy`d, so that is turned off by default
+# - The solution vector/initial guess u0 can't be created with
+#   fill!(similar(b, size(A, 2)), false) since HYPREArrays are not AbstractArrays.
```
fredrikekre, line 30: Could be an idea to define an internal function for this that take A and b and allocate a appropriate solution vector. (Won't help in this case due to the copy issue though).

ext/LinearSolveHYPRE.jl
```diff
@@ -0,0 +1,193 @@
...
+function LinearSolve.init_cacheval(alg::HYPREAlgorithm, A, b, u, Pl, Pr, maxiters::Int,
+                                   abstol, reltol,
+                                   verbose::Bool, assumptions::OperatorAssumptions)
+    return HYPRECache(nothing, nothing, nothing, UInt(0), nothing)
+end
+
+# Note:
+# SciMLBase.init is overloaded here instead of just LinearSolve.init_cacheval for two
+# reasons:
+# - HYPREArrays can't really be `deepcopy`d, so that is turned off by default
```
fredrikekre, line 28: Why are they copied unconditionally? And why is this copy using `deepcopy` and not `copy`? IIUC `deepcopy` shouldn't be used for things like this.

It seems more useful to let the algorithm decide, many solvers don't modify the matrix, so it is a waste to copy it, no??

ext/LinearSolveHYPRE.jl
```diff
@@ -0,0 +1,193 @@
...
+         alg.solver isa HYPRESolver)
+        HYPRE.Init()
+    end
+
+    # Move matrix and vectors to HYPRE, if not already provided as HYPREArrays
+    hcache = cache.cacheval
+    if hcache.A === nothing || cache.isfresh
+        hcache.A = cache.A isa HYPREMatrix ? cache.A : HYPREMatrix(cache.A)
+    end
+    isfresh_b = hcache.oidb != objectid(cache.b) # isfresh for b
```
fredrikekre, line 145: This isn't so nice, but I don't see another good way to know whether `b` has been changed. Perhaps overload `set_b(::LinearCache{..., HYPRECache, ...}, b)` would be one way?

I also think the same has to be done for `u`, since there is `set_u`.

ext/LinearSolveHYPRE.jl
ChrisRackauckas: That's what `alias_A` and `alias_b` are there to handle, though we should probably make it on demand as a solver trait, i.e. if `alias_A = false` but `solver_doesnt_alias_A(alg)`, then still don't copy. We need to do safety by default since otherwise it confuses users (always better to safety by default, and then opt in to remove the allocation), we had quite a few issues opened about this before. That said, I agree that if the solver doesn't modify `A` anyways, then it doesn't need to deepcopy it. So we can keep this init overload for now and do a trait-based fix as a separate thing.

ext/LinearSolveHYPRE.jl
```diff
@@ -0,0 +1,219 @@
...
+        hcache.isfresh_A = false
+    end
+    if hcache.isfresh_b || hcache.b === nothing
+        hcache.b = cache.b isa HYPREVector ? cache.b : HYPREVector(cache.b)
+        hcache.isfresh_b = false
+    end
+    if hcache.isfresh_u || hcache.u === nothing
+        hcache.u = cache.u isa HYPREVector ? cache.u : HYPREVector(cache.u)
+        hcache.isfresh_u = false
+    end
```
ChrisRackauckas, line 176: Does this play nicely with Distributed? So for example, if someone used a `DistributedArray` `A` and `b`, will this nicely use the memory? Copy it again? Distribute the same way? I'm curious because if we could have a default solver choice on DistributedArray values then that will fix a few downstream issues.

test/hypretests_mpi.jl
```diff
@@ -0,0 +1,72 @@
+using HYPRE
```
ChrisRackauckas, line 1: Is there no way to do MPI CI?

ext/LinearSolveHYPRE.jl
fredrikekre: Yea, thats what I meant, it should only be used as a hint to the alg whether or not to copy.

ext/LinearSolveHYPRE.jl
```diff
@@ -0,0 +1,219 @@
...
+
+    # Set the options
+    HYPRE.Internals.set_options(solver, pairs(solver_options))
+
+    return solver
+end
+
+# TODO: How are args... and kwargs... supposed to be used here?
+function SciMLBase.solve(cache::LinearCache, alg::HYPREAlgorithm, args...; kwargs...)
```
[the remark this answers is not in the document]
fredrikekre: Okay, I couldn't see them being used by other algs.

test/hypretests_mpi.jl
fredrikekre: This is doing MPI CI; this file is executed here: https://github.com/SciML/LinearSolve.jl/blob/fe4cba4de23ea5656bd5d0fd02ca618f12bfa85b/test/hypretests.jl#L166-L170

ext/LinearSolveHYPRE.jl
fredrikekre: Right now it will only work if you have either set up your matrix and vector with HYPRE to begin with or if you use a `SparseMatrixCSC` or `SparseMatrixCSR`. For the latter the matrix will be duplicated inside the hypre library, and some buffers need to be allocated in order to send the data to HYPRE.

I don't think it would work with Distributed or DistributedArray since those are not MPI processes, right?

ChrisRackauckas: Makes sense. With Distributed you could make it use the MPI cluster manager, but that's not too common.

ext/LinearSolveHYPRE.jl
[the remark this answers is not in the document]
ChrisRackauckas: Interesting, it might just be old cruft to clean up.

ext/LinearSolveHYPRE.jl
ChrisRackauckas: It can probably just be a copy.

ChrisRackauckas:
I'll merge and format

ChrisRackauckas:
Thanks!

PR #257, opened 2023-01-20
[HYPRE] Query solver for final (relative) residual and number of iterations

fredrikekre:


fredrikekre:
Error is unrelated (#258).

ChrisRackauckas (approved):

PR #260, opened 2023-01-21
Doc format

ArnoStrouwen:
Should this be applied to all repos?


ChrisRackauckas:
Yeah if this is working well now then we should enable it. It was disabled because the markdown formatting was weirdly broken for a long time.

PR #263, opened 2023-01-24
steps towards iterative solvers + tests for general numbers

j-fu:
Doesn't work for Duals in the moment


j-fu:
See https://github.com/JuliaLang/julia/issues/41753, https://github.com/JuliaSmoothOptimizers/Krylov.jl/issues/646

ChrisRackauckas:
I don't think we want to do this approach. Instead, we should do an approach like https://github.com/SciML/NonlinearSolve.jl/blob/master/src/ad.jl where an overload catches the floats and just makes BLAS solve the problem in Float64s by appropriately reinterpreting. That would be faster than using the BLAS fallbacks.

ChrisRackauckas:
Dual number support needs to happen differently, but the rest is fine? Can you rebase this?

Issue #268, opened 2023-02-02
Make `issquare` extend `SciMLOperators.issquare`


vpuri3:
https://github.com/SciML/LinearSolve.jl/blob/af09c4edb5337724218b3cb7d59c084749a71227/src/common.jl#L6

PR #269, opened 2023-02-02
make issquare extend SciMLOperators.issquare

vpuri3:
close https://github.com/SciML/LinearSolve.jl/issues/268


vpuri3:
@ChrisRackauckas this is done

Project.toml
```diff
@@ -17,19 +17,14 @@ Preferences = "21216c6a-2e73-6563-6e65-726566657250"
...
 Sparspak = "e56a9233-b9d6-4f03-8d0f-1825330902ac"
 SuiteSparse = "4607b0f0-06f3-5cda-b6b1-a6196a1729e9"
 UnPack = "3a884ed6-31ef-47d7-9d2a-63182c4928ed"
 
-[weakdeps]
-HYPRE = "b5ffcf37-a2bd-41ab-a3da-4bd9bc8ad771"
-
-[extensions]
-LinearSolveHYPRE = "HYPRE"
-
```
fredrikekre, line 32: Unrelated change (and below). Better to keep it in the 1.9 sorted way.

fredrikekre (commented):
Why can't/isn't this just an internal method?

vpuri3:
> Why can't/isn't this just an internal method?

We're integrating `SciMLOperators`, which will be replace `DiffEqOperators`, `DiffEqArrayOperators`, etc. As it will be used to represent linear operators, and `LinearSolve` solves linear systems, the two packages need to support each other. `SciMLOperators` will be upstream of `SciMLBase`, and will be reexported by it. We already have `SciMLOperators.issquare` as a utility. Check: https://github.com/SciML/SciMLOperators.jl/blob/master/src/interface.jl

```julia
help?> SciMLOperators.issquare                                                         
  No documentation found.                                                              
                                                                                         SciMLOperators.issquare is a Function.                                               
                                                                                       
  # 5 methods for generic function "issquare":                                           [1] issquare(::Union{Number, LinearAlgebra.UniformScaling}) in SciMLOperators at /Use
rs/vp/.julia/dev/SciMLOperators/src/interface.jl:197                                   
  [2] issquare(::SciMLOperators.AbstractSciMLScalarOperator) in SciMLOperators at /User
s/vp/.julia/dev/SciMLOperators/src/scalar.jl:29                                        
  [3] issquare(x::MatrixOperator, args...; kwargs...) in SciMLOperators at /Users/vp/.j
ulia/packages/Lazy/9Xnd3/src/macros.jl:297                                             
  [4] issquare(L) in SciMLOperators at /Users/vp/.julia/dev/SciMLOperators/src/interfac
e.jl:189                                                                               
  [5] issquare(A...) in SciMLOperators at /Users/vp/.julia/dev/SciMLOperators/src/inter
face.jl:198                                                                            ```

So it makes sense to just extend `SciMLBase.issquare` to act on `LinearSolve.OperatorAssumptions`. There's no reason they can't be separate. But since `SciMLOperators` will be compiled with `LinearSolve` anyway, it's just easier to extend the same function than to have two different functions with slightly different names doing the same thing.

Project.toml
vpuri3: `Pkg` must've done that. Lemme fix it

vpuri3: done

vpuri3:
with https://github.com/SciML/DiffEqFlux.jl/pull/805 merged, the `DiffEqFlux` error should be fixed.

PR #270, opened 2023-02-04
Support SciMLOperators in LinearSolve

vpuri3:
Plan: https://github.com/SciML/SciMLOperators.jl/issues/142

Merge these first:
https://github.com/SciML/SciMLBase.jl/pull/377
https://github.com/SciML/LinearSolve.jl/pull/269

The changes in the PR are as follows:

1. Replaced `AbstractDiffEqOperator` with `AbstractSciMLOperators`.
2. Created `DirectLdiv! <: AbstractLinearSolveAlgorithm` so that types that have a `ldiv!` defined may skip a second factorization step and simply use their `ldiv!`. Example, when `A isa Factorization,` or a `BiDiagonal`, or a `SciMLOperator` with `has_ldiv!(A) == true`.
3. removed `InvPreconditioner`, as it is no longer needed: https://github.com/JuliaSmoothOptimizers/Krylov.jl/issues/612#issuecomment-1227391841.
4. Haven't removed `ComposePreconditioner` as it is used in `OrdinaryDiffEq`
5. Replaced `IterativeSolvers.I` with `SciMLOperators.IdentityOperator{N}()` as identity matrix placeholder. wrote static method `_isidentity_struct` for checking.
6. wrote `TODO`'s for future changes WRT plan, some `OperatorAssumptions` notes.
7. wrote tests for these changes


src/preconditioners.jl
```diff
@@ -1,4 +1,11 @@
 # Tooling Preconditioners
+#
+# TODO - update Preconditioner docs
+# TODO - replace ComposePreconditoner with ComposedOperator after
+#        ComposePreconditioner is deprecated in OrdinaryDiffEq
+
+#const ComposePreconditioner = ComposedOperator
+#@deprecate ComposePreconditioner ComposedOperator
```
ChrisRackauckas, line 8: todos in issues.

src/preconditioners.jl
```diff
@@ -21,11 +28,5 @@ function LinearAlgebra.ldiv!(y, A::ComposePreconditioner, x)
...
-struct InvPreconditioner{T}
-    P::T
-end
-
-Base.eltype(A::InvPreconditioner) = Base.eltype(A.P)
-LinearAlgebra.ldiv!(A::InvPreconditioner, x) = mul!(x, A.P, x)
-LinearAlgebra.ldiv!(y, A::InvPreconditioner, x) = mul!(y, A.P, x)
-LinearAlgebra.mul!(y, A::InvPreconditioner, x) = ldiv!(y, A.P, x)
+#const InvPreconditioner = InvertedOperator
+#@deprecate InvPreconditioner InvertedOperator
```
ChrisRackauckas, line 32: todos in issues.

test/basictests.jl
```diff
@@ -27,20 +27,20 @@ function test_interface(alg, prob1, prob2)
...
     cache = SciMLBase.init(prob1, alg; cache_kwargs...) # initialize cache
-    y = solve(cache)
-    @test A1 * y ≈ b1
+    sol = solve(cache)
+    @test A1 * sol.u ≈ b1
 
-    cache = LinearSolve.set_A(cache, copy(A2))
-    y = solve(cache; cache_kwargs...)
-    @test A2 * y ≈ b1
+    cache = LinearSolve.set_A(cache, deepcopy(A2))
```
ChrisRackauckas, line 37: why is this necessary?

ChrisRackauckas:
This looks good, I like the direct ldiv! route. Just needs some clean up and upstream merges first.

test/basictests.jl
vpuri3: because `Base.copy` is not defined for some `SciMLOperators` like `FunctionOperator`.

vpuri3:
gotta put `InvPreconditioner` back. It is used in `NonlinearSolve`.

test/basictests.jl
```diff
@@ -349,10 +349,10 @@ end
     end
 
     @testset "Solve Function" begin
-        A1 = rand(n) |> Diagonal
+        A1 = rand(n) |> Diagonal |> Array
         b1 = rand(n)
         x1 = zero(b1)
-        A2 = rand(n) |> Diagonal
+        A2 = rand(n) |> Diagonal |> Array
```
ChrisRackauckas: If it's not working on v1.6, the answer isn't to remove the test 😅

ChrisRackauckas: Just make the direct ldiv version require a newer Julia version

vpuri3: reverted this change. `lu` isn't defined for `Diagonal` in `1.6` so I just did `lu(Array(D))` haha

src/common.jl
```diff
@@ -1,11 +1,11 @@
-struct OperatorAssumptions{issquare} end
+struct OperatorAssumptions{issq} end
 function OperatorAssumptions(issquare = nothing)
-    issquare = something(_unwrap_val(issquare), Nothing)
-    OperatorAssumptions{issquare}()
+    issq = something(_unwrap_val(issquare), Nothing)
+    OperatorAssumptions{issq}()
 end
-issquare(::OperatorAssumptions{issq}) where {issq} = issq
+SciMLOperators.issquare(::OperatorAssumptions{issq}) where {issq} = issq
```
ChrisRackauckas, line 6: why is this the same function as the operators one? They don't seem to really do the same thing.

test/basictests.jl
vpuri3: In the highlighted lines, `A1, A2` need to be `Diagonal` becuase I am doing `ldiv!(_,A1, _)` in some of the tests

src/common.jl
vpuri3: `SciMLOperators.issquare(A)` says whether a matrix is square or not. `LinearSolve.issquare(assumptions)` says whether the assumption is for a square matrix or not. pretty much the same thing. If you don't think so LMK and I'll change `LinearSolve.issquare` to `LinearSolve._issquare` to avoid name conflict

vpuri3:
rerunning CI. some tests were cancelled.

src/common.jl
ChrisRackauckas: OperatorAssumptions is not an operator, so it's a little weird. I'd keep them separate and do `__issquare`

test/basictests.jl
ChrisRackauckas: bump the version required for the ldiv! test

src/common.jl
vpuri3: sure

vpuri3: done

test/basictests.jl
vpuri3: It's not necessary. I reverted the change in these lines. So `A1, A2` are `Diagonal` (as before) and the test with `ldiv!` in `@testset "Linear Solve Function"` remains the same. I am using the same `A1, A2` in test for `DirectLdiv!` but was doing `F = lu(A)` which was failing in `1.6.7` because `A isa Diagonal`. So the only change I needed to make was `F = lu(Array(A))`.

ext/LinearSolveHYPRE.jl
```diff
@@ -4,7 +4,7 @@ using HYPRE.LibHYPRE: HYPRE_Complex
 using HYPRE: HYPRE, HYPREMatrix, HYPRESolver, HYPREVector
 using IterativeSolvers: Identity
 using LinearSolve: HYPREAlgorithm, LinearCache, LinearProblem, LinearSolve,
-                   OperatorAssumptions, default_tol, init_cacheval, issquare, set_cacheval
+                   OperatorAssumptions, default_tol, init_cacheval, __issquare, set_cacheval
```
ChrisRackauckas, line 7: we need to be careful with this release.

test/basictests.jl
vpuri3: in fact, i took a second look at the test and I realized I dont need to do a factorization. So I'll remove that now

ext/LinearSolveHYPRE.jl
vpuri3: yes. good we have CI test for 1.9

vpuri3: tests passing with lts, release, beta locally.

PR #272, opened 2023-02-06
fix typo when setting Pr to default

j-fu:
Prevents these warnings:

┌ Warning: cg! doesn't support right preconditioning.
└ @ LinearSolve ~/.julia/packages/LinearSolve/fBayl/src/iterative_wrappers.jl:168


vpuri3:
thanks for catching my mistake!

ChrisRackauckas (approved):

Issue #190, opened 2022-08-31
Use AbstractSparseMatrixCSC in the API


j-fu:
This would make the packages work for other subtypes other than SparseMatrixCSC, e.g. in ExtendableSparse.jl.

Opening this for reference re our slack communication with @Wimmerer .

j-fu:
Solved with #273 being merged.

PR #273, opened 2023-02-08
Use AbstractSparseMatrixCSC for sparse factorizations

j-fu:
This e.g. would allow to supply a matrix type with multithreaded factorization etc. Would solve #190. 

However, see https://github.com/JuliaSparse/SparseArrays.jl/issues/265 .


ChrisRackauckas:
Anywhere that assumes it's a sparsematrixcsc should not remove the conversion because that could concrete things like MatrixOperators of SparseMatrixCSC. For things that are already concrete matrices it should be a no-op so that's fine, and then afterwards it should assert it's an AbstractSparseMatrixCSC and use the sparse matrix functions on it.

ChrisRackauckas (approved):

j-fu:
:+1:

ChrisRackauckas:
It should probably be changed into an extension package in the near future.

PR #295, opened 2023-04-04
Attempt to close #293

### LINKED ISSUE #293: precompilation warning
On `1.9.0-rc2`:
```julia
WARNING: Method definition init_cacheval(LinearSolve.GenericFactorization{F} where F, Any, Any, Any, Any, Any, Int64, Any, Any, Bool, LinearSolve.OperatorAssumptions{issq, condition} where condition where issq) in module LinearSolve at [...]/.julia/packages/LinearSolve/P5a5J/src/factorization.jl:313 overwritten at [...]/.julia/packages/LinearSolve/P5a5J/src/factorization.jl:326.
  ** incremental compilation may be fatally broken for this module **
```

On `[7ed4a6bd] LinearSolve v1.41.1`.

thomvet:
- Simplified dispatch in factorization.jl
- Tests pass locally
- Also brought back cholesky! tests (needed a "don't alias" fix in basictests.jl)

### DIFF
--- a/src/factorization.jl
+++ b/src/factorization.jl
@@ -312,7 +312,8 @@ end
 # Try to never use it.
 
 # Cholesky needs the posdef matrix, for GenericFactorization assume structure is needed
-function init_cacheval(alg::Union{GenericFactorization, GenericFactorization{typeof(cholesky)},
+function init_cacheval(alg::Union{GenericFactorization,
+                                  GenericFactorization{typeof(cholesky)},
                                   GenericFactorization{typeof(cholesky!)}}, A, b, u, Pl, Pr,
                        maxiters::Int, abstol, reltol, verbose::Bool,
                        assumptions::OperatorAssumptions)
--- a/test/basictests.jl
+++ b/test/basictests.jl
@@ -39,7 +39,7 @@ function test_interface(alg, prob1, prob2)
     sol = solve(cache; cache_kwargs...)
     @test A2 * sol.u ≈ b1
 
-    cache = LinearSolve.set_A(cache, deepcopy(A2))
+    cache = LinearSolve.set_A(cache, A2)
     cache = LinearSolve.set_b(cache, b2)
     sol = solve(cache; cache_kwargs...)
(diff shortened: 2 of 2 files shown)


src/factorization.jl
```diff
@@ -310,21 +310,20 @@ end
...
+#                        A, b, u, Pl, Pr, maxiters::Int, abstol, reltol, verbose::Bool,
+#                        assumptions::OperatorAssumptions)
+#     newA = copy(convert(AbstractMatrix, A))
+#     if newA isa AbstractSparseMatrix
+#         fill!(nonzeros(newA), true)
+#     else
+#         fill!(newA, true)
+#     end
+#     do_factorization(alg, newA, b, u)
+# end
```
ChrisRackauckas, line 323:
```suggestion
```

test/basictests.jl
```diff
@@ -39,6 +39,7 @@ function test_interface(alg, prob1, prob2)
     sol = solve(cache; cache_kwargs...)
     @test A2 * sol.u ≈ b1
 
+    cache = LinearSolve.set_A(cache, deepcopy(A2))
```
ChrisRackauckas, line 42: this is no longer testing that the no-aliasing works.

thomvet: I added that to make the last test work for `cholesky!`. Does that indicate that the no-aliasing behavior is not correctly defined for this algorithm? Where?

ChrisRackauckas: Seems like it. `default_alias_A(::Any, ::Any, ::Any) = false` should make it internally deepcopy. This this is a separate topic so it shouldn't be in the same PR.

ChrisRackauckas (approved):

test/basictests.jl
thomvet: Yeah, I saw that and thought that it should do the deep copy internally, but the failing test tells me that it isn't *shrug*. I reverted back to not deep-copying here and re-commented the failing test.

Issue #302, opened 2023-05-13
solving the same problem twice is broken


karajan9:
On Julia 1.9.0 with LinearSolve 1.42.0 solving the same problem twice doesn't give the same result:
```
A = [1.0 2.0; 3.0 4.0]
b = [1.0, 2.0]
prob = LinearProblem(A, b)
linsolve = init(prob)
sol1 = solve(linsolve)
# julia> sol1.u
# 2-element Vector{Float64}:
#  0.0
#  0.5
sol2 = solve(linsolve)
# julia> sol2.u
# 2-element Vector{Float64}:
#  -10.999999999999995
#    8.499999999999996
```
This is also true when `set_b` is used in between.
```
A = [1.0 2.0; 3.0 4.0]
b = [1.0, 2.0]
prob = LinearProblem(A, b)
linsolve = init(prob)
sol1 = solve(linsolve)
# julia> sol1.u
# 2-element Vector{Float64}:
#  0.0
#  0.5
linsolve = LinearSolve.set_b(linsolve, [2.0, 3.0])
sol2 = solve(linsolve)
# julia> sol2.u
# 2-element Vector{Float64}:
# -15.999999999999993
#  12.499999999999995
# should be [-1.0, 1.5]
```

That means that the examples in the README are currently broken.

DanielVandH:
Could be related to the cache being still treated as fresh after solving:

```julia
julia> A = [1.0 2.0; 3.0 4.0]
       b = [1.0, 2.0]
       prob = LinearProblem(A, b)
       linsolve = init(prob)
       solve(linsolve)
       linsolve.isfresh
true
```

karajan9:
Interestingly,
```
sol = solve(linsolve)
sol.isfresh == false
```

karajan9:
I also noticed that 
```
A2 = [1.0 3.0; 3.0 4.0]
linsolve = LinearSolve.set_A(linsolve, A2)
solve(linsolve)

julia> A2
2×2 Matrix{Float64}:
 3.0       4.0
 0.333333  1.66667
```
if you don't `deepcopy` it.

Also, basically all tests fail as soon as you replace `A` in
```
# top of `basictests.jl`
n = 8
A = Matrix(I, n, n)
b = ones(n)
A1 = rand(n, n);
b1 = rand(n);
x1 = zero(b);
A2 = A / 2;
b2 = rand(n);
x2 = zero(b);

[...]

PR #303, opened 2023-05-14
Mutable cache form

ChrisRackauckas:
I think this is required to fix https://github.com/SciML/LinearSolve.jl/issues/302 . Otherwise there is no way to set some kind of "has been factorized". The reason for the functional immutable form was to be easier on the compiler, but I don't think that has worked out in the end. It has been a bit unergonomic to have to keep finding the latest cache, and in scenarios where it's kept longer (i.e. `init` the ODE solver) it will allocate just the same, while when it's in a more constrained context (`solve` in the ODE solver) the compiler will elide the allocation for both the mutable and immutable forms these days. Also, this puts LinearSolve more in line with the rest of the SciML interface libraries.


YingboMa (approved):

ChrisRackauckas:
This is majorly breaking so it will be 2.0.

fredrikekre:
> This is majorly breaking so it will be 2.0.

Can I request to change the default behavior of copying the matrix/vector if there will be a breaking release? :)

ChrisRackauckas:
What's the requested change?

fredrikekre:
To not copy by default.

j-fu:
Hi, this seems to be an interesting improvement regarding usability of the API. I will try to test this with VoronoiFVM.jl, I should be able to find some time this week, would report back.

ChrisRackauckas:
> To not copy by default, flipping these

Is there a way to do that and still solve the problem that two solves by default give the same answer? I'm not sure there is?

fredrikekre:
Do you have an MWE for what you mean?

ChrisRackauckas:
https://github.com/SciML/LinearSolve.jl/issues/302 is something that for example I'm not sure would work.

j-fu:
Seems to work for me, simplifies things big time! I one place I had a mutable struct which just wrapped a LinearCache to carry around...


But why make this immediately breaking ? IMHO it could be possible to keep the old API via `solve` and the new one via `solve!` . May be it could be deprecated and removed  later.

This also would prevent making packages using LinearSolve 1.0 and 
LinearSolve 2.0 incompatible.

ChrisRackauckas:
I should/will go back and add deprecation paths, but it's much easier to test things by removing the syntax that shouldn't be used and then after everything works adding the deprecation paths back.

ChrisRackauckas:
@fredrikekre if you think that's solvable, please open an issue and let's discuss what to do there. I don't plan to tag until the end of the week so this can get some more testing, and a few more necessary changes in.

j-fu:
Checked for compatibility of the immutatable cache stuff - this works as well  :+1:

PR #307, opened 2023-05-29
RFC/WIP: Rework default algorithm to be fully type stable

ChrisRackauckas:
Two possible ideas. One is to put all algorithms and caches in the default algorithm into a unityper. The other is to make a "mega algorithm" with runtime information on the choice. This takes the latter approach because that keeps most of the package code intact and makes it so that any algorithm choice by the user will not have any runtime behavior.

This uses an enum inside of the algorithm struct in order to choose the actual solver for the given process. In init and solve static dispatching is done through hardcoded branches.

Todo:

- [ ] Finish the implementation to make it actually work (check the generated functions)
- [ ] Make the OperatorAssumptions be dynamic information
- [ ] Test and time to see the real overhead
- [ ] Munge the output into a solution struct that doesn't have the actual algorithm so that the output is type stable

Things to consider:

The one thing that may be a blocker is the `init` cost. While I don't think it will be an issue, we will need to see if constructing all of the empty arrays simply to hold a thing of the right type is too costly. Basically, what we may need is for `Float64[]` to be optimized to be a no-op zero allocation, in which case essentially the whole cache structure should be free to build. As it stands, this may be a cause of overhead as it needs to build all of the potential caches even if it only ever uses one.

### DIFF
--- a/src/default.jl
+++ b/src/default.jl
@@ -1,123 +1,144 @@
+struct DefaultLinearSolver <: SciMLLinearSolveAlgorithm
+    alg::DefaultAlgorithmChoice
+end
+
+EnumX.@enumx DefaultAlgorithmChoice begin
+    LUFactorization
+    QRFactorization
+    DiagonalFactorization
+    DirectLdiv!
+    SparspakFactorization
+    KLUFactorization
+    UMFPACKFactorization
+    KrylovJL_GMRES
+    GenericLUFactorization
+    RowMaximumGenericLUFactorization
+    RFLUFactorization
+    PivotedRFLUFactorization
+    LDLtFactorization
+    SVDFactorization
+end
+
 # Legacy fallback
 # For SciML algorithms already using `defaultalg`, all assume square matrix.
 defaultalg(A, b) = defaultalg(A, b, OperatorAssumptions(Val(true)))
 
 function defaultalg(A::Union{DiffEqArrayOperator, MatrixOperator}, b,
                     assumptions::OperatorAssumptions)
-    defaultalg(A.A, b, assumptions)
+    DefaultLinearSolver(defaultalg(A.A, b, assumptions))
 end
 
 # Ambiguity handling
 function defaultalg(A::Union{DiffEqArrayOperator, MatrixOperator}, b,
                     assumptions::OperatorAssumptions{nothing})
-    defaultalg(A.A, b, assumptions)
+    DefaultLinearSolver(defaultalg(A.A, b, assumptions))
 end
 
 function defaultalg(A::Union{DiffEqArrayOperator, MatrixOperator}, b,
                     assumptions::OperatorAssumptions{false})
-    defaultalg(A.A, b, assumptions)
+                    DefaultLinearSolver(defaultalg(A.A, b, assumptions))
 end
 
 function defaultalg(A::Union{DiffEqArrayOperator, MatrixOperator}, b,
                     assumptions::OperatorAssumptions{true})
-    defaultalg(A.A, b, assumptions)
+                    DefaultLinearSolver(defaultalg(A.A, b, assumptions))
 end
 
 function defaultalg(A, b, ::OperatorAssumptions{Nothing})
     issq = issquare(A)
-    defaultalg(A, b, OperatorAssumptions(Val(issq)))
+    DefaultLinearSolver(defaultalg(A, b, OperatorAssumptions(Val(issq))))
 end
 
 function defaultalg(A::Tridiagonal, b, ::OperatorAssumptions{true})
-    GenericFactorization(; fact_alg = lu!)
+    DefaultAlgorithmChoice.LUFactorization
 end
 function defaultalg(A::Tridiagonal, b, ::OperatorAssumptions{false})
-    GenericFactorization(; fact_alg = qr!)
+    DefaultAlgorithmChoice.QRFactorization
 end
 function defaultalg(A::SymTridiagonal, b, ::OperatorAssumptions{true})
-    GenericFactorization(; fact_alg = ldlt!)
+    DefaultAlgorithmChoice.LDLtFactorization
 end
 function defaultalg(A::Bidiagonal, b, ::OperatorAssumptions{true})
-    DirectLdiv!()
+    DefaultAlgorithmChoice.DirectLdiv!
 end
 function defaultalg(A::Factorization, b, ::OperatorAssumptions{true})
-    DirectLdiv!()
+    DefaultAlgorithmChoice.DirectLdiv!
 end
 function defaultalg(A::Diagonal, b, ::OperatorAssumptions{true})
-    DiagonalFactorization()
+    DefaultAlgorithmChoice.DiagonalFactorization
 end
 function defaultalg(A::Diagonal, b, ::OperatorAssumptions{Nothing})
-    DiagonalFactorization()
+    DefaultAlgorithmChoice.DiagonalFactorization
 end
 
 function defaultalg(A::AbstractSparseMatrixCSC{Tv, Ti}, b,
                     ::OperatorAssumptions{true}) where {Tv, Ti}
-    SparspakFactorization()
+    DefaultAlgorithmChoice.SparspakFactorization
 end
 
 @static if INCLUDE_SPARSE
     function defaultalg(A::AbstractSparseMatrixCSC{<:Union{Float64, ComplexF64}, Ti}, b,
                         ::OperatorAssumptions{true}) where {Ti}
         if length(b) <= 10_000
-            KLUFactorization()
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
```julia
julia> @benchmark solve($prob)
BenchmarkTools.Trial: 10000 samples with 9 evaluations.
 Range (min … max):  2.653 μs …  1.069 ms  ┊ GC (min … max):  0.00% … 74.47%
 Time  (median):     2.875 μs              ┊ GC (median):     0.00%
 Time  (mean ± σ):   4.443 μs ± 36.512 μs  ┊ GC (mean ± σ):  24.39% ±  2.96%

    ▃▅▆▇█████▇▆▆▅▅▄▃▃▂▂▂▁▁▁▁▁  ▁ ▁▁▁▁                        ▃
  ▆████████████████████████████████████████████▇█▇▇▅▇▆▇▇▅▆▅▅ █
  2.65 μs      Histogram: log(frequency) by time     4.02 μs <

 Memory estimate: 8.83 KiB, allocs estimate: 101.

julia> @benchmark solve($prob, $(LUFactorization()))
BenchmarkTools.Trial: 10000 samples with 112 evaluations.
 Range (min … max):  762.652 ns … 35.605 μs  ┊ GC (min … max): 0.00% … 96.61%
 Time  (median):     778.643 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   858.785 ns ±  1.231 μs  ┊ GC (mean ± σ):  8.18% ±  5.54%

  ▂▅▇██▇▇▅▄▃▂▂▂▂▂▁▁▁                                           ▂
  ██████████████████████▇▆▇▇▇▇█▇▇▇▇▆▇▆▆▆▆▆▆▇▇▆▆▆▆▆▆▆▆▅▅▆▆▅▅▅▃▅ █
  763 ns        Histogram: log(frequency) by time       966 ns <

 Memory estimate: 1.47 KiB, allocs estimate: 15.
```

ChrisRackauckas:
Needs https://github.com/JuliaArrays/ArrayInterface.jl/pull/415

ChrisRackauckas:
```julia
using LinearSolve

A = rand(100, 100)
b = rand(100)
prob = LinearProblem(A, b)

using BenchmarkTools
@benchmark solve($prob)
@benchmark solve($prob, $(LUFactorization()))
```

```julia
julia> @benchmark solve($prob)
BenchmarkTools.Trial: 10000 samples with 1 evaluation.
 Range (min … max):  66.166 μs …  2.565 ms  ┊ GC (min … max): 0.00% … 89.21%
 Time  (median):     72.355 μs              ┊ GC (median):    0.00%
 Time  (mean ± σ):   77.852 μs ± 88.509 μs  ┊ GC (mean ± σ):  4.89% ±  4.16%

            ▃██▅▂                                              
  ▂▄▅▄▂▂▂▁▁▄█████▆▆▆█▇▆▅▄▃▃▃▂▂▃▂▂▂▂▂▂▂▂▂▂▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁ ▂
  66.2 μs         Histogram: frequency by time        91.7 μs <

 Memory estimate: 110.23 KiB, allocs estimate: 124.

julia> @benchmark solve($prob, $(LUFactorization()))
BenchmarkTools.Trial: 10000 samples with 1 evaluation.
 Range (min … max):  107.875 μs …  6.946 ms  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     116.917 μs              ┊ GC (median):    0.00%
 Time  (mean ± σ):   121.663 μs ± 92.707 μs  ┊ GC (mean ± σ):  1.87% ± 3.81%

        ▆█▁  ▆▂                                                 
  ▁▁▃▂▂▅███▄███▅▃▃▂▂▂▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁ ▂
  108 μs          Histogram: frequency by time          163 μs <

 Memory estimate: 81.80 KiB, allocs estimate: 16.
```

The problem seems to be that we need a way to lazily init the GMRES cache

ChrisRackauckas:
```julia
using LinearSolve

A = rand(100, 100)
b = rand(100)
prob = LinearProblem(A, b)

using BenchmarkTools
@benchmark solve($prob)
@benchmark solve($prob, $(RFLUFactorization()))
```

```julia

BenchmarkTools.Trial: 10000 samples with 1 evaluation.
 Range (min … max):  65.333 μs …  2.537 ms  ┊ GC (min … max): 0.00% … 87.08%
 Time  (median):     70.583 μs              ┊ GC (median):    0.00%
 Time  (mean ± σ):   76.125 μs ± 91.330 μs  ┊ GC (mean ± σ):  4.66% ±  3.77%

   ▄▅▄▁   ▁▆▇█▇▆▅▃▄▄▅▅▄▄▃▃▃▂▂▁▁▁▁▂▂▁▁                         ▂
  ▇████▆▅▂█████████████████████████████████████▇▆▇▇▇▅▇▆▆▄▅▄▅▅ █
  65.3 μs      Histogram: log(frequency) by time      90.2 μs <

 Memory estimate: 88.84 KiB, allocs estimate: 105.

BenchmarkTools.Trial: 10000 samples with 1 evaluation.
 Range (min … max):  63.458 μs …  2.124 ms  ┊ GC (min … max): 0.00% … 95.95%
 Time  (median):     68.208 μs              ┊ GC (median):    0.00%
 Time  (mean ± σ):   72.105 μs ± 57.144 μs  ┊ GC (mean ± σ):  3.27% ±  3.97%

   ▁       ▅█▅                                                 
  ▄█▂▁▁▁▁▁▅████▄▃▃▃▄▄▄▄▃▂▂▂▂▂▂▂▂▂▂▂▂▂▁▁▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁ ▂
  63.5 μs         Histogram: frequency by time        87.5 μs <

 Memory estimate: 81.81 KiB, allocs estimate: 16.
```

reasonable scaling

ChrisRackauckas:
This is actually very good now.

```julia
using LinearSolve

A = rand(4, 4)
b = rand(4)
prob = LinearProblem(A, b)

solve(prob).alg.alg

using BenchmarkTools
@benchmark solve($prob)
@benchmark solve($prob, $(GenericLUFactorization(LinearSolve.RowMaximum())))
```

```julia
BenchmarkTools.Trial: 10000 samples with 202 evaluations.
 Range (min … max):  379.743 ns …   7.574 μs  ┊ GC (min … max): 0.00% … 93.00%
 Time  (median):     398.099 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   427.764 ns ± 306.002 ns  ┊ GC (mean ± σ):  4.29% ±  5.58%

       ▇█▄                                                       
  ▁▁▂▃████▇▆▆▅▃▃▃▃▃▃▃▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁ ▂
  380 ns           Histogram: frequency by time          516 ns <

 Memory estimate: 880 bytes, allocs estimate: 8.

BenchmarkTools.Trial: 10000 samples with 197 evaluations.
 Range (min … max):  452.198 ns …  11.249 μs  ┊ GC (min … max): 0.00% … 92.86%
 Time  (median):     463.411 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   498.177 ns ± 441.242 ns  ┊ GC (mean ± σ):  4.74% ±  5.09%

   ▃▇██▆▄▃▃▃▄▄▃▃▃▂▁▁▁▂▂▂▁▁▁▁▁   ▁                               ▂
  ▇█████████████████████████████████▇▇█▇▇▇▇█▇▆▆▄▇▇▅▆▅▄▆▅▅▄▅▅▅▄▄ █
  452 ns        Histogram: log(frequency) by time        599 ns <

 Memory estimate: 704 bytes, allocs estimate: 9.
```

Not sure why the default method is faster, but it's pretty consistently faster. And inference works:

```julia
using Test
@inferred solve(prob)
@inferred init(prob, nothing)
```

ChrisRackauckas:
Next up I want to make IterativeSolvers and KrylovKit into extensions, and then setup every solver with docstrings, and that's 2.0.

ChrisRackauckas:
I didn't know about that, but I did check Cthulhu and it worked. If you can PR to add the JET testing that would be helpful.

ChrisRackauckas:
https://github.com/SciML/LinearSolve.jl/pull/318

some work, others hit random stuff like print.

PR #309, opened 2023-05-31
Update documentation regarding alias behavior

fredrikekre:
This patch documents that `alias_A` and `alias_b` may depend on whether
the algorithm is known not to modify `A` or `b`. As a minimal first
change, this patch updates the default value for
`alg::AbstractKrylovSubspaceMethod` from `false` to `true`.


fredrikekre:
What is the relation to the `isinplace` type parameter for `LinearProblem`? That one seems to be unused by this package.

PR #316, opened 2023-06-01
Re-enable some precompilation

ChrisRackauckas:
Fix precompilation, requires the `@generated` function move until after the defaults are all defined. A bit ehh, but it works.


oscardssmith:
IMO, LinearSolve probably shouldn't be doing precompilation. Other packages that have specific types of solves should do the precompilation instead. Since Julia 1.8, the precompilation workload doesn't have to be in the same package as the code, so putting lots of precompiles here isn't necessary.

ChrisRackauckas:
But that duplicates precompilation. With such a quick precompilation, I don't see why it would be an issue.

PR #339, opened 2023-07-02
rm diffeqarrayoperator

vpuri3:
and bump version please


ChrisRackauckas:
This is missing the deprecation path, which would lead to issues like https://github.com/SciML/DifferentialEquations.jl/issues/967

ChrisRackauckas:
Already in master

Issue #340, opened 2023-07-03
`LinearSolve.solve` much slower than `A\b` for sparse matrices


lampretl:
I tried the new library LinearSolve.jl to improve the performance of solving large sparse linear systems of equations, that are implemented in SparseArrays.jl and SuiteSparse.jl. However, I noticed a significant worsening (in timing as well as accuracy):

```
using SparseArrays, LinearSolve
n, s = 10^4, 20;
A, b = sprand(n,n,s/n), sprand(n,s/n);

@time x = A \ Array(b);  
>  8.695524 seconds (94 allocations: 1.986 GiB, 0.11% gc time)
println(maximum(abs.(A*x-b)))  # check accuracy of solution
> 5.910361089433991e-11

@time (prb = init(LinearSolve.LinearProblem(A,b));   
       x = LinearSolve.solve(prb).u);
>  270.299651 seconds (449 allocations: 1.164 GiB, 0.00% gc time)
println(maximum(abs.(A*x-b)))  # check accuracy of solution
> 1.723447147217172e-7
```

Am I doing something wrong? I used Julia 1.9.1 in the REPL with LinearSolve v2.3.0.

ChrisRackauckas:
Fun one. The issue is that your matrix "isn't sparse". See:

```julia
using SparseArrays, LinearSolve
n, s = 10^4, 20;
A, b = sprand(n,n,s/n), sprand(n,s/n);

@time x = A \ Array(b); 
# 2.748556 seconds (88 allocations: 2.247 GiB, 0.15% gc time)

prb = LinearSolve.LinearProblem(A,b);
@time LinearSolve.solve(prb, UMFPACKFactorization());
# 2.583652 seconds (94 allocations: 2.250 GiB, 0.12% gc time)

@time LinearSolve.solve(prb, KLUFactorization());
# 88.481454 seconds (106.15 k allocations: 1.173 GiB, 0.00% gc time, 0.03% compilation time: 100% of which was recompilation)
```

Julia's `\` on sparse matrices is dumb and defaults to only using UMFPACK. UMFPACK is much faster than KLU for equations   because how it works is that it finds transformations to get "sufficiently dense" blocks and then does the multithreaded LAPACK LU factorization on those subsystems. Since your system is so dense, it just essentially does LAPACK on it, multithreaded and all, and is fast treating it as just a dense matrix.

But this then begs the question, how should the heuristic fix this? Of course it should take into account the amount of sparsity, but what is the cutoff?

[...]

PR #343, opened 2023-07-16
Set a sparsity threshold for KLU

ChrisRackauckas:
Closes https://github.com/SciML/LinearSolve.jl/issues/340. While many results are very much based on more details about the sparsity pattern, it's at least clear that if the matrix is not very sparse then it should probably use UMFPACK, so this is a conservative bound. When mixed with the length bound, I find that this seems to work rather well, though is not always optimal, is at least better than always choosing UMFPACK


ChrisRackauckas:
Screenshot 2023-07-15 at 9 36 13 PM

PR #349, opened 2023-07-31
Setup MKL direct factorizations

ChrisRackauckas:
MWE:

```julia
using LinearSolve, MKL_jll
A = rand(4, 4); b = rand(4); u0 = zeros(4);
lp = LinearProblem(A, b);
truesol = solve(lp, LUFactorization())
mklsol = solve(lp, MKLLUFactorization())
@test truesol ≈ mklsol
```

The segfault can be reproduced just with the triangular solver. MWE without LinearSolve:

```julia
using MKL_jll
using LinearAlgebra: BlasInt, LU
using LinearAlgebra.LAPACK: require_one_based_indexing, chkfinite, chkstride1, chkargsok
const usemkl = MKL_jll.is_available()

function getrf!(A::AbstractMatrix{<:Float64}; ipiv = similar(A, BlasInt, min(size(A,1),size(A,2))), info = Ref{BlasInt}(), check = false)
    require_one_based_indexing(A)
    check && chkfinite(A)
    chkstride1(A)
    m, n = size(A)
    lda  = max(1,stride(A, 2))
    ccall((:dgetrf_, MKL_jll.libmkl_rt), Cvoid,
            (Ref{BlasInt}, Ref{BlasInt}, Ptr{Float64},
            Ref{BlasInt}, Ptr{BlasInt}, Ptr{BlasInt}),
            m, n, A, lda, ipiv, info)
    chkargsok(info[])
    A, ipiv, info[] #Error code is stored in LU factorization type
end

A = rand(4,4); b = rand(4)
getrf!(A)
LU(getrf!(A)...) \ b
```


ChrisRackauckas:
@staticfloat or @chriselrod maybe you might have an idea of what I'm doing wrong here to cause a segfault?

ext/LinearSolveMKLExt.jl
```diff
@@ -0,0 +1,54 @@
...
+    check && chkfinite(A)
+    chkstride1(A)
+    m, n = size(A)
+    lda  = max(1,stride(A, 2))
+
+    if isempty(ipiv)
+        ipiv = similar(A, BlasInt, min(size(A,1),size(A,2)))
+    end
+
+    ccall((:dgetrf_, MKL_jll.libmkl_rt), Cvoid,
```
chriselrod, line 23: https://github.com/JuliaLinearAlgebra/BLASBenchmarksCPU.jl/blob/master/src/ccallblas.jl

I don't seem to be adding a trailing underscore to `dgetrf_` `_`

staticfloat:
You've got MKL in 32-bit mode, but passing it `Int64`'s, so the pivot indices that come back are gibberish, which causes the future LU factorization to have bogus indices, and the `\` to segfault:

```
julia> A, ipiv, info = getrf!(A);

julia> ipiv
4-element Vector{Int64}:
  8589934595
 17179869187
           0
           0
```

There are a bunch of ways to fix this, but the best is to import `@blasfunc` (remember that guy we dropped when copy-pasting) from `LinearAlgebra.BLAS` and use that.  That's good because you're already polymorphic by using `similar(..., BlasInt, ...)` when initializing `ipiv`, so this just makes it match up to the correct version by default.

PR #361, opened 2023-08-09
Setup Accelerate and MKL for 32-bit, MKL getrf, and Metal.jl integration

ChrisRackauckas:
Kind of a lot in a single PR but this is what I ended up with after a flight


christiangnrd:
Tried to run the Metal benchmark and I'm getting this error:

`ERROR: MethodError: no method matching do_factorization(::MetalLUFactorization, ::Matrix{Float32}, ::Vector{Float32}, ::Vector{Float32})`

ChrisRackauckas:
You need to re-instantiate for it to bring in the extension from a PR branch.

benchmarks/cudalu.jl
```diff
@@ -0,0 +1,51 @@
...
+                     sum((k + 1):n) do j
+            isempty((k + 1):m) ? 0 : sum((k + 1):m) do i
+                innerflop
+            end
+        end
+        invflop + scaleflop + updateflop
+    end
+end
+
+algs = [MKLLUFactorization(), CUDAOffloadFactorization()]
```
christiangnrd, line 21:
```suggestion
algs = [MKLLUFactorization(), CudaOffloadFactorization()]
```

ViralBShah:
Why do we need MKL 32-bit?

ChrisRackauckas:
GPU comparison, Neural ODE training, mixed precision algorithms, etc.

PR #365, opened 2023-08-14
Add sparselu benchmark

j-fu:
So far with direct solvers.


ChrisRackauckas:
Awesome. I'll accept it here and open the SciMLBenchmarks one.

PR #371, opened 2023-09-16
try documenter 1.0 upgrade

ArnoStrouwen:


ArnoStrouwen:
@mortenpi Could you have a look at the failure here:
https://github.com/SciML/LinearSolve.jl/actions/runs/6208844246/job/16855431354?pr=371#step:5:11

mortenpi:
Alright, the main error is a Documenter regression. It actually comes from the use of the `@doc_str` macro here: https://github.com/SciML/SciMLBase.jl/blob/8d293bfc0b859b1ae870772aaf023d2a93f6eb45/src/problems/basic_problems.jl#L1

PR #390, opened 2023-10-06
HOTFIX: Don't use MKL with numbers other that Float64

ChrisRackauckas:


src/default.jl
```diff
@@ -170,7 +170,7 @@ function defaultalg(A, b, assump::OperatorAssumptions)
                     DefaultAlgorithmChoice.RFLUFactorization
                     #elseif A === nothing || A isa Matrix
                     #    alg = FastLUFactorization()
-                elseif usemkl
+                elseif usemkl && eltype(A) <: Float64
```
ChrisRackauckas, line 173:
```suggestion
                elseif usemkl && eltype(A) <: Union{Float64, Float32}
```

src/default.jl
```diff
@@ -179,7 +179,7 @@ function defaultalg(A, b, assump::OperatorAssumptions)
                 DefaultAlgorithmChoice.QRFactorization
             elseif __conditioning(assump) === OperatorCondition.SuperIllConditioned
                 DefaultAlgorithmChoice.SVDFactorization
-            elseif usemkl
+            elseif usemkl && eltype(A) <: Float64
```
ChrisRackauckas, line 182:
```suggestion
            elseif usemkl && eltype(A) <: Union{Float64, Float32}
```

PR #393, opened 2023-10-16
Proper handling for BandedMatrices

avik-pal:
This is being developed parallel to https://github.com/SciML/BoundaryValueDiffEq.jl/pull/124. I will mark it as ready once properly tested!


ChrisRackauckas:
What's going on with v1.8? Should we just version gate `ldiv!`?

avik-pal:
Seems like a weird one-off failure? The previous CIs for 1.8 did pass

avik-pal:
This is good to go on my end!

PR #397, opened 2023-10-20
Promote abstol and reltol type to match eltype(A)

### LINKED ISSUE #396: Promote `abstol` and `reltol` type to match `eltype(A)`

ma-sadeghi:
Fixes #396


src/common.jl
```diff
@@ -150,6 +150,10 @@ function SciMLBase.init(prob::LinearProblem, alg::SciMLLinearSolveAlgorithm,
         fill!(u0, false)
     end
 
+    # Guard against type mismatch for user-specified reltol/abstol
+    reltol = eltype(prob.A)(reltol)
+    abstol = eltype(prob.A)(abstol)
```
ChrisRackauckas, line 155:
```suggestion
    reltol = eltype(prob.b)(reltol)
    abstol = eltype(prob.b)(abstol)
```

A might be an operator.

ma-sadeghi: Nice catch, thanks!

ma-sadeghi:
Should we change here as well? (to `prob.b`)

https://github.com/SciML/LinearSolve.jl/blob/f4f69409250ee359bb26e1409d56d53ca3d53e29/src/common.jl#L118-L119

ChrisRackauckas:
yes that would be good to change.

PR #399, opened 2023-10-23
Fix type promote in case b is complex

### LINKED ISSUE #398: Automatic type promotion of `b` breaks for complex values
See https://github.com/SciML/LinearSolve.jl/pull/397#issuecomment-1775661289

ma-sadeghi:
Fixes #398


ChrisRackauckas:
add a test?

ma-sadeghi:
> add a test?

Sure! Just to clarify, do you want to test the "complex" part or the "tol-type-mismatch"? If the former, I was thinking of adding a complex `prob3` to `basictests.jl`

ChrisRackauckas (approved):

ChrisRackauckas:
Tests are failing

test/basictests.jl
```diff
@@ -32,6 +42,9 @@ function test_interface(alg, prob1, prob2)
     sol = solve(prob1, alg; cache_kwargs...)
     @test A1 * sol.u ≈ b1
 
+    sol = solve(prob1, alg; cache_kwargs...)
```
ChrisRackauckas, line 45:
```suggestion
    sol = solve(prob2, alg; cache_kwargs...)
```

test/basictests.jl
```diff
@@ -9,37 +9,52 @@ const Dual64 = ForwardDiff.Dual{Nothing, Float64, 1}
...
 b2 = rand(n);
 x2 = zero(b);
+# Complex systems + mismatch types with eltype(tol)
+A3 = (A .+ rand(n) .* im) .|> ComplexF32;
+b3 = rand(n) .|> ComplexF32;
+x3 = zero(b) .|> ComplexF32;
+# A4 is similar to A3; created to test cache reuse
+A4 = (A .+ rand(n) .* im) .|> ComplexF32;
+b4 = b3 / 4 .|> ComplexF32;
+x4 = zero(b) .|> ComplexF32;
```
ma-sadeghi, line 27: @ChrisRackauckas A bunch of tests failed due to the type mismatch between A3 and A4 (used to be ComplexF64 and ComplexF32), so I made this change; but not sure if this was supposed to be handled internally, or it was expected that the cache mechanism should only work on systems with the same dtype.

test/basictests.jl
```diff
@@ -9,37 +9,52 @@ const Dual64 = ForwardDiff.Dual{Nothing, Float64, 1}
...
 
 function test_interface(alg, prob1, prob2)
-    A1 = prob1.A
-    b1 = prob1.b
-    x1 = prob1.u0
-    A2 = prob2.A
-    b2 = prob2.b
-    x2 = prob2.u0
+    A1, b1 = prob1.A, prob1.b
+    A2, b2 = prob2.A, prob2.b
```
ma-sadeghi, line 38: @ChrisRackauckas Not sure if this goes against SciML style, so feel free to revert back

ma-sadeghi:
All tests are passing except for MKL and SVD. From the error messages, I think the MKL wrapper doesn't support complex systems, but not sure what's going on with SVD.

test/basictests.jl
ChrisRackauckas: that's fine

test/basictests.jl
ChrisRackauckas: The cache mechanism only works with the same dtype.

ChrisRackauckas:
Yes the MKL_jll direct wrapper doesn't support complex, so mark that as broken.

SVD, mark as broken and open a Base Julialang issue.

ChrisRackauckas:
oh actually that seems like an issue with the cache construction, so ArrayInterface should get some updates and tests for SVD initialization on complex

ChrisRackauckas:
https://github.com/JuliaArrays/ArrayInterface.jl/blob/master/src/ArrayInterface.jl#L669

The SVD instance creation is only based on A, so the arrays are all complex typed. Should we just always convert b? I think that in order to hit the fast LAPACK functions it's required that b is complex as well and converts internally?

test/basictests.jl
```diff
@@ -202,25 +218,23 @@ end
...
+
+    if VERSION >= v"1.9"
+        push!(test_algs, MKLLUFactorization())
     end
 
     @testset "Concrete Factorizations" begin
         for alg in test_algs
             @testset "$alg" begin
                 test_interface(alg, prob1, prob2)
+                alg isa MKLLUFactorization || test_interface(alg, prob3, prob4)
```
ChrisRackauckas, line 237:
```suggestion
                VERSION >= v"1.9" && (alg isa MKLLUFactorization || test_interface(alg, prob3, prob4))
```

test/basictests.jl
```diff
@@ -202,25 +218,26 @@ end
...
+        SVDFactorization(),
+        RFLUFactorization(),
+        LinearSolve.defaultalg(prob1.A, prob1.b),
+    ]
+
+    if VERSION >= v"1.9" && LinearSolve.usemkl
+        push!(test_algs, MKLLUFactorization())
+    end
+
     end
```
ChrisRackauckas, line 234:
```suggestion
```

ma-sadeghi:
Thanks for the fixes. Quick question: Should we fix formatting here or in a separate PR (formatter complains about ext/LinearSolveEnzymeExt.jl, ext/LinearSolveIterativeSolversExt.jl, src/appleaccelerate.jl, src/default.jl, src/mkl.jl, test/basictests.jl, test/default_algs.jl, test/enzyme.jl, test/resolve.jl).

ChrisRackauckas:
in a separate PR. Right now we're checking if the new format is sufficient and will format all of SciML at the same time.

PR #400, opened 2023-10-24
Add `needs_square_A` trait

avik-pal:
Makes the following changes:

* New `needs_square_A` trait for solvers
* For underdetermined systems, we automatically use Pivoted QR similar to Base

### DIFF
--- a/src/LinearSolve.jl
+++ b/src/LinearSolve.jl
@@ -143,6 +143,46 @@ end
     include("factorization_sparse.jl")
 end
 
+# Solver Specific Traits
+## Needs Square Matrix
+"""
+    needs_square_A(alg)
+
+Returns `true` if the algorithm requires a square matrix.
+
+Note that this checks if the implementation of the algorithm needs a square matrix by
+trying to solve an underdetermined system. It is recommended to add a dispatch to this
+function for custom algorithms!
+"""
+needs_square_A(::Nothing) = false  # Linear Solve automatically will use a correct alg!
+function needs_square_A(alg::SciMLLinearSolveAlgorithm)
+    try
+        A = [1.0 2.0;
+            3.0 4.0;
+            5.0 6.0]
+        b = ones(Float64, 3)
+        solve(LinearProblem(A, b), alg)
+        return false
+    catch err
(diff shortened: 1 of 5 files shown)


src/LinearSolve.jl
```diff
@@ -143,6 +147,46 @@ end
...
+    needs_square_A(alg)
+
+Returns `true` if the algorithm requires a square matrix.
+
+Note that this checks if the implementation of the algorithm needs a square matrix by
+trying to solve an underdetermined system. It is recommended to add a dispatch to this
+function for custom algorithms!
+"""
+needs_square_A(::Nothing) = false  # Linear Solve automatically will use a correct alg!
+function needs_square_A(alg::SciMLLinearSolveAlgorithm)
```
ChrisRackauckas, line 162: Just remove the fallback please.

ChrisRackauckas: Or rather, the fallback is true. That's a safe fallback, methods can opt-in to supporting non-square, but anything that supports non-square supports square so that's a safe assumption.

src/default.jl
```diff
@@ -199,9 +200,19 @@ function defaultalg(A, b, assump::OperatorAssumptions)
     elseif assump.condition === OperatorCondition.WellConditioned
         DefaultAlgorithmChoice.NormalCholeskyFactorization
     elseif assump.condition === OperatorCondition.IllConditioned
-        DefaultAlgorithmChoice.QRFactorization
+        if is_underdetermined(A)
+            # Underdetermined
+            DefaultAlgorithmChoice.QRFactorizationPivoted
```
ChrisRackauckas, line 205: QRFactorization is already pivoted?

ChrisRackauckas: oh I see. @YingboMa what do you think of just always defaulting to pivoted QR?

YingboMa: Yeah, pivoted QR handles degenerate systems too.

YingboMa: But it's quite a bit slower though.

ChrisRackauckas (approved):

PR #401, opened 2023-10-25
Banded QR for non square matrices

avik-pal:


ChrisRackauckas:
Test?

avik-pal:
Can't install BlockDiagonals and BandedMatrices together in tests due to https://github.com/invenia/BlockDiagonals.jl/pull/126

PR #411, opened 2023-10-30
Don't use AppleAccelerate on number types that are not floats

SebastianM-C:
@ChrisRackauckas I saw that this was done for MKL in #390, but it seems to be missing for AppleAccelerate. Is this the correct way to handle it?


ElisabethRoesch:
I can confirm this fixed the issue on my Mac

ChrisRackauckas (approved):

ChrisRackauckas:
yup this is correct thanks

PR #417, opened 2023-11-05
Patch SimpleGMRES when `b`  is updated

avik-pal:


avik-pal:
@ChrisRackauckas buildkite doesn't seem to be running on PRs

ChrisRackauckas:
It only runs on the main package, not on forks

avik-pal:
@ChrisRackauckas  can you make https://buildkite.com/julialang/linearsolve-dot-jl/builds/1504 public? I am not sure what is failing. It passed locally for me

ChrisRackauckas:
Public.

PR #426, opened 2023-11-08
Remove SuiteSparse

sharanry:
We remove SuiteSparse as it has now been merged with SparseArrays.


sharanry:
@ChrisRackauckas this should be ready to go

Issue #427, opened 2023-11-09
Errors: No BLAS/LAPACK library loaded!


andreasvarga:
Recently, I tried to prepare a new version of [PeriodicSystems](https://github.com/andreasvarga/PeriodicSystems.jl) by starting with several commits. Running the tests, the following error pops up repeatedly

`Error: no BLAS/LAPACK library loaded!`

(see [here](https://github.com/andreasvarga/PeriodicSystems.jl/actions/runs/6798211169/job/18481922357))

The cause of these error messages is apparently the new version of `LinearSolve`, where `MKL` is loaded now by default instead `OpenBLAS` (see the discussion [here](https://discourse.julialang.org/t/again-errors-with-no-blas-lapack-library-loaded/105943)). I implemented several wrappers to routines from the [SLICOT](https://github.com/SLICOT/SLICOT-Reference) library, using the automatically generated [SLICOT_jll](https://github.com/JuliaBinaryWrappers/SLICOT_jll.jl). Apparently, some calls to `BLAS/LAPACK` are not covered there by` MKL`. I assume that other people using `OrdinaryDiffEq` may face the same issue.

I would appreciate if you could help me to overcome this situation, by indicating a way to preserve the behaviour before the last update. Many thanks in advance

akirakyle:
I also wanted to chime in and say that I started experiencing a different, but related issue updating today which brought LinearSolve to v2.18.0. I started experience serious performance regressions due to #408. In my case I'm running a parallel workload via Distributed and realized that after updating, suddenly every worker was launching a bunch of threads causing my system to be completely oversubscribed. It turns out, by default local distributed workers have `OPENBLAS_NUM_THREADS=1` set (https://github.com/JuliaLang/julia/commit/a8b399416208d91061324814ff8ae080a918e48b) but since `MKL_NUM_THREADS` is unset, MKL launches its default number of threads for the system on each worker.

I'll open a separate issue about setting `MKL_NUM_THREADS=1` on worker threads.

However I'm also often comparing the difference between openblas and mkl for by use cases and hardware and #408 makes it non-obvious how I can control this behavior. Especially since I now see that the first half of my computation uses openblas, then the second half uses mkl since apparently I'm only calling certain functions in `QuantumOptics` that don't require `OrdinaryDiffEq` in the first computationally expensive phase.

[...]

PR #428, opened 2023-11-10
Remove MKL

ChrisRackauckas:
See the discussion in https://github.com/SciML/LinearSolve.jl/issues/427


chriselrod:
Why weren't we just `ccall`ing the functions we want from `MLK_jll`?

ChrisRackauckas:
See the discussion in https://github.com/SciML/LinearSolve.jl/issues/427. You can't do that with SuiteSparse.

PR #435, opened 2023-11-29
Aqua + typos CI

ArnoStrouwen:


ArnoStrouwen:
@ChrisRackauckas Not sure what to do with the EnzymeCore stale dep error:
https://github.com/SciML/LinearSolve.jl/actions/runs/7032498284/job/19136359022?pr=435#step:6:829

ChrisRackauckas:
https://github.com/SciML/LinearSolve.jl/pull/435/files#diff-72ed386c2a0cd1d23c0968297e70023ed98c22490d146dd89fc91f48369bad4dR11 we have a dependency on EnzymeCore but I guess we don't use it in the package, since we just use Enzyme in the extension? That can probably be removed then.

PR #436, opened 2023-11-29
Support StaticArrays Properly

avik-pal:
## Checklist

  [contributor guidelines](https://github.com/SciML/.github/blob/master/CONTRIBUTING.md), in particular the [ScioML Style Guide](https://github.com/SciML/SciMLStyle) and


avik-pal:
@ChrisRackauckas this is ready

avik-pal:
To summarize what this does:

1. changes the default for StaticArrays from GMRES to LU for square and SVD for non-square (QR(..) \ b is not defined so can't default to that)
2. In a bunch of places `<fact>!` was used, for SArray that is changed to `fact`.
3. For `NormalCholesky` the instance was being created for `A`, which should have been for `Symmetric(A' * A)` -- unfortunately `A * A` pre-1.10 doesn't pass JET opt tests, so testing only 1.10 and above for that.

PR #437, opened 2023-12-03
start fixing ambiguities

ArnoStrouwen:
Fixes many of the method ambiguities in this package.
Things I'm still unsure about:

First, for `defaultalg` there is dispatch for `AbstractSciMLOperator` as well as dispatch for `OperatorAssumptions{Nothing}`.
`AbstractSciMLOperator` dispatches based on `A.A`.
`OperatorAssumptions{Nothing}` dispatches based on `issquare(A)`.
If both occur together, I don't know which order to apply them.
Currently, it first gets rid of the `Nothing`, but this means `issquare` has to be defined for `Operators`.

Second, there are remaining ambiguities regarding:
```julia
function init_cacheval(alg::GenericFactorization,
    A::Union{Hermitian{T, <:SparseMatrixCSC},
        Symmetric{T, <:SparseMatrixCSC}}, b, u, Pl, Pr,
    maxiters::Int, abstol, reltol, verbose::Bool,
    assumptions::OperatorAssumptions) where {T}
    newA = copy(convert(AbstractMatrix, A))
    do_factorization(alg, newA, b, u)
end
```
e.g. with:
```
function init_cacheval(alg::GenericFactorization{typeof(svd!)}, A::AbstractMatrix, b, u, Pl, Pr,
    maxiters::Int,
    abstol, reltol, verbose::Bool, assumptions::OperatorAssumptions)
    ArrayInterface.svd_instance(A)
end
```
I'm not sure how to resolve these, I'm assuming we want as little `do_factorization` as possible?

Third, for some special matrix types, such as diagonal, there is dispatch to always use the same method, no matter the algorithm. But there are also fallback methods for most algorithms to work for general `A`. This design choice leads to tons of additional function definitions to resolve ambiguities between the two.

### DIFF
--- a/src/factorization.jl
+++ b/src/factorization.jl
@@ -402,15 +402,15 @@ function do_factorization(alg::GenericFactorization, A, b, u)
     return fact
 end
 
-function init_cacheval(alg::GenericFactorization{typeof(lu)}, A, b, u, Pl, Pr,
+function init_cacheval(alg::GenericFactorization{typeof(lu)}, A::AbstractMatrix, b, u, Pl, Pr,
     maxiters::Int,
     abstol, reltol, verbose::Bool, assumptions::OperatorAssumptions)
-    ArrayInterface.lu_instance(convert(AbstractMatrix, A))
+    ArrayInterface.lu_instance(A)
 end
-function init_cacheval(alg::GenericFactorization{typeof(lu!)}, A, b, u, Pl, Pr,
+function init_cacheval(alg::GenericFactorization{typeof(lu!)}, A::AbstractMatrix, b, u, Pl, Pr,
     maxiters::Int,
     abstol, reltol, verbose::Bool, assumptions::OperatorAssumptions)
-    ArrayInterface.lu_instance(convert(AbstractMatrix, A))
+    ArrayInterface.lu_instance(A)
 end
 
 function init_cacheval(alg::GenericFactorization{typeof(lu)},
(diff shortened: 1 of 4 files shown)


src/factorization.jl
```diff
@@ -1242,37 +1318,44 @@ function FastQRFactorization()
...
         maxiters::Int, abstol, reltol, verbose::Bool,
         assumptions::OperatorAssumptions)
         ws = QRWYWs(A; blocksize = alg.blocksize)
         return WorkspaceAndFactors(ws,
-            ArrayInterface.qr_instance(convert(AbstractMatrix, A)))
+            ArrayInterface.qr_instance(A))
     end
 
-    function init_cacheval(::FastQRFactorization{Val{true}}, A, b, u, Pl, Pr,
+    function init_cacheval(::FastQRFactorization{Val{true}}, A::AbstractMatrix, b, u, Pl, Pr,
```
ChrisRackauckas, line 1329: what about the operator dispatches?

ChrisRackauckas: That's only for matrix operator. What about other abstract scimloperators?

ArnoStrouwen: There is none for that,
but note that this is currently not working anyway:

```julia
using LinearSolve
using SciMLOperators

N = 4
f = (u, p, t) -> u .* u

M = MatrixOperator(rand(N, N))
D = DiagonalOperator(rand(N))
F = FunctionOperator(f, zeros(N), zeros(N))

b = rand(4)

solve(LinearProblem(rand(N, N), b),FastQRFactorization()) # works
solve(LinearProblem(M, b),FastQRFactorization()) # fails
solve(LinearProblem(D, b),FastQRFactorization()) # fails
solve(LinearProblem(F, b),FastQRFactorization()) # fails
```

julia> solve(LinearProblem(M, b),FastQRFactorization())
ERROR: MethodError: init_cacheval(::FastQRFactorization{LinearAlgebra.NoPivot}, ::MatrixOperator{Float64, Matrix{Float64}, SciMLOperators.FilterKwargs{typeof(SciMLOperators.DEFAULT_UPDATE_FUNC), Tuple{}}, SciMLOperators.FilterKwargs{typeof(SciMLOperators.DEFAULT_UPDATE_FUNC), Tuple{}}}, ::Vector{Float64}, ::Vector{Float64}, ::IdentityOperator, ::IdentityOperator, ::Int64, ::Float64, ::Float64, ::Bool, ::OperatorAssumptions{Bool}) is ambiguous.

Candidates:
  init_cacheval(alg::FastQRFactorization, A::MatrixOperator, b, u, Pl, Pr, maxiters::Int64, abstol, reltol, verbose::Bool, assumptions::OperatorAssumptions)
    @ LinearSolve C:\Users\arno\.julia\packages\LinearSolve\8EgR1\src\factorization.jl:1180
  init_cacheval(alg::FastQRFactorization{LinearAlgebra.NoPivot}, A, b, u, Pl, Pr, maxiters::Int64, abstol, reltol, verbose::Bool, assumptions::OperatorAssumptions)
    @ LinearSolve C:\Users\arno\.julia\packages\LinearSolve\8EgR1\src\factorization.jl:1066

Possible fix, define
  init_cacheval(::FastQRFactorization{LinearAlgebra.NoPivot}, ::MatrixOperator, ::Any, ::Any, ::Any, ::Any, ::Int64, ::Any, ::Any, ::Bool, ::OperatorAssumptions)

ArnoStrouwen: For it to ever have worked for FunctionOperators it needed to have this defined
```jl
WorkspaceAndFactors(ws, ArrayInterface.qr_instance(A))
```

Anyway, the solution for DiagonalOperators seems straightforward to me. Just add the same block of code as for MatrixOperators. But how do you want to handle FunctionOperators?

ChrisRackauckas:
Resolve the conflicts?

ChrisRackauckas:
@avik-pal can you review this one?

avik-pal:
When can `issquare` be `nothing`? If we can't resolve the dimensions can't we assume rectangular

avik-pal:
Oh I see it is to disambiguate between no assumption provided from the user. This PR seems fine in that case

PR #439, opened 2023-12-04
Remove the unnecessary < 1.9 checks

avik-pal:


avik-pal:
@ChrisRackauckas this is good to go

PR #444, opened 2023-12-11
Proper handling of static arrays

avik-pal:


ChrisRackauckas:
Could we add an inference and no alloc check on static array usage too?

avik-pal:
There is this 1 allocation while creating the LinearCache which I can't seem to track
image

ChrisRackauckas:
This is kind of the same thing as the SimpleNonlinearSolve/SimpleDiffEq case. If someone calls `solve` and not `init`, then with static things doing something simpler is advantageous. I would think that if someone calls `solve` on a StaticArray we should basically just do `A\b` and bypass the rest of LinearSolve.jl. I did a bunch of testing when making https://github.com/SciML/OrdinaryDiffEq.jl/blob/master/src/derivative_utils.jl#L1-L19 and for sufficiently small problems caching the LU with a static array (or the inv) is simply not helpful (track down the PR with the old benchmarks). This could be size dependent too.

Now if they do `init` I think this all makes sense though, just `solve` should have a fast path.

avik-pal:
Yeah that makes sense. Though I feel even for `init` with SArrays we shouldn't really cache anything, just pretend to cache it. I have found operations mixing MArrays and SArrays to be slower than just doing the uncached SArrays

ChrisRackauckas:
Agreed

PR #449, opened 2023-12-20
Adjoints for Linear Solve

### LINKED ISSUE #198: adjoint support
```julia
using LinearSolve, SciMLSensitivity, Zygote

N = 4
b = rand(4)
p = rand(4)

function loss(p)
    A = p * p'
    prob = LinearProblem(A, b)
    u = solve(prob)

    u' * u
end

@show loss(p)
@show Zygote.gradient(loss, p)
```
```julia
julia> include("ln.jl")                                                             
loss(p) = 3.7452907979624124e34                                                     
ERROR: LoadError: Need an adjoint for constructor SciMLBase.LinearSolution{Float64, 
1, Vector{Float64}, Nothing, GenericLUFactorization{LinearAlgebra.RowMaximum}, Linea
rSolve.LinearCache{Matrix{Float64}, Vector{Float64}, Vector{Float64}, SciMLBase.Null
Parameters, Nothing, LinearAlgebra.LU{Float64, Matrix{Float64}, Vector{Int64}}, Iter
ativeSolvers.Identity, IterativeSolvers.Identity, Float64, Nothing}}. Gradient is of
 type Vector{Float64}                                                               
Stacktrace:                                                                         
  [1] error(s::String)                                                              
    @ Base ./error.jl:35                                                            
  [2] (::Zygote.Jnew{SciMLBase.LinearSolution{Float64, 1, Vector{Float64}, Nothing, 
GenericLUFactorization{LinearAlgebra.RowMaximum}, LinearSolve.LinearCache{Matrix{Flo
at64}, Vector{Float64}, Vector{Float64}, SciMLBase.NullParameters, Nothing, LinearAl
gebra.LU{Float64, Matrix{Float64}, Vector{Int64}}, IterativeSolvers.Identity, Iterat
iveSolvers.Identity, Float64, Nothing}}, Nothing, false})(Δ::Vector{Float64})       
    @ Zygote ~/.julia/packages/Zygote/PD12J/src/lib/lib.jl:327                      
```

### LINKED ISSUE #322: Dispatch on `__solve` instead of `solve`
Similar to other packages in SciMLverse, we should have the solvers dispatch on `__solve`. This would allow us to define the adjoint equations in SciMLSensitivity.jl https://github.com/SciML/SciMLSensitivity.jl/pull/808#issuecomment-1574984670

avik-pal:
Fixes #198, Fixes #322

## TODOs:

- [ ] Preconditioning. These need to be added to the Adjoint Sensitivity Struct. Can we use a left preconditioner for the forward problem as the transpose right preconditioner? Need someone with a better grasp of linear algebra to chime in here
- [x] Support the `linsolve !== nothing` case. This is useful if we know that $A^T$ has a structure exploitable by a different solver
- [x] Tests -- Take from Enzyme and ChainRules
- [x] https://github.com/SciML/SciMLBase.jl/pull/566

## Example:

```julia
using LinearSolve, Zygote

A = rand(4, 4)
b = rand(4)

test_func_1(A, b) = sum(abs2, A \ b)

test_func_1(A, b)

∂A_1, ∂b_1 = @btime Zygote.gradient(test_func_1, copy(A), copy(b))
display(∂A_1)
display(∂b_1)

function test_func_2(A, b)
    prob = LinearProblem(A, b)
    sol = solve(prob)
    return sum(abs2, sol.u)
end

test_func_2(A, b)

∂A_2, ∂b_2 = @btime Zygote.gradient(test_func_2, copy(A), copy(b))
display(∂A_2)
display(∂b_2)
```

In the following case the cache stores the correct gradients but they are not propagated to `A` and `b`. @ChrisRackauckas  any idea how to fix this?

```julia
cache = init(LinearProblem(copy(A), copy(b)), nothing);
function test_func_3(cache, A, b)
    cache.A = A
    cache.b = b
    sol = solve!(cache)
    return sum(abs2, sol.u)
end

test_func_3(cache, copy(A), copy(b))

∂cache, ∂A_3, ∂b_3 = @btime Zygote.gradient(test_func_3, cache, copy(A), copy(b))
∂cache.A
∂cache.b
display(∂A_3)  # nothing
display(∂b_3)  # nothing
```


src/adjoint.jl
```diff
@@ -0,0 +1,103 @@
...
+            A_ = cache.alias_A ? deepcopy(A) : A
+        end
+    else
+        error("Not Implemented Yet!!!")
+    end
+
+    # Forward Solve
+    sol = solve!(cache, alg, args...; kwargs...)
+
+    function ∇solve!(∂sol)
```
ChrisRackauckas, line 63: Don't we technically have to deepcopy in here?

ChrisRackauckas:
> In the following case the cache stores the correct gradients but they are not propagated to A and b. @ChrisRackauckas any idea how to fix this?

Is this not just an inherent limitation of Zygote with mutation? I would presume we just need to stay away from that and only support `solve` with CRC.

src/adjoint.jl
avik-pal: I guess so, it can be problematic if there are 2 subsequent solve calls on the cache.

ChrisRackauckas (approved):

PR #468, opened 2024-02-11
Avoid constructing any Krylov cache with default dense

ChrisRackauckas:


ChrisRackauckas:
@avik-pal try it when this patch is out.

PR #474, opened 2024-02-23
Fix `UMFPACK` to use `check=false`

oscardssmith:
This prevents singular exceptions.

## Checklist

### DIFF
--- a/test/default_algs.jl
+++ b/test/default_algs.jl
@@ -36,11 +36,12 @@ solve(prob)
       LinearSolve.DefaultAlgorithmChoice.QRFactorization
 
 
-A = spzeros(100, 100)
-A[1,1]=1
+A = spzeros(2, 2)
 # test that solving a singular problem doesn't error
-prob = LinearProblem(A, ones(100))
-solve(prob)
+prob = LinearProblem(A, ones(2))
+@test solve(prob, UMFPACKFactorization()).retcode == ReturnCode.Infeasible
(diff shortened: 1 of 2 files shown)


test/default_algs.jl
```diff
@@ -35,6 +35,13 @@ solve(prob)
     LinearSolve.OperatorAssumptions(false)).alg ===
       LinearSolve.DefaultAlgorithmChoice.QRFactorization
 
+
+A = spzeros(100, 100)
+A[1,1]=1
+# test that solving a singular problem doesn't error
+prob = LinearProblem(A, ones(100))
+solve(prob)
```
ChrisRackauckas, line 43: This only shows one solver. Can you make KLU and UMFPACK explicit?

oscardssmith: updated. KLU is still broken (as far as I can see it still doesn't support disabling the checks).

ChrisRackauckas: @rayegun didn't you handle KLU.jl?

ChrisRackauckas: https://github.com/SciML/LinearSolve.jl/pull/456#issuecomment-1935228441

https://github.com/search?q=repo%3AJuliaSparse%2FKLU.jl%20check&type=code

You said you added `check=false` to klu but I don't see it in the repo. We can merge for now with this test failing but can you follow up with a test passing verison?

Issue #383, opened 2023-10-01
How to set UMFPACK control vector ?


sjdaines:
Apologies if I've missed something, but I can't see how to set the UMFPACK control vector...

Would this be possible perhaps as a new `control` kwarg to `UMFPACKFactorization` ?

which could then be used eg to reenable iterative refinement:

    import SparseArrays 

    umfpack_control = SparseArrays.UMFPACK.get_umfpack_control(Float64, Int64) # get Julia defaults (NB: Julia modifies the UMFPACK default to switch off iterative refinement)
    umfpack_control[SparseArrays.UMFPACK.JL_UMFPACK_IRSTEP] = 2.0 # reenable UMFPACK default max iterative refinement steps

    UMFPACKFactorization(;reuse_symbolic=true, check_pattern=true, control=umfpack_control)

where this could perhaps be implemented as 

    Base.@kwdef struct UMFPACKFactorization <: AbstractFactorization
        reuse_symbolic::Bool = true
        check_pattern::Bool = true # Check factorization re-use
        control::Vector{Float64} = Float64[] # empty vector will be filled in with defaults from a call to SparseArrays.UMFPACK.get_umfpack_control
    end

ChrisRackauckas:
I didn't know there was more controls here. Are they documented somewhere?

Instead of specifying the control vector, we should raise it to a higher level interface. `UMFPACKFactorization(;max_iterative_refinement_steps = 2)` or something.

sjdaines:
> I didn't know there was more controls here. Are they documented somewhere?

Just made a PR to add some more doc https://github.com/JuliaSparse/SparseArrays.jl/pull/449

ChrisRackauckas:
Cool thanks for sharing. I don't have a free hand right now to complete this, but would happily take a PR if anyone has a second.

PaoloBiolghini:
I noticed in the SparseArrays code that they force the value of the control vector to zero when you want to change it.
![image](https://github.com/SciML/LinearSolve.jl/assets/61747117/9daec93f-641a-4dda-ad09-cd1d85850480)

We need to open a PR for SparseArrays to create a way to plug in the control vector into the lu function, which is called from LinearSolver.jl.

So, in order to change the control parameters for LinearSolve, as you suggested, we have to modify the UMFPACKFactorization struct and also the solve function, adding the control vector as parameters to the lu function.

I'm going to open the PR and link it to these issues.

PR #482, opened 2024-03-15
Added the posibility to modify the control paramteres of UMFPACK alg

PaoloBiolghini:
#383 
I have added a vector of type Float 64 in the UMFPACKFactorization structure in order to give the possibility to the user to plug in their customized control vector int the UMFPACK algorithm; 
This control vector is used in the solve function and is passed to SparseArrays through the lu function.

## Checklist

## Additional context

For a deeper understing on the control vector i sugest to read subsection 5.10 at page 23 in the following pdf
https://github.com/DrTimothyAldenDavis/SuiteSparse/blob/dev/UMFPACK/Doc/UMFPACK_UserGuide.pdf

### DIFF
--- a/src/factorization.jl
+++ b/src/factorization.jl
@@ -750,10 +750,6 @@ patterns with “more structure”.
     `A` has the same sparsity pattern as the previous `A`. If this algorithm is to
     be used in a context where that assumption does not hold, set `reuse_symbolic=false`.
 
-    There is also the possibility to set your own control array for the factorization that will be  
-    passed at the SparseArrays package. To do so u have to pass a Vector{Float64} with 20 elements.
-    If no vecetor is passed the default control array is used.
-    More details on the UMFPACK doc
 """
 
 Base.@kwdef struct UMFPACKFactorization{T <: Union{Nothing, Vector{Float64}}} <: AbstractFactorization
--- a/src/factorization.jl
+++ b/src/factorization.jl
@@ -735,6 +735,14 @@ const default_control=[1.0, 0.2, 0.2, 0.1, 32.0, 0.0, 0.7, 0.0, 1.0, 0.3, 1.0, 1
 A fast sparse multithreaded LU-factorization which specializes on sparsity
 patterns with “more structure”.
 
+## Fields
+
+* `reuse_symbolic`: Whether the symbolic factorization is reused between calls. This requires that the sparsity pattern is
+  preserved. Defaults to true.
+* `check_pattern`: Whether the sparsity pattern is checked for changes to allow for symbolic factorization caching. 
+  The default is true.
+* `control`: A control vector for more options to pass to UMFPACK. See the UMFPACK documentation for more details.
+
 !!! note
 
     By default, the SparseArrays.jl are implemented for efficiency by caching the
--- a/src/factorization.jl
+++ b/src/factorization.jl
@@ -727,7 +727,7 @@ end
 
 #!!!ATTENTION
 #These array could change and whenever there will be a change in the SparseArrays package this array must be changed
-const default_control=[1.0, 0.2, 0.2, 0.1, 32.0, 0.0, 0.7, 0.0, 1.0, 0.3, 1.0, 1.0, 0.9, 0.0, 10.0, 0.001, 1.0, 0.5, 0.0, 1.0]
+const default_control=SparseArrays.UMFPACK.get_umfpack_control(Float64,Int64)
 
 """
 `UMFPACKFactorization(;reuse_symbolic=true, check_pattern=true, control=Vector{Float64}(20))`
--- a/src/factorization.jl
+++ b/src/factorization.jl
@@ -738,10 +738,10 @@ patterns with “more structure”.
 """
 const default_control::Array{FLoat64}=[1.0, 0.2, 0.2, 0.1, 32.0, 0.0, 0.7, 0.0, 1.0, 0.3, 1.0, 1.0, 0.9, 0.0, 10.0, 0.001, 1.0, 0.5, 0.0, 1.0]
 
-Base.@kwdef struct UMFPACKFactorization <: AbstractFactorization
+Base.@kwdef struct UMFPACKFactorization{T <: Union{Nothing, Vector{Float64}} <: AbstractFactorization
     reuse_symbolic::Bool = true
     check_pattern::Bool = true # Check factorization re-use
-    control::Vector{Float64}=nothing
+    control::T=nothing
 end
 
 const PREALLOCATED_UMFPACK = SparseArrays.UMFPACK.UmfpackLU(SparseMatrixCSC(0, 0, [1],
--- a/src/factorization.jl
+++ b/src/factorization.jl
@@ -736,7 +736,7 @@ patterns with “more structure”.
     `A` has the same sparsity pattern as the previous `A`. If this algorithm is to
     be used in a context where that assumption does not hold, set `reuse_symbolic=false`.
 """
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
Needs docs.

src/factorization.jl
```diff
@@ -739,6 +739,7 @@ patterns with “more structure”.
 Base.@kwdef struct UMFPACKFactorization <: AbstractFactorization
     reuse_symbolic::Bool = true
     check_pattern::Bool = true # Check factorization re-use
+    control::Vector{Float64}=SparseArrays.UMFPACK.get_umfpack_control(Float64,Int64) # Control Parameters of UMFPACK
```
ChrisRackauckas, line 742: This will heap allocate each time in the default algorithm. Instead, make a const of the default, set the default here to `nothing`, and then if `nothing` use the const global.

src/factorization.jl
```diff
@@ -736,9 +736,12 @@ patterns with “more structure”.
     `A` has the same sparsity pattern as the previous `A`. If this algorithm is to
     be used in a context where that assumption does not hold, set `reuse_symbolic=false`.
 """
+const default_control::Array{FLoat64}=[1.0, 0.2, 0.2, 0.1, 32.0, 0.0, 0.7, 0.0, 1.0, 0.3, 1.0, 1.0, 0.9, 0.0, 10.0, 0.001, 1.0, 0.5, 0.0, 1.0]
+
 Base.@kwdef struct UMFPACKFactorization <: AbstractFactorization
     reuse_symbolic::Bool = true
     check_pattern::Bool = true # Check factorization re-use
+    control::Vector{Float64}=nothing
```
ChrisRackauckas, line 744:
```suggestion
Base.@kwdef struct UMFPACKFactorization{T <: Union{Nothing, Vector{Float64}} <: AbstractFactorization
    reuse_symbolic::Bool = true
    check_pattern::Bool = true # Check factorization re-use
    control::T=nothing
```

src/factorization.jl
```diff
@@ -736,9 +736,12 @@ patterns with “more structure”.
     `A` has the same sparsity pattern as the previous `A`. If this algorithm is to
     be used in a context where that assumption does not hold, set `reuse_symbolic=false`.
 """
+const default_control::Array{FLoat64}=[1.0, 0.2, 0.2, 0.1, 32.0, 0.0, 0.7, 0.0, 1.0, 0.3, 1.0, 1.0, 0.9, 0.0, 10.0, 0.001, 1.0, 0.5, 0.0, 1.0]
```
ChrisRackauckas, line 739: This  is blocking the docstring above from going to the right thing.

ChrisRackauckas, line 739:
```suggestion
const default_control=[1.0, 0.2, 0.2, 0.1, 32.0, 0.0, 0.7, 0.0, 1.0, 0.3, 1.0, 1.0, 0.9, 0.0, 10.0, 0.001, 1.0, 0.5, 0.0, 1.0]
```

src/factorization.jl
```diff
@@ -736,9 +736,12 @@ patterns with “more structure”.
     `A` has the same sparsity pattern as the previous `A`. If this algorithm is to
```
ChrisRackauckas, line 736: Missing docs

src/factorization.jl
```diff
@@ -723,8 +723,14 @@ end
...
+#!!!ATTENTION
+#These array could change and whenever there will be a change in the SparseArrays package this array must be changed
+const default_control=[1.0, 0.2, 0.2, 0.1, 32.0, 0.0, 0.7, 0.0, 1.0, 0.3, 1.0, 1.0, 0.9, 0.0, 10.0, 0.001, 1.0, 0.5, 0.0, 1.0]
+
 """
-`UMFPACKFactorization(;reuse_symbolic=true, check_pattern=true)`
+`UMFPACKFactorization(;reuse_symbolic=true, check_pattern=true, control=Vector{Float64}(20))`
 
 A fast sparse multithreaded LU-factorization which specializes on sparsity
 patterns with “more structure”.
```
ChrisRackauckas, line 736:
```suggestion
`UMFPACKFactorization(;reuse_symbolic=true, check_pattern=true, control=Vector{Float64}(20))`

A fast sparse multithreaded LU-factorization which specializes on sparsity
patterns with “more structure”.

## Fields

* `reuse_symbolic`: Whether the symbolic factorization is reused between calls. This requires that the sparsity pattern is
  preserved. Defaults to true.
* `check_pattern`: Whether the sparsity pattern is checked for changes to allow for symbolic factorization caching. 
  The default is true.
* `control`: A control vector for more options to pass to UMFPACK. See the UMFPACK documentation for more details.
```

src/factorization.jl
```diff
@@ -723,22 +723,43 @@ end
...
 
     By default, the SparseArrays.jl are implemented for efficiency by caching the
     symbolic factorization. I.e., if `set_A` is used, it is expected that the new
     `A` has the same sparsity pattern as the previous `A`. If this algorithm is to
     be used in a context where that assumption does not hold, set `reuse_symbolic=false`.
+
+    There is also the possibility to set your own control array for the factorization that will be  
+    passed at the SparseArrays package. To do so u have to pass a Vector{Float64} with 20 elements.
+    If no vecetor is passed the default control array is used.
+    More details on the UMFPACK doc
```
ChrisRackauckas, line 756:
```suggestion
```

src/factorization.jl
```diff
@@ -723,22 +723,39 @@ end
 
 ################################## Factorizations which require solve! overloads
 
+# Default control array of 20 elements from Umfpack
+
+#!!!ATTENTION
+#These array could change and whenever there will be a change in the SparseArrays package this array must be changed
+const default_control=[1.0, 0.2, 0.2, 0.1, 32.0, 0.0, 0.7, 0.0, 1.0, 0.3, 1.0, 1.0, 0.9, 0.0, 10.0, 0.001, 1.0, 0.5, 0.0, 1.0]
```
ChrisRackauckas, line 730:
```suggestion
const default_control=SparseArrays.UMFPACK.get_umfpack_control(Float64,Int64)
```

src/factorization.jl
```diff
@@ -723,22 +723,39 @@ end
...
 A fast sparse multithreaded LU-factorization which specializes on sparsity
 patterns with “more structure”.
 
+## Fields
+
+* `reuse_symbolic`: Whether the symbolic factorization is reused between calls. This requires that the sparsity pattern is
+  preserved. Defaults to true.
+* `check_pattern`: Whether the sparsity pattern is checked for changes to allow for symbolic factorization caching. 
+  The default is true.
+* `control`: A control vector for more options to pass to UMFPACK. See the UMFPACK documentation for more details.
```
ChrisRackauckas, line 744: We should fill in the details here, since it has been unchanged for 20 years. Is there something simple that can be pasted, quoted, or linked to?

ChrisRackauckas:
Tests fail https://github.com/SciML/LinearSolve.jl/actions/runs/8315251867/job/22753397359?pr=482#step:6:543

src/factorization.jl
PaoloBiolghini: eventualli we could somehow  insert the table present in the doc of UFMPACK 
image

PaoloBiolghini: do u think would be appropriate?

AJ0070:
@ChrisRackauckas #1200 covers #383, so this one can probably be closed.

For the record on why it stalled: `lu!` has no `control` keyword on 1.10, 1.11 or 1.12, so passing it on the symbolic reuse path is the test failure you linked above. `default_control` is also a load-time const pinned to `(Float64, Int64)`, so it does not match a `ComplexF64` or `Int32` matrix, and being a shared mutable vector it can be edited by any caller.

#1200 takes the settings by name rather than the raw vector, and builds the control per element type at solve time. Credit to @PaoloBiolghini for spotting the control-zeroing behaviour in SparseArrays, which is what the newer version is designed around.

PR #493, opened 2024-04-19
Symbolic Factorization Reuse in the standard LUFactorization

ChrisRackauckas:
This should be sufficient for CUDSS.jl to be optimally used as well


ChrisRackauckas:
This is hard to test, but it is now working with CUDSS and I can check the symbolic factorizations are reused! MWE:

```julia
using CUDA, CUDA.CUSPARSE
using CUDSS
using SparseArrays, LinearAlgebra
using LinearSolve

T = Float64
n = 100
A_cpu = sprand(T, n, n, 0.05) + I
x_cpu = zeros(T, n)
b_cpu = rand(T, n)

A_gpu = CuSparseMatrixCSR(A_cpu)
x_gpu = CuVector(x_cpu)
b_gpu = CuVector(b_cpu)
prob = LinearProblem(A_gpu, b_gpu)
sol = solve(prob, LUFactorization())

cache = init(prob, LUFactorization())
solve!(cache)
cache.b = CuVector(rand(T, n))
solve!(cache)
cache.A = CuSparseMatrixCSR(A_cpu + 3I)
solve!(cache)
cache.b = CuVector(rand(T, n))
solve!(cache)
```

This is with https://github.com/JuliaArrays/ArrayInterface.jl/pull/433 and https://github.com/exanauts/CUDSS.jl/pull/33

PR #431, opened 2023-11-12
WIP: Wrap BLIS

ChrisRackauckas:
Test case:

```julia
using LinearSolve, blis_jll

A = rand(4, 4)
b = rand(4)
prob = LinearProblem(A, b)
sol = solve(prob,LinearSolve.BLISLUFactorization())
sol.u
```

throws:

```julia
julia> sol = solve(prob,LinearSolve.BLISLUFactorization())
ERROR: TypeError: in ccall: first argument not a pointer or valid constant expression, expected Ptr, got a value of type Tuple{Symbol, Ptr{Nothing}}
Stacktrace:
 [1] getrf!(A::Matrix{Float64}; ipiv::Vector{Int64}, info::Base.RefValue{Int64}, check::Bool)
   @ LinearSolveBLISExt ~/.julia/dev/LinearSolve/ext/LinearSolveBLISExt.jl:67
 [2] getrf!
   @ LinearSolveBLISExt ~/.julia/dev/LinearSolve/ext/LinearSolveBLISExt.jl:55 [inlined]
 [3] #solve!#9
   @ LinearSolveBLISExt ~/.julia/dev/LinearSolve/ext/LinearSolveBLISExt.jl:222 [inlined]
 [4] solve!
   @ LinearSolveBLISExt ~/.julia/dev/LinearSolve/ext/LinearSolveBLISExt.jl:216 [inlined]
 [5] #solve!#6
   @ LinearSolve ~/.julia/dev/LinearSolve/src/common.jl:209 [inlined]
 [6] solve!
   @ LinearSolve ~/.julia/dev/LinearSolve/src/common.jl:208 [inlined]
 [7] #solve#5
   @ LinearSolve ~/.julia/dev/LinearSolve/src/common.jl:205 [inlined]
 [8] solve(::LinearProblem{…}, ::LinearSolve.BLISLUFactorization)
   @ LinearSolve ~/.julia/dev/LinearSolve/src/common.jl:202
 [9] top-level scope
   @ REPL[8]:1
Some type information was truncated. Use `show(err)` to see complete types.
```

PR #498, opened 2024-05-07
Enable "WIP: Wrap BLIS" with reference LAPACK

jd-foster:
Based on #431, this demonstrates how to get the BLIS extension working using `libblastrampoline` combined with `LAPACK_jll` (based on Reference LAPACK 3.11 https://github.com/Reference-LAPACK/lapack/releases/tag/v3.11)

Note that `suffix_hint="64_",` in `BLAS.lbt_forward` enables ILP versions.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -39,13 +39,14 @@ HYPRE = "b5ffcf37-a2bd-41ab-a3da-4bd9bc8ad771"
 IterativeSolvers = "42fd0dbc-a981-5370-80f2-aaf504508153"
 KernelAbstractions = "63c18a36-062a-441e-b654-da1e3ab1ce7c"
 KrylovKit = "0b1a1467-8014-51b9-945f-bf0ae24f4b77"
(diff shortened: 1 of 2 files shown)


jd-foster:
The MWE as it stands needs to include `LAPACK_jll` now:
```
using LinearSolve, blis_jll, LAPACK_jll

A = rand(4, 4)
b = rand(4)
prob = LinearProblem(A, b)
sol = solve(prob,LinearSolve.BLISLUFactorization())
sol.u
```

jd-foster:
I see now, you're trying to do avoid LBT by going directly like you do with MKL?
https://github.com/SciML/LinearSolve.jl/blob/f5282e549d62eccd67df5f7f72d272366983d034/src/mkl.jl#L7

ChrisRackauckas:
Yes, we'd prefer not to trigger global changes and instead be as local as possible.

jd-foster:
We'll see if we can get https://github.com/JuliaPackaging/Yggdrasil/issues/7660 done then.

PR #511, opened 2024-06-13
Check KrylovKit bump

ChrisRackauckas:


ChrisRackauckas:
Fails due to https://github.com/JuliaArrays/StaticArrays.jl/issues/1261

ChrisRackauckas:
The test that found it was passing last week https://github.com/SciML/LinearSolve.jl/actions/runs/9383265753/job/25836491777

ChrisRackauckas:
At least all of the krylov parts passed.

PR #501, opened 2024-05-13
various CI improvements

ArnoStrouwen:

Issue #308, opened 2023-05-31
Unified handling of preconditoners and factorizations ?


j-fu:
Hi, 

I see that 2.0 is underway, so this may be an occasion to have some discussion about the API. The current API handles preconditioners and direct solvers very differently. IMHO they both are (possibly incomplete) factorizations and could be handeled in a similar way.

OTOH, the current API puts UMFPACKFactorization() and the like and iterative solvers into one basket which seems to be not very logical to me either.

So my view is the following:

- A complete factorization provides a solver: Pardiso, Umfpack,  KLU, Cholesky etc.

- whereas an approximate factorization provides a preconditioner:
  ILU0, Jacobi, AlgebraicMultigrid etc.
  

Iterative algorithms are CG, GMRES etc.  Direct solution essentially corresponds to a "OneStep" algorithm which works together with a complete factorization aka solver.

In that sense, a solve API could look as follows:

```
solve(P::LinearProblem, F::Factorization, I::Iteration)
```

and one could have something like

```
solve(P::LinearProblem, F::Factorization)= solve(P,F,OneStep())
```
(with a possible warning that an incomplete factorization doesn't solve)

As factorization one would provide a preconditioner constructor  instead of a the preconditioner itself. So we could have

```
solve(p, KLUFactorization())
solve(p, ILU0Factorization(), CG())
```

Repeated solve after updating the matrix could update the preconditioner in a similar way as a factorization.

May be this proposal comes too late or doesn't fit for some other reasons, I can live either way (and already coded around this...).

ChrisRackauckas:
This sounds interesting, though v2 is already late and hopefully going out today, so unless you plan on making a PR soon it sounds like a 3.0. I'm not sure it's breaking either since it looks like an extension: we do not handle preconditioners in the interface right now other than passing them along to some solvers, so elevating that to some new preconditioner interface would be possible without even breaking any of the "onestep" method interfaces.

Couldn't we just have `solve(p, ILU0Factorization())` as something that solves with ILU0 and then you can pass `ILU0Factorization()` as a preconditioner choice and it'll know to do that?

oscardssmith:
Closing in favor of https://github.com/SciML/LinearSolve.jl/pull/514

PR #514, opened 2024-06-27
make preconditioners part of the solver rather than a random extra

oscardssmith:
This still needs tests, but I'm putting it up as a PR first to get feedback. The intended goal here is to make the `dolinsolve` functionality in OrdinaryDiffEq of being able to pass in a `precs` function that takes in `A` and possibly some parameters and remakes your preconditioner for the new `A` value.

Here's an example of the new functionality in action:
```
using LinearSolve, Krylov, KrylovPreconditioners, LinearAlgebra, SparseArrays
A, b = sprand(10,10, .5), rand(10);
prob = LinearProblem(A,b)
solver = KrylovJL_CG(precs=(A,p=nothing) -> (BlockJacobiPreconditioner(A, 2), I), ldiv=false)
cache = init(prob, solver)
solve!(cache)
reinit!(cache, A = sprand(10,10, .2), b = rand(10))
solve!(cache)
```


src/common.jl
```diff
@@ -179,6 +193,33 @@ function SciMLBase.init(prob::LinearProblem, alg::SciMLLinearSolveAlgorithm,
     return cache
 end
 
+
+function SciMLBase.reinit!(cache::LinearCache;
```
termi-official, line 197: Just for my understanding here:
1.  Is it intended that the input cache is not modified? I.e. shouldn't we have something like if A !== nothing then cache.A = A and so on?
2. I think the preconditioners are still not updated if we do `cache.A = A`. Is this intended?

oscardssmith: 1. Logic bug. Fix incoming.
2. updating preconditioners with a manual setting of `cache.A` (or worse, mutating `cache.A`). I think it will be required for users to `reinit!` if they want this functionality.

j-fu:
... This is the one functionality I missed with LinearSolve. I had a slightly different view on this:  see https://github.com/SciML/LinearSolve.jl/issues/308 . Unfortunately I had no time to work on this, so thank you  for this PR!

As for the functionality: Sometimes, e.g. in Newton solvers, it makes sense to  update the matrix and keep the preconditioner. The "old" preconditioner may still be good enough for the "new" matrix in the sense that few additional iterations are cheaper than the re-creation of the preconditioner (thinking e.g. of Algebraic multigrid etc.)

May be a kwarg `keep_precs` (with default value false) in `reinit!` can help to achieve this.

oscardssmith:
@j-fu yeah, the only reason I thought about this API is that OrdinaryDiffEq has a (slightly broken very hacky) version of this API already (`dolinsolve`) which someone pointed out to me a few weeks ago that it's slightly broken.

src/common.jl
termi-official: 2. xref https://github.com/SciML/LinearSolve.jl/blob/11441b1ca37ca11ec946da76722c8fb5a2277ae5/src/common.jl#L86-L98

j-fu:
A couple of thoughts for this API extension: 
- What about keeping  the Pl/Pr idiom (as functions) instead of `precs` handling tuples ? IMHO this would cover the vast majority of use cases.

- LinearSolve (and the whole SciML) has this  pattern that we write e.g.`UMFPACKFactorization()` to specify a factorization. In a similar way, users could expect to specify 
```
precs=ILUZero_ILU0()
```
instead of 
```
precs= (A,p=nothing)-> (ilu0(A),I)
```

May one could achieve both ways via a dispatch between `Function` and something like `AbstractPreconditioningAlgorithm`. Also  this would give the possibility to have a couple of "curated" preconditioners predefined (e.g. in package extensions).

oscardssmith:
Yeah, we likely would want to add aliases to create iterative algorithms with builtin preconditioners. My reasoning behind `precs` rather than `Pr`/`Pl` is that if you want both preconditioners, it's likely that they will share computational work. We could make it so you are allowed to pass up to one of `Pr`, `Pl`, or `precs` (where passing `Pr` or `Pl` would set the other one to `I`)

j-fu:
Let's have a chat on this at JuliaCon!

oscardssmith:
This is mostly one that I think I just need to put a couple more hours into to get the Pr/Pl/precs handling uniform across everything. This is my next priority after https://github.com/SciML/OrdinaryDiffEq.jl/pull/229.

docs/src/advanced/custom.md
```diff
@@ -33,7 +33,7 @@ The inputs to the function are as follows:
   - `p`, a set of parameters
   - `newA`, a `Bool` which is `true` if `A` has been modified since last solve
   - `Pl`, left-preconditioner
-  - `Pr`, right-preconditioner
+  - `Pr`, right-preconditionerz
```
termi-official, line 36: Typo?

```suggestion
  - `Pr`, right-preconditioner
```

oscardssmith:
This is now working and tested. I'm planning on merging soon assuming CI is clear.

oscardssmith:
This is failing a lot of tests, but so is master. is this ready to merge?

ChrisRackauckas:
@fredrikekre can you figure out what `LoadError: mpiexec() is deprecated, use the non-do-block form` is?

fredrikekre:
https://github.com/SciML/LinearSolve.jl/pull/501#issuecomment-2109009251 + https://github.com/SciML/LinearSolve.jl/pull/501#issuecomment-2109357906 :)

PR #517, opened 2024-07-28
Default preference for MKL checks if Ryzen

ChrisRackauckas:


chriselrod (approved):
Looks good; 0 runtime cost.

chriselrod:
Actually, hold on, let me see if I can find a good way to query vendor.

chriselrod:
I couldn't find anything in Base.
If you're willing to add a dependency on `CpuId.jl`, you could do something like
```julia
@static if Sys.ARCH === :x86_64
    using CpuId
    if CpuId.cpuvendor() === :Intel
        # using MKL
    else
        # don't
    end
else
    # don't
end
```

ChrisRackauckas:
This is good enough. Besides, it's not vendor anyways, AMD is fine with MKL from my tests, it's just the server chips.

But wait, that wouldn't be Ryzen that would be EPYC, so this may need to change.

chriselrod:
Note that this is actually `znver4`, not `znver3`, but the detection is outdated:
```julia
julia> Sys.CPU_NAME
"znver3"

julia> first(Sys.cpu_info()).model
"AMD EPYC 9354 32-Core Processor"

julia> occursin("EPYC", first(Sys.cpu_info()).model)
true

julia> using CpuId

julia> cpubrand()
"AMD EPYC 9354 32-Core Processor                "

julia> occursin("EPYC", cpubrand())
true
```
`znver*` will detect Ryzen, whether laptop, desktop, or EPYC.
If you want to detect EPYC-only, hopefully the model name will work?
You could use `CpuId.cpubrand()`, or get it from `Sys.cpu_info()` to avoid the dependency.

PR #530, opened 2024-08-22
WIP: Make more libraries into an extension?

ChrisRackauckas:
I think we should be considering some kind of split to LinearSolveCore vs LinearSolve, a la OrdinaryDiffEq. The default algorithm requires RecursiveFactorization and MKL, but we could definitely make a version that doesn't have these deps. But at least sparse stuff can go into an extension since those always dispatch on the existance of sparse matrices.

This PR isn't quite complete yet, more needs to be removed with sparse, but it shows the right direction.


ChrisRackauckas:
@rayegun can you complete this?

rayegun:
There's not actually strong reason to do this for KLU since it's bundled with Julia but yes

ChrisRackauckas:
`juliac`

ChrisRackauckas:
Newer version is https://github.com/SciML/LinearSolve.jl/pull/570

PR #533, opened 2024-08-22
update precs when cache.isfresh

oscardssmith:
successor to https://github.com/SciML/LinearSolve.jl/pull/528

### DIFF
--- a/ext/LinearSolvePardisoExt.jl
+++ b/ext/LinearSolvePardisoExt.jl
@@ -132,6 +132,11 @@ function SciMLBase.solve!(cache::LinearSolve.LinearCache, alg::PardisoJL; kwargs
     @unpack A, b, u = cache
     A = convert(AbstractMatrix, A)
     if cache.isfresh
+        if hasproperty(alg, :precs) && !isnothing(alg.precs)
+            Pl, Pr = cache.alg.precs(x, cache.p)
+            cache.Pl = Pl
+            cache.Pr = Pr
+        end
         phase = alg.cache_analysis ? Pardiso.NUM_FACT : Pardiso.ANALYSIS_NUM_FACT
         Pardiso.set_phase!(cache.cacheval, phase)
         Pardiso.pardiso(cache.cacheval, A, eltype(A)[])
--- a/src/iterative_wrappers.jl
+++ b/src/iterative_wrappers.jl
@@ -226,6 +226,11 @@ end
 
 function SciMLBase.solve!(cache::LinearCache, alg::KrylovJL; kwargs...)
     if cache.isfresh
+        if hasproperty(alg, :precs) && !isnothing(alg.precs)
+            Pl, Pr = cache.alg.precs(x, cache.p)
+            cache.Pl = Pl
+            cache.Pr = Pr
+        end
         solver = init_cacheval(alg, cache.A, cache.b, cache.u, cache.Pl, cache.Pr,
             cache.maxiters, cache.abstol, cache.reltol, cache.verbose,
             cache.assumptions, zeroinit = false)
--- a/src/common.jl
+++ b/src/common.jl
@@ -84,19 +84,8 @@ mutable struct LinearCache{TA, Tb, Tu, Tp, Talg, Tc, Tl, Tr, Ttol, issq, S}
 end
 
 function Base.setproperty!(cache::LinearCache, name::Symbol, x)
-    if name === :A
-        if hasproperty(cache.alg, :precs) && !isnothing(cache.alg.precs)
-            Pl, Pr = cache.alg.precs(x, cache.p)
-            setfield!(cache, :Pl, Pl)
-            setfield!(cache, :Pr, Pr)
-        end
+    if name === :A || name === :p
         setfield!(cache, :isfresh, true)
-    elseif name === :p
-        if hasproperty(cache.alg, :precs) && !isnothing(cache.alg.precs)
-            Pl, Pr = cache.alg.precs(cache.A, x)
-            setfield!(cache, :Pl, Pl)
-            setfield!(cache, :Pr, Pr)
-        end
     elseif name === :b
         # In case there is something that needs to be done when b is updated
         update_cacheval!(cache, :b, x)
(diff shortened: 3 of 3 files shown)


ext/LinearSolvePardisoExt.jl
```diff
@@ -132,6 +132,11 @@ function SciMLBase.solve!(cache::LinearSolve.LinearCache, alg::PardisoJL; kwargs
     @unpack A, b, u = cache
     A = convert(AbstractMatrix, A)
     if cache.isfresh
+        if hasproperty(alg, :precs) && !isnothing(alg.precs)
+            Pl, Pr = cache.alg.precs(x, cache.p)
+            cache.Pl = Pl
+            cache.Pr = Pr
+        end
```
j-fu, line 139: Why is this needed here ? Pl, Pr are not used by Pardiso.

oscardssmith: oh, is krylov the only one that uses them? I thought there was another...

ChrisRackauckas: IterativeSolvers, KrylovKit.

src/iterative_wrappers.jl
```diff
@@ -226,6 +226,11 @@ end
 
 function SciMLBase.solve!(cache::LinearCache, alg::KrylovJL; kwargs...)
     if cache.isfresh
+        if hasproperty(alg, :precs) && !isnothing(alg.precs)
+            Pl, Pr = cache.alg.precs(x, cache.p)
```
j-fu, line 230: Here we may need
```
Pl, Pr = cache.alg.precs(cache.A, cache.p)
```

src/iterative_wrappers.jl
```diff
@@ -226,6 +226,11 @@ end
...
+            cache.Pl = Pl
+            cache.Pr = Pr
+        end
```
j-fu, line 233: And generally, could we have it like this:
```
function SciMLBase.solve!(cache::LinearCache, alg::KrylovJL; reuse_precs=true, kwargs...)
    if cache.isfresh
        if hasproperty(alg, :precs) && !isnothing(alg.precs)
            if reuse_precs || cache.Pl == nothing || cache.Pr == nothing
                Pl, Pr = cache.alg.precs(cache.A, cache.p)
                cache.Pl = Pl
                cache.Pr = Pr
            end
        end
 ...
```

j-fu: Also, in the moment this leads to `precs` being called twice: once in `init` and once here.

ChrisRackauckas: init should set fresh to false.

oscardssmith: ah that's tricky. I think I will need to add an extra flag to fix this.

oscardssmith: @ChrisRackauckas really? I think the code relies on it being true to set up some of the C libraries.

oscardssmith: I guess that possibly could move to init?

oscardssmith: So this is annoyingly complicated. The basic issue is that the LU-based factorizations put "fake" factorizations in at init time (i.e. factorizations of 1x1 matrices) presumably to reduce the time and memory cost. This means that they require `isfresh=true` after init, since during the first solve, they need to resize the arrays and do the real factorization. This sort of strategy can't work for preconditioners, since we don't know ahead of time what type the preconditioner function will return.

The only solution to this that I can think of would be to make the factorization solvers `init` do the actual factorization so when `!cache.isfresh`, all that happens is the ldiv, but this seems sub-optimal as it would make the `init` pretty expensive...

Any other options here?

ChrisRackauckas: >  This sort of strategy can't work for preconditioners, since we don't know ahead of time what type the preconditioner function will return.

Make a query for that part of the API?

oscardssmith: the preconditioner is user generated so unless we want to require users to implement a `precondition_interface` function or something, it's somewhat hard for me to see how this would work.

j-fu: I made a PR to this branch: https://github.com/oscardssmith/LinearSolve.jl/pull/3 which shifts the problem to the question why zeroinit is needed.

ChrisRackauckas: I think we have to impose some kind of interface anyways. There can be a slow fallback but a way to speed it up and some extensions.

j-fu: Generaly I indeed think that preconditioners and factorizations could be handled in the very same way. With a DirectSolver algorithm type one could have e.g. alg=DirectSolver(UMFPACKFactorization) or alg=KrylovJL_CG(ILUPreconiditioner).   However this would be a significantly breaking change now.

ChrisRackauckas: Nows the time to propose and make any significant breaking change to preconditioners.

oscardssmith: would it be an option to make it so `init` just does the factorization? I understand that it might lead to an extra factorization for the ODE case (since they will sometimes do a first factorization with fake values), but it does seem a lot simpler if we can just say that when you init, all the things that only depend on `A` are done, and the cache_isfresh is set to false.

ChrisRackauckas: What's wrong with the current way? The current way is exactly that: `init` "just does the factorization". But that's a fallback. There's a function which by default factorizes, and then it's overloaded for specific factorizations to just build the factorization object as an optimization. It's a small optimization we then add to every method that's actually used, but it's usually not a big deal. That's exactly what I'm proposing here.

oscardssmith: the problem with the current way is that it makes the fallback slow. if we didn't have the "fast path", we could set cache.isfresh=false which would recover the speed of the fast path for all types, as long as the user inits on a matrix they actually want to solve.

ChrisRackauckas: You could just set it to false on the slow path.

oscardssmith: the current approach is better by 1 factorization for users that create a LinearCache with garbage data, but those users will almost always be doing lots of factorizations, minimizing the advantage. OTOH, the current approach has overhead for users who just call solve (since the fake factorization isn't free), or users of the cache form who init on data they want to solve eventually.

oscardssmith: we could set it to false on the slow path, but doing so would be pretty hard since init_cacheval doesn't get the cache.

oscardssmith: and actually users that are initing with fake matrices can init with a 1x1 themselves, which will completely remove the penalty of my suggestion

PR #534, opened 2024-08-23
fix KyrlovJL remake

oscardssmith:
we had this `args` field that was never being used, but did break `remake`.
```
# before:
julia> remake(KrylovJL_GMRES(), precs=nothing).kwargs
pairs(::NamedTuple) with 1 entry:
  :args => ()
# after:
julia> remake(KrylovJL_GMRES(), precs=nothing).kwargs
Base.Pairs{Symbol, Union{}, Tuple{}, @NamedTuple{}}()
```
As a side-note, we should fix the default SciMLBase remake for structs with fields named both `args` and `kwargs`

### DIFF
--- a/src/iterative_wrappers.jl
+++ b/src/iterative_wrappers.jl
@@ -2,7 +2,7 @@
 
 """
 ```julia
-KrylovJL(args...; KrylovAlg = Krylov.gmres!,
+KrylovJL(KrylovAlg = Krylov.gmres!,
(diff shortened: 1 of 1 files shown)


oscardssmith:
I think the proper fix here is actually https://github.com/SciML/SciMLBase.jl/pull/769 (although it does appear that this functionality may not quite be working as intended)

PR #535, opened 2024-08-25
remove the lazy init and set isfresh-false on init

oscardssmith:
as discussed in https://github.com/SciML/LinearSolve.jl/pull/533


ChrisRackauckas:
When initing you might not have real values so you might get errors due to singularities and such

oscardssmith:
Given that the fallbacks have to factorize anyway, this doesn't seem like a problem to me. If the factorizations don't have a `check=false`, IMO that's a bug in the factorization.

ChrisRackauckas:
The point of the ArrayInterface calls is to avoid the factorization and also avoid the issues with factorization failing.

ChrisRackauckas:
What's the status on this? Is it still needed?

PR #538, opened 2024-09-02
Test master

ChrisRackauckas:


wsmoses:
we realized we we'rent throwing errors for llvm insert instructions that were runtime active [aka storing, via an insert instruction, a contant pointer, into a differentiable data structure]

ChrisRackauckas:
So it should've errored before and it just didn't? 😅

ChrisRackauckas:
I guess like the others, LinearProblem might need to be mutable?

ChrisRackauckas:
Bump, the question is:

> I guess like the others, LinearProblem might need to be mutable?

wsmoses:
@ChrisRackauckas can you mark https://github.com/SciML/LinearSolve.jl/blob/33911f60bda7ecd0eeff2d9f75230e48d47503a8/test/enzyme.jl#L178 as const A

The issue yells that https://github.com/SciML/LinearSolve.jl/blob/33911f60bda7ecd0eeff2d9f75230e48d47503a8/test/enzyme.jl#L187 is type unstable, and uses a constant variable (A) being stored into a differentiable LinearProblem. Marking it as const/type stable may optimize out the linear problem, solving the issue

ChrisRackauckas:
Fixing the type stability and marking things constant didn't seem to effect it.

PR #539, opened 2024-09-16
Adapt to pending Enzyme breaking change

wsmoses:


wsmoses:
@ChrisRackauckas this is the proper fix [presumably]

PR #543, opened 2024-09-26
Adapt to pending Enzyme breaking change

avik-pal:


avik-pal:
@wsmoses the current failure seems to come from enzymerules internal https://github.com/SciML/LinearSolve.jl/actions/runs/11049672650/job/30695752840?pr=543#step:6:1178?

wsmoses:
@avik-pal can you see if https://github.com/EnzymeAD/Enzyme.jl/pull/1896 resolves this for you?

Note that this is an error in the error handler itself, but hopefully will say what's wrong with the rule

ext/LinearSolveEnzymeExt.jl
```diff
@@ -25,17 +25,32 @@ function EnzymeCore.EnzymeRules.forward(
         return Duplicated(res, dres)
```
wsmoses, line 25: this code should be removed [the return type from RT]

avik-pal:
@ChrisRackauckas there was an aliasing issue in the rules that had correctness issues. Fixed that, and once the enzyme PR @wsmoses mentioned is tagged, we should merge this. We can deal with the runtime activity later

avik-pal:
Somehow the tests now pass without the upstream patch (or maybe it got released?). I will call that a win and move on...

wsmoses:
so the upstream patch wasn't yet released, it just actually gave you the right error message that you presumably used to fix the underlying problem

ChrisRackauckas:
I'll take it.

Issue #552, opened 2024-10-22
Batch mode


vboussange:
Hey there,
I am trying to optimise the ConScape.jl package, a library for ecological connectivity analysis. The computational bottleneck relates to these lines, where a large linear problem is solved.

https://github.com/ConScape/ConScape.jl/blob/c6d8e3d82a26f181c4f734f5d34e6dcd1b3dc99b/src/gridrsp.jl#L21-L25

I'd love to be able to try some of LinearSolve.jl solvers for this task, but it requires batch mode, a feature that is supported by `\`:
```julia
A = randn(4,4)
B = randn(4,4)
A\B
```
but not by LinearSolve.jl
```julia
prob = LinearProblem(A, B)
sol = solve(prob)
```
I have suggested this feature on Slack in July, but I make it official here. @amontoison raised the point that other SciML users could be interested, and had some idea on how to implement it.
It is not in my capacity to investigate this, but maybe one of the developers here could have a look at it?
Cheers!

ChrisRackauckas:
Yeah we really need to do this at some point.

@avik-pal you setup a batch mode for SimpleGMRES at one time? Is that code still around?

avik-pal:
That is a different kind of batching. It is essentially $A x = b$, where $A$ is `N x M x B` and $b$ is `N x B`, so $x$ is `M x B`. Here we have $A$ -- `N x M` and $b$ is `N x B`

amontoison:
@avik-pal I reviewed SimpleGMRES again, but I’m still not entirely sure what it's doing.
The docstring mentions it’s optimized for block diagonal matrices, but there’s no mention of batching.
If there is batching, I assume it means solving multiple systems `A_i x_i = b_i` simultaneously, if I understand the problem correctly.

It does make sense to combine some operations to leverage BLAS2/BLAS3 instead of BLAS1.
However, this isn’t a “tensor-like” Krylov method, which is a different approach altogether. If all matrices A_i are the same, you should avoid this and instead use a block-Krylov method.
You will directly see the difference in terms of performance?

I quickly opened PR #554 in case it’s helpful.

ChrisRackauckas:
[...]

PR #554, opened 2024-10-28
Interface Block-GMRES from Krylov.jl

amontoison:
#552 
You probably want to use `block_gmres!` instead of `gmres!` by default if the right-hand is an `AbstractMatrix`.
Note that for block-Krylov methods, the solution is stored in `solver.X` and not `solver.x`.


src/simplegmres.jl
```diff
@@ -88,38 +88,7 @@ function update_cacheval!(cache::LinearCache, cacheval::SimpleGMRESCache, name::
...
-        ρ = b / s  # Computationally better than ρ = a / c since |c| ≤ |s|.
-    else
-        t = b / a
-        c = sign(a) / sqrt(one(T) + t * t)
-        s = c * t
-        ρ = a / c  # Computationally better than ρ = b / s since |s| ≤ |c|
-    end
-    return (c, s, ρ)
-end
+_sym_givens(a::T, b::T) where {T <: AbstractFloat} = Krylov.sym_givens(a, b)
```
ChrisRackauckas, line 91: this is misplaced if it's a hard dep.

ChrisRackauckas:
> You probably want to use block_gmres! instead of gmres! by default if the right-hand is an AbstractMatrix.

That's not the right interpretation because it doesn't solve the `vec` form? So it would get the wrong answer?

ChrisRackauckas:
> Otherwise, it's not directly a linear system.

No? In that case `B` is not an element of the vector space that `A` acts on, it's a set of b's in the vector space. It's playing fast and loose which is fine when everything is just an `Array` and gives some computational improvements in a special case but it breaks down when dealing with arbitrary `AbstractArray` types very fast. If you always interpret the right hand side as an element of the vector space on which `A` is solving for then you can expect things to actually work more naturally, and correctness always comes first.

amontoison:
Solving AX = vec(B) and AX = B is completely different, do we agree on that?
The second system can be reformulated as a linear system with a vector as right-hand but the left-hand side is not anymore A but block_diagonal(A, ..., A).
But A, X and B are still linearly related.

It's not the case with `AX = vec(B)`, which is counter intuitive for a package called LinearSolve.jl.

amontoison:
You don't solve a problem with a different operator. I just explained how a linear system with a vector as right-hand and a linear system with a matrix as right-hand side is equivalent.

ChrisRackauckas:
They aren't equivalent, it's overloaded syntax. The latter is actually solving with `block_diagonal(A, ..., A)`. `block_diagonal(A, ..., A) != A`. It's fancy overloaded syntax that lets it work for vectors and matrices but it breaks down really fast when generalizing.

amontoison:
I don't understand why you talk about averloading.
It's just linear algebra.
Solving `AX = B` where `B` has `p` columns is equivalent to solving `p` systems `Ax_i = b_i` where `b_i` is the column `i` of `B`.
All these systems can be merged into one:
`block_diagonal(A, ..., A) [x_1; x2; ...; x_p] = [b_1; ...; b_p]` that we can call `Dy = c`.

But the main idea was to explain that it's completely different than solving AX = vec(B).
A \ B doesn't return the solution of AX = vec(B), so why do you to interprete the problem like this if the right'-hand side is a matrix?

ChrisRackauckas:
This is literally the same thing as taking vector transposes seriously. Yes, obviously it's homomorphic to solving with `block_diagonal(A, ..., A)`. But that's not `A`, and making it work in the general case requires specifying a system for which you have homomorphisms of linear systems. Take it to a separate issue since this would be a major breaking things and there's many other things that would have to be handled in the type system. Also, it would need more general support as sparse solvers generally do not support this.

ChrisRackauckas:
@AJ0070 look into whether this is needed now: with LinearSolve v5 we have the block form, and I think this was completed.

amontoison:
Note that Krylov.jl also has a `block_minres` solver now for symmetric / Hermitian linear systems with multiple right-hand sides.

Issue #573, opened 2025-02-07
Circular dependency on 1.10


j-fu:
**Describe the bug 🐞**

I am getting circular dependency errors with LinearSolve v3 on Julua 1.10 (but not on 1.11 and nightly).
In particular, this makes the Aqua test for persistent tasks fail. 

**Expected behavior**

Aqua test for persistent tasks works.

**Minimal Reproducible Example 👇**

See https://github.com/j-fu/LSolveMWE

Pkg.test the package under 1.10 vs 1.11.

**Error & Stacktrace ⚠️**
This is the test output on 1.10:

```
┌ Warning: Circular dependency detected.
│ Precompilation will be skipped for dependencies in this cycle:
│  ┌ LinearSolve → LinearSolveSparseArraysExt
│  └─ LinearSolve → LinearSolveRecursiveArrayToolsExt
│ Precompilation will also be skipped for the following, which depend on the above cycle:
│   jl_CjwJIzUtOD
│   LSolveMWE
└ @ Pkg.API.Precompilation ~/.julia/juliaup/julia-1.10.8+0.x64.linux.gnu/share/julia/stdlib/v1.10/Pkg/src/precompilation.jl:584
┌ Error: Unexpected error: /tmp/jl_CjwJIzUtOD/done.log was not created, but precompilation exited
└ @ Aqua ~/.julia/packages/Aqua/epbUr/src/persistent_tasks.jl:118
Test Failed at /home/fuhrmann/.julia/packages/Aqua/epbUr/src/persistent_tasks.jl:38
Expression: !(has_persistent_tasks(package; kwargs...))

ERROR: LoadError: There was an error during testing
in expression starting at /home/fuhrmann/Wias/work/julia/test/LSolveMWE/test/runtests.jl:5
ERROR: Package LSolveMWE errored during testing
```

**Environment (please complete the following information):**

See the CI results in  https://github.com/j-fu/LSolveMWE

**Additional context**

@JPThiele

j-fu:
Ok, this seems to be related to something like this:
https://github.com/JuliaLang/julia/issues/52511#issuecomment-2609736748

I updated the MWE to remove the SparseArrays dependency of the project, and
it is still in the manifest:

```
(LSolveMWE) pkg> why SparseArrays
LinearSolve → Krylov → SparseArrays
LinearSolve → LazyArrays → SparseArrays
LinearSolve → SciMLBase → RecursiveArrayTools → Statistics → SparseArrays
[...]

PR #577, opened 2025-02-12
Make RAT a hard dep

ChrisRackauckas:
Fixes https://github.com/SciML/LinearSolve.jl/issues/573


oscardssmith (approved):
Can confirm that this fixes 1.10

PR #582, opened 2025-02-17
Update adjoint of Linear Solve for complex matrices

cmrace:
## Checklist

## Additional context

For complex matrices, `A \ b` and `LinearProblem(A, b)` give different gradients via Zygote. The issue is that Zygote uses `adjoint` 
while LinearSolve uses `transpose`. This should only affect behavior for complex matrices and this PR changes LinearSolve behavior to match Zygote.


ChrisRackauckas:
The tests fail because ForwardDiff cannot handle the complex numbers. It should split that part to keep real numbers in the test with FD

ChrisRackauckas:
Thanks for sticking with this!

Issue #583, opened 2025-02-25
Malformed Project.toml


Datseris:
**Describe the bug 🐞**

I just tried to update my project and I keep getting
```
ERROR: LinearSolve has a malformed Project.toml, the extension package SparseArrays is not listed in [weakdeps] 
```

Is this a problem on my side? I wouldn't guess so. 

The problem is I can't even report what version of LinearSolve is causing this, because `st -m LinearSolve` throws this error as well. I did a full delete of my local Manifest.toml file but the problem persists after trying to update again.

  - Output of `versioninfo()`

```julia
Julia Version 1.10.6
Commit 67dffc4a8a (2024-10-28 12:23 UTC)
Build Info:
  Official https://julialang.org/ release
Platform Info:
  OS: Windows (x86_64-w64-mingw32)
  CPU: 8 × 11th Gen Intel(R) Core(TM) i7-1165G7 @ 2.80GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-15.0.7 (ORCJIT, tigerlake)
Threads: 6 default, 0 interactive, 3 GC (on 8 virtual cores)
Environment:
  JULIA_CONDAPKG_BACKEND = Current
  JULIA_CONDAPKG_ENV = @python
  JULIA_PKG_PRECOMPILE_AUTO = 0
  JULIA_EDITOR = code
  JULIA_NUM_THREADS = 6
```

Datseris:
Do you have any advice of what version I should downgrade to to solve the issue for now? Unforutnately I can't run my project anymore and I don't know where to revert to. I've never used LinearSolve directly.

devmotion:
I'd try LinearSolve@2.39.0.

ChrisRackauckas:
IIUC this is a Julia bug, and if you upgrade to the patch v1.10.8 it should go away? I don't know what causes the bug so it's not clear how to avoid it.

devmotion:
[...]

PR #584, opened 2025-02-25
Make SparseArrays a weakdeps and add it as dependency of Sparspak and Pardiso extensions

devmotion:
See #583. Let's see what happens.


.github/workflows/Downgrade.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        version: ['1']
+        version: ['min']
```
devmotion, line 18:
```suggestion
        version: ['1']
```

Project.toml
```diff
@@ -102,7 +102,7 @@ Pkg = "1"
 PrecompileTools = "1.2"
 Preferences = "1.4"
 Random = "1"
-RecursiveArrayTools = "3.8"
+RecursiveArrayTools = "3.27.2"
```
devmotion, line 105:
```suggestion
RecursiveArrayTools = "3.8"
```

PR #590, opened 2025-03-27
Update ForwardDiff compat

oscardssmith:


oscardssmith:
Now that SciMLBase is released, this should be good to go.

PR #595, opened 2025-04-18
Support Krylov.jl v0.10

amontoison:
Note that I also added a generic API with the latest release of `Krylov.jl` but I don't want to break how you switch the solver: https://jso.dev/Krylov.jl/dev/generic_interface/


ChrisRackauckas (approved):

ChrisRackauckas:
Thanks for helping with the bump!

Issue #607, opened 2025-05-12
ForwardDiff overloads


ChrisRackauckas:
`A(p)x = b(p)` does not need to differentiate w.r.t. `p`, since `dA/dp dx/dp = db/dp` is just another linear equation, so that can be solved. Thus the duals should just be split and it should solve using the same solver. 

The implementation can follow the same setup. You find cases with Dual numbers:

https://github.com/SciML/NonlinearSolve.jl/blob/v4.8.0/lib/NonlinearSolveBase/ext/NonlinearSolveBaseForwardDiffExt.jl#L22-L32

If you catch that, then you `remake` using `ForwardDiff.value` to de-dual and solve the non-dual part:

https://github.com/SciML/NonlinearSolve.jl/blob/v4.8.0/lib/NonlinearSolveBase/ext/NonlinearSolveBaseForwardDiffExt.jl#L43-L59

and then you push forward the partials:

https://github.com/SciML/NonlinearSolve.jl/blob/v4.8.0/lib/NonlinearSolveBase/ext/NonlinearSolveBaseForwardDiffExt.jl#L64-L78

It's actually the same thing as this code, since `f(x,p) = 0` when `x(p)` is the solution has the derivative `f_x x_p + f_p = 0` which means `x_p = - f_x \ f_p` which is this line https://github.com/SciML/NonlinearSolve.jl/blob/v4.8.0/lib/NonlinearSolveBase/ext/NonlinearSolveBaseForwardDiffExt.jl#L66. So it's effectively the same thing as this code right here is just turned into `x_p = A_p \ b_p` instead, i.e. just grabbing the partials out of `A` and `b`.

This is the core of the downstream issue https://github.com/SciML/ModelingToolkit.jl/issues/3589

oscardssmith:
We want to do this for all `A`.

jClugstor:
But `p` is only every used if `A` is a SciMLOperator, right? So if it's not then `A_p` should just be 0?

Also, can `b` actually depend on `p`? I don't see anything like that in the LinearSolve docs.

jClugstor:
> But with autodiff, you're instead calculating `dx/dA` and `dx/db`.

Only if `A` and `b` are Duals / arrays of Duals when using ForwardDiff, right?

Oh, I think I get it, for some reason I was thinking that this was only for when `p` was Dual. 

So the point is that it should also go through the overload if we're differentiating wrt. `A` or `b`? 

If we have 
```julia 
[...]

PR #621, opened 2025-06-03
Overloads for LinearProblems with ForwardDiff Dual numbers

jClugstor:
## Checklist

## Additional context
Addresses https://github.com/SciML/LinearSolve.jl/issues/607

This gives the same result as when Dual numbers go through the actual solver, and can handle cases when either `A` or `b` do not depend on anything, i.e. when only one of `A` or `b` consists of Dual numbers.


jClugstor:
@oscardssmith I need to make sure that the results here are the same as when the Dual numbers actually go through the solver. I tried using invoke, but I can't seem to make it go through the more general solve. Any ideas?

ext/LinearSolveForwardDiffExt.jl
```diff
@@ -0,0 +1,241 @@
...
+using ForwardDiff
+using ForwardDiff: Dual, Partials
+using SciMLBase
+using RecursiveArrayTools
+
+const DualLinearProblem = LinearProblem{
+    <:Union{Number, <:AbstractArray, Nothing}, iip,
+    <:Union{<:Dual{T, V, P}, <:AbstractArray{<:Dual{T, V, P}}},
+    <:Union{<:Dual{T, V, P}, <:AbstractArray{<:Dual{T, V, P}}},
+    <:Union{Number, <:AbstractArray, SciMLBase.NullParameters}
```
oscardssmith: `p` should be allowed to be `Any`

jClugstor:
This is now working when
- `A` and `b` is a Dual or AbstractArray of Duals
- Just `A` is a Dual or AbstractArray of Duals
- Just `b` is a Dual or AbstractArray of Duals

Works with Krylov / iterative methods, including when `u0` is Dual / AbstractArray of Dual (in that case it just uses the primal part of the Dual `u0`, for a LinearSolve the derivative wrt `u0` should be 0 anyway). 

Also works when reusing the cache, e.g. updating both `A` and `b`, or just `A` or just `b`. 

When A is a `SciMLOperator`, the problem isn't a `DualAbstractLinearProblem`, so none of this code is used. 

I copy a couple of things, like the original primal solution and the original `b`. This allows me to reuse the original cache, avoiding refactorization though. 

In order to rebuild the solution with the correct Dual number types, I just store the type in the DualLinearCache and then apply it to the results of the Dual part of the solve, which seems like it could be problematic, but I'm not sure of a better way.

ext/LinearSolveForwardDiffExt.jl
jClugstor: Oh yes, thanks.

ChrisRackauckas:
Oh can you  follow up with 2 small things:

1. Check that KLUFactorization and UMFPACKFactorization new work with dual A. This is not something that would have worked without these overloads.
2. Check that the default algorithm choice is based on the primal types, not the dual types. Since the duals are not sent through the solver anymore, it should just choose the solver that is best for the primal types.

I think both are true/will work, but are missing from the tests here.

PR #629, opened 2025-07-02
ForwardDiff Overload Fixes 

jClugstor:
## Checklist

### DIFF
--- a/ext/LinearSolveForwardDiffExt.jl
+++ b/ext/LinearSolveForwardDiffExt.jl
@@ -36,8 +36,14 @@ const DualAbstractLinearProblem = Union{
 LinearSolve.@concrete mutable struct DualLinearCache
     linear_cache
     dual_type
+
     partials_A
     partials_b
+    partials_u
+
+    dual_A
+    dual_b
+    dual_u
 end
 
 function linearsolve_forwarddiff_solve(cache::DualLinearCache, alg, args...; kwargs...)
@@ -55,16 +61,16 @@ function linearsolve_forwarddiff_solve(cache::DualLinearCache, alg, args...; kwa
 
     rhs_list = xp_linsolve_rhs(uu, ∂_A, ∂_b)
 
-    partial_cache = cache.linear_cache
-    partial_cache.u = dual_u0
-
+    cache.linear_cache.u = dual_u0
+    # We can reuse the linear cache, because the same factorization will work for the partials.
     for i in eachindex(rhs_list)
-        partial_cache.b = rhs_list[i]
-        rhs_list[i] = copy(solve!(partial_cache, alg, args...; kwargs...).u)
+        cache.linear_cache.b = rhs_list[i]
+        rhs_list[i] = copy(solve!(cache.linear_cache, alg, args...; kwargs...).u)
     end
 
-    # Reset to the original `b`, users will expect that `b` doesn't change if they don't tell it to
-    partial_cache.b = primal_b
+    # Reset to the original `b` and `u`, users will expect that `b` doesn't change if they don't tell it to
+    cache.linear_cache.b = primal_b
+    Main.@infiltrate
 
     partial_sols = rhs_list
 
@@ -144,7 +150,6 @@ function SciMLBase.init(
     ∂_A = partial_vals(A)
     ∂_b = partial_vals(b)
 
-    #primal_prob = LinearProblem(new_A, new_b, u0 = new_u0)
     primal_prob = remake(prob; A = new_A, b = new_b, u0 = new_u0)
 
     if get_dual_type(prob.A) !== nothing
@@ -157,7 +162,7 @@ function SciMLBase.init(
         primal_prob, alg, args...; alias = alias, abstol = abstol, reltol = reltol,
         maxiters = maxiters, verbose = verbose, Pl = Pl, Pr = Pr, assumptions = assumptions,
         sensealg = sensealg, u0 = new_u0, kwargs...)
-    return DualLinearCache(non_partial_cache, dual_type, ∂_A, ∂_b)
+    return DualLinearCache(non_partial_cache, dual_type, ∂_A, ∂_b, !isnothing(∂_b) ? zero.(∂_b) : ∂_b, A, b, zero.(b))
 end
 
 function SciMLBase.solve!(cache::DualLinearCache, args...; kwargs...)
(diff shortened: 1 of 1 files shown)


jClugstor:
The overloads won't work with higher order derivatives or nested Duals. We would need a special way to handle higher order derivatives / maybe a general expression for `x_p` for higher order derivatives. So for now I just want to exclude nested Duals from going through this.
So I did this to try to exclude `NestedDualLinearProblem` from going through the overloads, but this is pretty restrictive. But if  I do ` SingleDual{T, V, P} = Dual{T, V, P} where {T, V <:Number , P}` then a NestedDualLinearProblem is also a NestedDualLinearProblem, which defeats the purpose. Essentially I want to make sure that if the value of the Dual is another Dual, go through the normal path and not the ForwardDiffExt path. @oscardssmith I wonder if you have any thoughts on how I could do that?

```
# Define type for non-nested dual numbers
const SingleDual{T, V, P} = Dual{T, V, P} where {T, V <:Float64 , P}

# Define type for nested dual numbers
const NestedDual{T, V, P} = Dual{T, V, P} where {T, V <: Dual, P}

const SingleDualLinearProblem = LinearProblem{
    <:Union{Number, <:AbstractArray, Nothing}, iip,
    <:Union{<:SingleDual, <:AbstractArray{<:SingleDual}},
    <:Union{<:SingleDual, <:AbstractArray{<:SingleDual}},
    <:Any
} where {iip}

const NestedDualLinearProblem = LinearProblem{
    <:Union{Number, <:AbstractArray, Nothing}, iip,
    <:Union{<:NestedDual, <:AbstractArray{<:NestedDual}},
    <:Union{<:NestedDual, <:AbstractArray{<:NestedDual}},
    <:Any
} where {iip}

const DualALinearProblem = LinearProblem{
    <:Union{Number, <:AbstractArray, Nothing},
    iip,
    <:Union{<:SingleDual, <:AbstractArray{<:SingleDual}},
    <:Union{Number, <:AbstractArray},
    <:Any
} where {iip}

const DualBLinearProblem = LinearProblem{
    <:Union{Number, <:AbstractArray, Nothing},
    iip,
    <:Union{Number, <:AbstractArray},
    <:Union{<:SingleDual, <:AbstractArray{<:SingleDual}},
    <:Any
} where {iip}

const DualAbstractLinearProblem = Union{
    SingleDualLinearProblem, DualALinearProblem, DualBLinearProblem}
```

oscardssmith:
I think making the higher order/nested cases work out shouldn't be too hard. Derivatives are a linear operator, so I think everything should compose nicely here.

ext/LinearSolveForwardDiffExt.jl
```diff
@@ -9,7 +9,7 @@ using RecursiveArrayTools
 
 
 # Define type for non-nested dual numbers
-const SingleDual{T, V, P} = Dual{T, V, P} where {T, V <:Float64 , P}
+const SingleDual{T, V, P} = Dual{T, V, P} where {T, V <:Union{Float64, Float32} , P}
```
oscardssmith:
```suggestion
const SingleDual{T, V, P} = Dual{T, V, P} where {T, V <: AbstractFloat , P}
```

jClugstor:
@oscardssmith this actually does work recursively, I just had to make a few adjustments first. I'll test it with more nested Dual stuff but this should be good pretty soon actually.

oscardssmith:
that's very odd. It's unclear to me why Radau would be any different here.

ChrisRackauckas:
Complex numbers?

ext/LinearSolveForwardDiffExt.jl
```diff
@@ -1,43 +1,64 @@
...
     <:Any
-} where {iip, T, V, P}
+} where {iip}
+
+const NestedDualLinearProblem = LinearProblem{
+    <:Union{Number, <:AbstractArray, Nothing}, iip,
+    <:Union{<:NestedDual, <:AbstractArray{<:NestedDual}},
+    <:Union{<:NestedDual, <:AbstractArray{<:NestedDual}},
+    <:Any
+} where {iip}
```
oscardssmith: is any of this actually necessary?

jClugstor: Not anymore, this should be able to be condensed down now.

jClugstor:
Yes, seems likely there's an issue there. Although it's strange because in the case of `Complex{Dual{ ...` values in A or b,  the overloads aren't even being gone through, because then it isn't a `DualLinearProblem`.

jClugstor:
Maybe a silly question, but when we call `dolinsolve` (from OrdinaryDiffEqDifferentiation) like this 
```julia
  linres1 = dolinsolve(integrator, linsolve1; A = W1, b = _vec(ubuff),
                linu = _vec(dw1))
```
we want `dw1` to end up with the same value as `linres1.u`, right?

oscardssmith:
yeah this relies on some weird aliasing.

ext/LinearSolveForwardDiffExt.jl
```diff
@@ -236,7 +236,7 @@ nodual_value(x::Dual{T, V, P}) where {T, V <: Dual, P} = x.value  # Keep the inn
 nodual_value(x::AbstractArray{<:Dual}) = map(nodual_value, x)
 
 
-function partials_to_list(partial_matrix::Vector)
+function partials_to_list(partial_matrix::AbstractArray{T, 1}) where {T}
```
oscardssmith: `AbstractVector{T}`?

jClugstor:
Ok, assuming that the MTK integration test is green, this all works with support for 

- Nested Duals, so Hessian / Jacobian of gradient works
- Sparse / Float only solvers work, including KLUFactorization and UMFPACKFactorization
- anything aliased with cache.u should also be updated when the cache is solved, preserving previous behavior
- Default solvers are chosen based on the type of the lowest level number i.e. the number type that actually goes through the linear solver

But 
- this does not work with problems where values are `Complex{Dual`. I don't think it would be difficult to implement, but for now those type of problems will just have the old behavior and have the Duals go through the solvers, so it won't change anything

ext/LinearSolveForwardDiffExt.jl
```diff
@@ -36,12 +37,19 @@ const DualAbstractLinearProblem = Union{
...
+    partials_u
+
+    dual_A
+    dual_b
+    dual_u
 end
 
 function linearsolve_forwarddiff_solve(cache::DualLinearCache, alg, args...; kwargs...)
     # Solve the primal problem
+    @info "here"
```
ChrisRackauckas:
```suggestion
```

Issue #630, opened 2025-07-11
Trivial `LinearProblem` retcodes with `Default`


AayushSabharwal:
**Describe the bug 🐞**

A trivial 1x1 linear problem successfully solves but retcodes with `Default`

**Expected behavior**

A successful retcode.

**Minimal Reproducible Example 👇**

Without MRE, we would only be able to help you to a limited extent, and attention to the issue would be limited. to know more about MRE refer to [wikipedia](https://en.wikipedia.org/wiki/Minimal_reproducible_example) and [stackoverflow](https://stackoverflow.com/help/minimal-reproducible-example).

```julia
julia> prob = LinearProblem([1.0;;], [1.0])
LinearProblem. In-place: true
b: 1-element Vector{Float64}:
 1.0

julia> sol = solve(prob)
retcode: Default
u: 1-element Vector{Float64}:
 1.0
```

**Environment (please complete the following information):**

  - Output of `using Pkg; Pkg.status()`

```julia
Status `/private/var/folders/8b/tx4_63kd6c51swrwfwj8g3mr0000gn/T/jl_lw6qMv/Project.toml`
  [7ed4a6bd] LinearSolve v3.18.2
```

  - Output of `using Pkg; Pkg.status(; mode = PKGMODE_MANIFEST)`

```julia
Status `/private/var/folders/8b/tx4_63kd6c51swrwfwj8g3mr0000gn/T/jl_lw6qMv/Manifest.toml`
  [47edcb42] ADTypes v1.15.0
  [7d9f7c33] Accessors v0.1.42
  [79e6a3ab] Adapt v4.3.0
  [4fba245c] ArrayInterface v7.19.0
  [4c555306] ArrayLayouts v1.11.1
  [d360d2e6] ChainRulesCore v1.25.2
  [38540f10] CommonSolve v0.2.4
  [34da2185] Compat v4.17.0
  [a33af91c] CompositionsBase v0.1.2
  [2569d6c7] ConcreteStructs v0.2.3
  [187b0558] ConstructionBase v1.6.0
  [a8cc5b0e] Crayons v4.1.1
  [9a962f9c] DataAPI v1.16.0
  [e2d170a0] DataValueInterfaces v1.0.0
  [ffbed154] DocStringExtensions v0.9.5
  [4e289a0a] EnumX v1.0.5
  [e2ba6199] ExprTools v0.1.10
  [55351af7] ExproniconLite v0.10.14
[...]

Issue #532, opened 2024-08-22
Factorization algorithms return `ReturnCode.Default` when it should be `ReturnCode.Success`


vpuri3:
```julia
julia> prob = LinearProblem(rand(4, 4), rand(4)); sol = solve(prob, LUFactorization())
retcode: Default
u: 4-element Vector{Float64}:
  0.0561509562131119
  2.8927484238929284
 -2.005290617428258
 -1.0147124069649773

julia> prob.A * sol.u - prob.b
4-element Vector{Float64}:
  1.1102230246251565e-16
  1.1102230246251565e-16
 -2.220446049250313e-16
  0.0
```

ChrisRackauckas:
Well it should check the retcode on the linear solver first, but yes it should report success if successful.

PR #631, opened 2025-07-14
Fix successful return codes for factorization methods

ChrisRackauckas:
Fixes https://github.com/SciML/LinearSolve.jl/issues/630 fixes https://github.com/SciML/LinearSolve.jl/issues/532

### DIFF
--- a/test/retcodes.jl
+++ b/test/retcodes.jl
@@ -14,7 +14,6 @@ alglist = (
 
 @testset "Success" begin
     for alg in alglist
-        @show alg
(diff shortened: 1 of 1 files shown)


test/retcodes.jl
```diff
@@ -14,12 +14,13 @@ alglist = (
 
 @testset "Success" begin
     for alg in alglist
+        @show alg
```
ChrisRackauckas:
```suggestion
```

PR #650, opened 2025-07-28
Add test for ForwardDiff with StaticArrays 

jClugstor:
## Checklist

## Additional context

We should check to make sure that the ForwardDiff overloads don't apply to StaticArrays.

### DIFF
--- a/src/common.jl
+++ b/src/common.jl
@@ -338,6 +338,7 @@ function SciMLBase.solve(prob::StaticLinearProblem,
         alg, u, nothing, prob; retcode = ReturnCode.Success)
 end
 
+# Here to make sure that StaticLinearProblems with Dual elements don't create a Dual linear cache
 function SciMLBase.init(prob::StaticLinearProblem, alg::SciMLLinearSolveAlgorithm,
         args...;
         alias = LinearAliasSpecifier(),
--- a/src/common.jl
+++ b/src/common.jl
@@ -337,3 +337,117 @@ function SciMLBase.solve(prob::StaticLinearProblem,
     return SciMLBase.build_linear_solution(
         alg, u, nothing, prob; retcode = ReturnCode.Success)
 end
+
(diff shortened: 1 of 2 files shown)


oscardssmith (approved):

ChrisRackauckas (approved):

ChrisRackauckas:
Shit that was a misclick, you should never `init` a static problem.

jClugstor:
> you should never `init` a static problem.

But that's what happens in the fallback case? 
https://github.com/SciML/LinearSolve.jl/blob/51ff7a10acfef4c9099091279357ab031468ccc8/src/common.jl#L334

If we want to completely disallow `init` for a StaticLinearProblem, I can set up an error for that fallback case, and an error for `init(::StaticLinearProblem ...)`, but wouldn't that be breaking?

jClugstor:
If you do something like this (not saying this is particularly smart or common) : 
```julia
using LinearSolve
using StaticArrays
using Krylov

A = @SMatrix rand(4,4)
b = @SVector rand(4)

prob = LinearProblem(A,b)

sol = solve(prob, KrylovJL_GMRES())
```
it goes through `init` because of the fallback here:
https://github.com/SciML/LinearSolve.jl/blob/51ff7a10acfef4c9099091279357ab031468ccc8/src/common.jl#L318-L336 .

So if you try to differentiate something like this: 
```julia
function f(p)
    A = rand(4,4) .+ p
    b = rand(4) .- p
    A = SMatrix{4,4}(A)
    b = SVector{4}(b)
    prob = LinearProblem(A,b)
    
    sol = solve(prob, KrylovJL_GMRES())
end

ForwardDiff.jacobian(f, [2.0])
```
it ends up going through `init`, creating a `DualLinearCache`, and then fails during `solve!`. I mean this doesn't work on previous version either so I guess it probably didn't break anything.

ChrisRackauckas:
Krylov doesn't make sense with static arrays. Any other case?

PR #680, opened 2025-08-06
Potentially fix triangular solve of MKL?

ChrisRackauckas:


ChrisRackauckas:
Tested locally to not wait for CI. Though it's very odd this could slip by before... The basictests.jl should fail if the triangular solve is off but they didn't in the previous merges so there is still something I don't understand here.

ChrisRackauckas:
Oh true. Hmm We might need to force alternative hardware on here so that MKL is hit. I used to always be testing that locally, but now I do more dev on my Mac so that is under tested 😅. I'll setup some self-hosted hardware to make sure this stays stable.

ChrisRackauckas:
This package also directly uses `MKLLUFactorization()`, so I'm not sure why that case wouldn't've done it 😅

PR #685, opened 2025-08-07
Make `GenericLUFactorization` bypass overloads when used on DualLinear Problems

jClugstor:
## Checklist

## Additional context

Needed so that GenericalLUFactorization on DualLinearProblems can go through the normal init. 
It's a little bit clunky but I don't think there's another way to exclude specific algorithms from going through the Dual version of `init`. Plus this way is explicit about exactly which algorithms will go through the Dual overloads.

### DIFF
--- a/test/forwarddiff_overloads.jl
+++ b/test/forwarddiff_overloads.jl
@@ -192,4 +192,4 @@ backslash_x_p = A \ b
 A, b = h([ForwardDiff.Dual(5.0, 1.0, 0.0), ForwardDiff.Dual(5.0, 0.0, 1.0)])
 
 prob = LinearProblem(A, b)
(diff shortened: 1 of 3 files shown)


test/forwarddiff_overloads.jl
```diff
@@ -87,13 +87,6 @@ backslash_x_p = A \ new_b
 @test ≈(x_p, backslash_x_p, rtol = 1e-9)
 
 # Nested Duals
-function h(p)
```
ChrisRackauckas, line 90: why was this removed?

jClugstor: It was a duplicate definition, it's first defined up above.

jClugstor:
Looks like some precompilation issues are happening? I can't seem to replicate the failures locally.

PR #719, opened 2025-08-11
Revert timeout handling in autotune to use maxtime blocking

ChrisRackauckas-Claude:
Reverts the recent timeout handling to use a simpler maxtime-based blocking mechanism


ChrisRackauckas-Claude:
Fixed the blocking logic issue! 

The problem was that the comparison was using `>` instead of `>=` when checking if an algorithm should be skipped. This meant that when an algorithm exceeded maxtime at size N, it would still run again at size N in future runs.

**The fix:**
- Changed line 150 from `if n > blocked_at_size` to `if n >= blocked_at_size`
- Updated variable name from `max_allowed_size` to `blocked_at_size` for clarity
- Fixed comments to accurately describe the behavior

**Testing:**
Created unit tests that verify the blocking logic works correctly for:
- Sizes smaller than the blocked size (should NOT skip) ✓
- Size equal to the blocked size (should skip - this was the bug) ✓  
- Sizes larger than the blocked size (should skip) ✓

The fix ensures that once an algorithm exceeds maxtime at a given matrix size, it will be properly skipped for that size and all larger sizes in subsequent benchmarking runs.

Project.toml
```diff
@@ -15,6 +15,7 @@ Krylov = "ba0b0d4f-ebba-5204-a429-3ac8c609bfb7"
 LazyArrays = "5078a376-72f3-5289-bfd5-ec5146d43c02"
 Libdl = "8f399da3-3557-5675-b5ff-fb832c97cbdb"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
+LinearSolveAutotune = "67398393-80e8-4254-b7e4-1b9a36a3c5b6"
```
```suggestion
```

PR #745, opened 2025-08-20
Add OpenBLASLUFactorization implementation

ChrisRackauckas-Claude:
## Summary
- Adds `OpenBLASLUFactorization` as a new solver that directly calls OpenBLAS_jll
- Provides an alternative to MKLLUFactorization that doesn't require MKL
- Implements the same interface and functionality as MKLLUFactorization

## Implementation Details

This PR introduces `OpenBLASLUFactorization`, which mirrors the existing `MKLLUFactorization` but uses OpenBLAS_jll directly instead of MKL. Key features:

- **Direct OpenBLAS calls**: Uses OpenBLAS_jll directly without going through libblastrampoline for optimal performance
- **Pre-allocated workspace**: Avoids allocations during solving, following the same pattern as MKL implementation  
- **Type support**: Handles Float32, Float64, ComplexF32, and ComplexF64
- **Consistent interface**: Follows the same AbstractFactorization interface as other solvers

## Testing

The implementation has been tested locally with all supported types and produces identical results to the standard LUFactorization:

```julia
julia> A = rand(10, 10); b = rand(10);
julia> prob = LinearProblem(A, b);
julia> sol_openblas = solve(prob, OpenBLASLUFactorization());
julia> sol_default = solve(prob, LUFactorization());
julia> norm(sol_openblas.u - sol_default.u)
2.7755575615628914e-17
```

## Use Case

This solver is particularly useful for:
- Systems where MKL is not available or desired
- Environments that prefer open-source BLAS implementations
- Benchmarking and comparison between different BLAS backends


src/LinearSolve.jl
```diff
@@ -59,6 +59,10 @@ else
     const usemkl = false
 end
 
+# OpenBLAS_jll is a standard library, always available
+using OpenBLAS_jll
```
ChrisRackauckas:
```suggestion
using OpenBLAS_jll: OpenBLAS_jll
```

ChrisRackauckas-Claude:
## Latest Updates

Added preference system for OpenBLAS loading as requested:

- Added `LoadOpenBLAS_JLL` preference with default value `true`
- Users can now disable OpenBLAS via preferences, matching the MKL pattern
- The implementation follows the exact same pattern as MKL:
  
```julia
if Preferences.@load_preference("LoadOpenBLAS_JLL", true)
    using OpenBLAS_jll: OpenBLAS_jll
    const useopenblas = OpenBLAS_jll.is_available()
else
    const useopenblas = false
end
```

This allows users to disable OpenBLAS if they prefer to use their system BLAS or another implementation.

The implementation is now feature-complete and matches the MKL implementation pattern in all aspects:
- Direct BLAS calls bypassing libblastrampoline
- Pre-allocated workspace to avoid allocations
- Preference system for conditional loading
- Full integration with the test suite
- Documentation included

PR #750, opened 2025-08-20
Remove LazyArrays.jl dependency

ChrisRackauckas-Claude:
## Summary
This PR removes the LazyArrays.jl dependency from LinearSolve.jl, simplifying the codebase as suggested.

## Changes
- Removed LazyArrays from Project.toml dependencies (both main dependency and compat entry)
- Replaced `BroadcastArray(@~ .-(λ .* tu))` with simple broadcast `.-(λ .* tu)` in `src/adjoint.jl:87`
- Removed LazyArrays import from `src/LinearSolve.jl`
- Updated tests in `test/adjoint.jl` to check for `AbstractMatrix` instead of `BroadcastArray`
- Removed LazyArrays import from test file

## Rationale
The LazyArrays dependency was only used in one place (`src/adjoint.jl:87`) to create a lazy broadcast array. The regular Julia broadcast operation is sufficient for this use case and doesn't require the additional dependency. This simplification reduces the dependency footprint of the package without any loss of functionality.

## Test Results
All adjoint tests pass successfully with the simplified implementation.


ChrisRackauckas:
@avik-pal I assume this is overkill if Reactant is used? Because it brings in some unwanted dependencies so we kind of need to get rid of it.

avik-pal:
It was https://github.com/SciML/LinearSolve.jl/commit/fa0cd3468e5f37fd87a7603a3315a4aa8928ca41 (not me)

From discourse it seems to be the gemm ordering (I have a half finished PR for this in reactant).

ChrisRackauckas-Claude:
## Additional fix

Found an additional issue: the `ext/LinearSolveMooncakeExt.jl` extension also used LazyArrays for `BroadcastArray` in the `solve!` adjoint rule. Updated it to use regular broadcast like the main adjoint code.

The fix is pushed and Mooncake tests pass locally now.

cc @ChrisRackauckas

PR #753, opened 2025-08-21
Add 32-bit mixed precision solvers for OpenBLAS and RecursiveFactorization

ChrisRackauckas-Claude:
## Summary
- Adds `OpenBLAS32MixedLUFactorization` for mixed precision LU factorization using OpenBLAS
- Adds `RF32MixedLUFactorization` for mixed precision LU factorization using RecursiveFactorization.jl
- Includes comprehensive test coverage for both new solvers

## Motivation
This PR extends LinearSolve.jl's mixed precision solver offerings by adding 32-bit mixed precision implementations for OpenBLAS and RecursiveFactorization, complementing the existing MKL and AppleAccelerate mixed precision solvers. These solvers provide significant performance benefits for memory-bandwidth limited problems while maintaining acceptable accuracy for many use cases.

### OpenBLAS32MixedLUFactorization
- Performs LU factorization in Float32 precision using OpenBLAS routines
- Automatically converts Float64 inputs to Float32 for factorization
- Converts results back to Float64 for the solution
- Supports both real and complex matrices (Float64/ComplexF64 → Float32/ComplexF32)

### RF32MixedLUFactorization
- Leverages RecursiveFactorization.jl's optimized blocking strategies in Float32 precision  
- Particularly effective for small to medium matrices (< 500×500)
- Supports pivoting options (can disable for additional speed at cost of stability)
- Threading support for multi-core performance

## Performance Benefits
- Can provide ~2x speedup for memory-bandwidth limited problems
- Reduces memory usage during factorization
- Particularly beneficial for:
  - Large systems where memory bandwidth is the bottleneck
  - Iterative algorithms where moderate precision is acceptable
  - Problems with well-conditioned matrices

## Testing
- Added comprehensive tests in `test/test_mixed_precision.jl`
- Tests verify:
  - Successful factorization and solve
  - Reasonable accuracy compared to full precision (relative error < 1e-5)
  - Support for both real and complex matrices
  - Proper handling when libraries are not available

## Compatibility
- Follows the same API pattern as existing mixed precision solvers
- No breaking changes
- Gracefully handles cases where OpenBLAS or RecursiveFactorization are not available

## Example Usage
```julia
using LinearSolve, RecursiveFactorization

A = rand(1000, 1000)
b = rand(1000)
prob = LinearProblem(A, b)

# OpenBLAS mixed precision
sol1 = solve(prob, OpenBLAS32MixedLUFactorization())

# RecursiveFactorization mixed precision
sol2 = solve(prob, RF32MixedLUFactorization())

# Without pivoting for maximum speed (less stable)
sol3 = solve(prob, RF32MixedLUFactorization(pivot=Val(false)))
```


test/test_mixed_precision.jl
```diff
@@ -1,8 +1,11 @@
 using Test
-using LinearSolve
 using LinearAlgebra
 using Random
 
+# Load LinearSolve with the working directory
+push!(LOAD_PATH, joinpath(@__DIR__, ".."))
```
ChrisRackauckas, line 6:
```suggestion
```

PR #775, opened 2025-09-09
Fix handling for default algorithms for DualLinearProblems

jClugstor:
## Checklist

## Additional context

This makes it so that the default alg is chosen based on the primal A and b _before_ going to make the Dual cache. This also means that for small problems GenericLU will be used, and the cache will be a normal LinearCache.

### DIFF
--- a/ext/LinearSolveForwardDiffExt.jl
+++ b/ext/LinearSolveForwardDiffExt.jl
@@ -1,7 +1,7 @@
 module LinearSolveForwardDiffExt
 
 using LinearSolve
-using LinearSolve: SciMLLinearSolveAlgorithm, __init
+using LinearSolve: SciMLLinearSolveAlgorithm, __init, DefaultLinearSolver, DefaultAlgorithmChoice, defaultalg
 using LinearAlgebra
 using ForwardDiff
 using ForwardDiff: Dual, Partials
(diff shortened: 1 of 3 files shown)


ext/LinearSolveForwardDiffExt.jl
```diff
@@ -129,6 +129,24 @@ function SciMLBase.init(prob::DualAbstractLinearProblem, alg::GenericLUFactoriza
     return __init(prob,alg, args...; kwargs...)
 end
 
+function SciMLBase.init(prob::DualAbstractLinearProblem, alg::DefaultLinearSolver, args...; kwargs...)
+    if alg.alg === DefaultAlgorithmChoice.GenericLUFactorization
+        return __init(prob, alg, args...; kwargs...)
+    else
+        return __dual_init(prob, alg, args...; kwargs...)
+    end
```
ChrisRackauckas, line 204: This needs a JET test

jClugstor: Tests added. Looks good on 1.11 but fails on LTS

jClugstor:
CI is having some issues here, some are complaining about Pkg not being available, and one says that Julia isn't a recognized executable.

jClugstor:
@ChrisRackauckas I'm not sure what to do here, the JET tests pass for 1.11, but not LTS, so whether I mark the test as broken or not there will still be a test failure.

oscardssmith:
can you mark it broken with a `Version` check?

oscardssmith:
@ChrisRackauckas I think this is now good to go.

ChrisRackauckas (approved):

PR #776, opened 2025-09-10
Performance fixes for DualLinearProblems

jClugstor:
## Checklist

## Additional context

### DIFF
--- a/ext/LinearSolveForwardDiffExt.jl
+++ b/ext/LinearSolveForwardDiffExt.jl
@@ -239,7 +239,6 @@ function __dual_init(
     partials_b_list = !isnothing(∂_b) ? partials_to_list(∂_b) : nothing
 
     # Determine size and type for rhs_list
-    n_partials = 0
     if !isnothing(partials_A_list)
         n_partials = length(partials_A_list)
         rhs_list = [similar(non_partial_cache.b) for _ in 1:n_partials]
--- a/ext/LinearSolveForwardDiffExt.jl
+++ b/ext/LinearSolveForwardDiffExt.jl
@@ -48,6 +48,7 @@ LinearSolve.@concrete mutable struct DualLinearCache{DT}
     # Cached intermediate values for calculations
     rhs_list
     dual_u0_cache
+    primal_u_cache
     primal_b_cache
 
     dual_A
@@ -60,6 +61,7 @@ function linearsolve_forwarddiff_solve!(cache::DualLinearCache, alg, args...; kw
     cache.dual_u0_cache .= cache.linear_cache.u
     sol = solve!(cache.linear_cache, alg, args...; kwargs...)  
 
+    cache.primal_u_cache .= cache.linear_cache.u
     cache.primal_b_cache .= cache.linear_cache.b
     uu = sol.u
 
@@ -77,12 +79,13 @@ function linearsolve_forwarddiff_solve!(cache::DualLinearCache, alg, args...; kw
     cache.linear_cache.u .= cache.dual_u0_cache
     # We can reuse the linear cache, because the same factorization will work for the partials.
     for i in eachindex(rhs_list)
-        cache.linear_cache.b .= rhs_list[i]
+        cache.linear_cache.b = copy(rhs_list[i])
         rhs_list[i] .= solve!(cache.linear_cache, alg, args...; kwargs...).u
     end
 
     # Reset to the original `b` and `u`, users will expect that `b` doesn't change if they don't tell it to
     cache.linear_cache.b .= cache.primal_b_cache
+    cache.linear_cache.u .= cache.primal_u_cache
 
     return primal_sol
 end
@@ -238,6 +241,7 @@ function __dual_init(
         partials_A_list,
         partials_b_list,
         rhs_list,
+        similar(new_b),
         similar(new_b),
         similar(new_b),
         A,
--- a/ext/LinearSolveForwardDiffExt.jl
+++ b/ext/LinearSolveForwardDiffExt.jl
@@ -41,6 +41,14 @@ LinearSolve.@concrete mutable struct DualLinearCache{DT}
     partials_b
     partials_u
 
+    # Cached lists of partials to avoid repeated allocations
+    partials_A_list
+    partials_b_list
+
+    # Cached intermediate values for calculations
+    Auu_list
+    rhs_list
+
     dual_A
     dual_b
     dual_u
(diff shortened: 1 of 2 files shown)


ext/LinearSolveForwardDiffExt.jl
```diff
@@ -65,46 +76,83 @@ function linearsolve_forwarddiff_solve(cache::DualLinearCache, alg, args...; kwa
...
+    # Compute A*uu products using cached storage
+    for i in eachindex(A_list)
+        mul!(cache.Auu_list[i], A_list[i], uu)
+    end
 
-    Auu = [A * uu for A in A_list]
+    # Compute rhs using cached storage (negative of A*uu)
+    for i in eachindex(A_list)
+        cache.rhs_list[i] .= .-cache.Auu_list[i]
+    end
```
oscardssmith: can't you use 5 arg `mul!` with to do the negation for you without using an extra cache?

jClugstor: Oh yeah, probably. Smart!

debug_dual.jl
```diff
@@ -0,0 +1,29 @@
+using LinearSolve
```
oscardssmith: this file probably wasn't meant to be committed, right? (or should it be a test?)

ext/LinearSolveForwardDiffExt.jl
```diff
@@ -41,70 +41,104 @@ LinearSolve.@concrete mutable struct DualLinearCache{DT}
...
-        iters = recursivecopy(sol.iters),
-        stats = recursivecopy(sol.stats)
-    )
 
-    # Solves Dual partials separately 
+    cache.primal_b_cache .= cache.linear_cache.b
+    cache.primal_u_cache .= cache.linear_cache.u
+
+    # Store solution metadata without copying - we'll return this
+    primal_sol = sol
```
oscardssmith: delete this and just return `sol`?

ext/LinearSolveForwardDiffExt.jl
```diff
@@ -247,7 +324,43 @@ partial_vals(x) = nothing
...
+    p = eachindex(first(partial_matrix))
+    for i in p
+        for j in eachindex(partial_matrix)
+            list_cache[i][j] = partial_matrix[j][i]
+        end
+    end
+    return list_cache
+end
+
+function update_partials_list!(partial_matrix, list_cache)
```
oscardssmith, line 371: these functions seem to be just shufling data back and forth. Can you not do this without the shuffling?

debug_dual.jl
jClugstor: Yeah this wasn't meant to be committed

jClugstor:
FYI these changes also broke the nested dual number handling so I'm trying to figure that out

ext/LinearSolveForwardDiffExt.jl
jClugstor: The issue is that we need to take a matrix of ForwardDiff.Partials and convert them in to a list of matrices where each matrix holds corresponding entries of the Partials. We need to do this in order to have matrices that `mul!` and LinearSolve can actually act on, because arithmetic isn't defined for ForwardDiff.Partials. So I'm not sure if there's a way to accomplish that without this.

jClugstor:
Nested Duals are back to working

ext/LinearSolveForwardDiffExt.jl
jClugstor: I guess one thing we could do is precompute the list of the necessary RHS matrices during `init`, since that will only need to be computed once, but would need to be recomputed if A or b change

ChrisRackauckas:
what's left here?

ext/LinearSolveForwardDiffExt.jl
```diff
@@ -114,10 +165,26 @@ end
...
-    # Handle single-level duals for arrays
-    partials_list = RecursiveArrayTools.VectorOfArray(partials)
-    return map(((uᵢ, pᵢ),) -> DT(uᵢ, Partials{N,V}(NTuple{N,V}(pᵢ))),
-        zip(u, partials_list[i, :] for i in 1:length(partials_list.u[1])))
+    # Optimized in-place version that reuses cache.dual_u
+    linearsolve_dual_solution!(getfield(cache, :dual_u), u, partials)
+    return getfield(cache, :dual_u)
+end
+
+function linearsolve_dual_solution!(dual_u::AbstractArray{DT}, u::AbstractArray, partials) where {T, V, N, DT <: Dual{T,V,N}}
```
oscardssmith, line 173: I still think this function shouldn't be necessary.

ext/LinearSolveForwardDiffExt.jl
oscardssmith: is there not a `reinterpret` that does what we want here?

ext/LinearSolveForwardDiffExt.jl
jClugstor: How so? We need a way to take the results of all the linear solves and wrap them back up in to the same type as the original dual numbers, which is all this does.

ext/LinearSolveForwardDiffExt.jl
jClugstor: That might be possible, `ForwardDiff.Partials` is really just an NTuple, and I guess as long as the entries are isbits it should work?

jClugstor: It would be reinterpreting something like this:
```julia 
3×3 Matrix{ForwardDiff.Partials{2, Float64}}:
 Partials(1.0, 0.0)   Partials(0.0, 1.0)  Partials(0.0, 75.0)
 Partials(3.0, 0.0)   Partials(1.0, 0.0)  Partials(5.0, 5.0)
 Partials(0.0, 10.0)  Partials(9.0, 0.0)  Partials(0.0, 1.0)
```
into an array of matrices:
```julia
infil> cache.partials_A_list
2-element Vector{Matrix{Float64}}:
 [1.0 0.0 0.0; 3.0 1.0 5.0; 0.0 9.0 0.0]
 [0.0 1.0 75.0; 0.0 0.0 5.0; 10.0 0.0 1.0]
 
 
infil> cache.partials_A_list[1]
3×3 Matrix{Float64}:
 1.0  0.0  0.0
 3.0  1.0  5.0
 0.0  9.0  0.0
 
 infil> cache.partials_A_list[2]
3×3 Matrix{Float64}:
  0.0  1.0  75.0
  0.0  0.0   5.0
 10.0  0.0   1.0
```

jClugstor: We can do 
```julia
partials_A_reinterpreted = reinterpret(reshape, eltype(first(∂_A)), ∂_A) 
view(partials_A_reinterpreted, k, :, :)
```
Will give us the kth partials matrix that we need. 

But this only works for dense arrays, doesn't work for sparse arrays since we'll get a matrix from the reinterpret, and the view in to that is a vector.

oscardssmith: makes sense.

ChrisRackauckas:
Small enough problems should probably just bypass.

PR #780, opened 2025-09-17
Add compat entry for MKL_jll

### LINKED ISSUE #779: `LinearSolve.getrf!` errors with MKL_jll < 2022.2.0
I ran into the following error with LinearSolve and MKL_jll@2021.1.1:

```julia
julia> using LinearSolve

julia> LinearSolve.getrf!([1.0;;])
ERROR: could not load symbol "dgetrf_64_": ...
```

This specific version of MKL_jll is part of a system image I have to use on JuliaHub.

In my opinion, the bug on the side of LinearSolve is that it defines MKL_jll-dependent functions but does not specify any compat entry for MKL_jll.
This would have made it obvious when building the system image that the MKL_jll version is too old.

Unfortunately, it seems no compat entry was added when the functionality was added in https://github.com/SciML/LinearSolve.jl/pull/349 (released in LinearSolve 2.5.0), and until today no compat entry is specified in the Project.toml file.

I tested other MKL versions as well, and it seems the first version that doesn't error is MKL_jll 2022.2.0.
For future versions, we can fix this issue by adding a lower bound for MKL_jll in the Project.toml file. For older versions >= 2.5.0 we would have to fix the registry.

devmotion:
Fixes #779 for new releases of LinearSolve.

Old versions (LinearSolve >= 2.5.0) would have to be fixed in the registry.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "LinearSolve"
 uuid = "7ed4a6bd-45f5-4d41-b270-4a48e9bafcae"
 authors = ["SciML"]
-version = "3.40.0"
+version = "3.40.1"
 
 [deps]
 ArrayInterface = "4fba245c-0d91-5ea0-9b3e-6abc04ee57a9"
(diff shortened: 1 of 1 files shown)


ChrisRackauckas (approved):

devmotion:
Just didn't know whether tests would pass since I didn't test it locally.

PR #785, opened 2025-09-18
Implement Butterfly Factorization method using RecursiveFactorization.jl

Shreyas-Ekanathan:
## Checklist

## Additional context

Implements a new `ButterflyFactorization()` algorithm

### DIFF
--- a/ext/LinearSolveRecursiveFactorizationExt.jl
+++ b/ext/LinearSolveRecursiveFactorizationExt.jl
@@ -130,7 +130,7 @@ function SciMLBase.solve!(cache::LinearSolve.LinearCache, alg::ButterflyFactoriz
 
     workspace, F = cache.cacheval
     (;A, b, ws, U, V, out, tmp, n) = workspace
-    b .= cache_b
+    b[1:M] .= cache_b
     mul!(tmp, U', b)
     TriangularSolve.ldiv!(F, tmp, thread)
     mul!(b, V, tmp)


ext/LinearSolveRecursiveFactorizationExt.jl
```diff
@@ -105,4 +104,105 @@ function SciMLBase.solve!(
...
+        U, V, F = RecursiveFactorization.🦋workspace(A, B, U, V)
+        cache.cacheval = (A, B, U, V, F)
+        cache.isfresh = false
+        if (M % 4 != 0)
+            b = [b; rand(4 - M % 4)]
+        end
+    end
+    A, B, U, V, F = cache.cacheval
+    sol = V * (F \ (U * b))    
+   SciMLBase.build_linear_solution(alg, sol[1:M], nothing, cache)
```
oscardssmith: make `out` as `similar(b, M)` as part of `🦋workspace` so you can do `out .= @view sol[1:M]` to avoid the allocation. Also fix indentation here.

ext/LinearSolveRecursiveFactorizationExt.jl
```diff
@@ -105,4 +104,105 @@ function SciMLBase.solve!(
...
+    A_32 = similar(A, T32)
+    b_32 = similar(b, T32)
+    u_32 = similar(u, T32)
+    luinst = ArrayInterface.lu_instance(rand(T32, 0, 0))
+    ipiv = Vector{LinearAlgebra.BlasInt}(undef, min(m, n))
+    # Return tuple with pre-allocated arrays
+    (luinst, ipiv, A_32, b_32, u_32)
+end
+
+function SciMLBase.solve!(
```
oscardssmith, line 61: can this be added in a separate PR?

Shreyas-Ekanathan: oops this is existing code that somehow got duplicated, must have been when I rebased.

Shreyas-Ekanathan:
@oscardssmith does this look good?

ext/LinearSolveRecursiveFactorizationExt.jl
```diff
@@ -105,4 +104,32 @@ function SciMLBase.solve!(
...
+function SciMLBase.solve!(cache::LinearSolve.LinearCache, alg::ButterflyFactorization;
+        kwargs...)
+    A = cache.A
+    A = convert(AbstractMatrix, A)
+    b = cache.b
+    M, N = size(A)
+    B, U, V = cache.cacheval[2], cache.cacheval[3], cache.cacheval[4]
+    if cache.isfresh
+        @assert M==N "A must be square"
+        U, V, F, out = RecursiveFactorization.🦋workspace(A, b, B, U, V, alg.thread)
```
ChrisRackauckas: Why is this not just a struct and `!` operation? It would be much easier to read. I assume this is all just in-place and non-allocating.

ext/LinearSolveRecursiveFactorizationExt.jl
```diff
@@ -105,4 +104,32 @@ function SciMLBase.solve!(
...
+    end
+    A, B, U, V, F = cache.cacheval
+    sol = V * (F \ (U * b))
+    out .= @view sol[1:M]    
+   SciMLBase.build_linear_solution(alg, out, nothing, cache)
+end
+
+function LinearSolve.init_cacheval(alg::ButterflyFactorization, A, b, u, Pl, Pr, maxiters::Int,
+        abstol, reltol, verbose::Bool, assumptions::LinearSolve.OperatorAssumptions)
+    A, A, A', A, RecursiveFactorization.lu!(rand(1, 1), alg.thread)
```
ChrisRackauckas:
```suggestion
    A, A, A', A, ArrayInterface.lu_instance(A)
```

ext/LinearSolveRecursiveFactorizationExt.jl
```diff
@@ -105,4 +104,32 @@ function SciMLBase.solve!(
...
+    b = cache.b
+    M, N = size(A)
+    B, U, V = cache.cacheval[2], cache.cacheval[3], cache.cacheval[4]
+    if cache.isfresh
+        @assert M==N "A must be square"
+        U, V, F, out = RecursiveFactorization.🦋workspace(A, b, B, U, V, alg.thread)
+        cache.cacheval = (A, B, U, V, F)
+        cache.isfresh = false
+        if (M % 4 != 0)
+            b = [b; rand(4 - M % 4)]
```
ChrisRackauckas: This allocation shouldn't be necessary in the loop? Why not just do it once? Also, if you loop this, the size will keep changing. That seems wrong?

ext/LinearSolveRecursiveFactorizationExt.jl
```diff
@@ -105,4 +104,32 @@ function SciMLBase.solve!(
...
+        end
+    end
+    A, B, U, V, F = cache.cacheval
+    sol = V * (F \ (U * b))
```
ChrisRackauckas: This is allocating and not using TriangularSolve.jl?

Shreyas-Ekanathan: I think we found that TriangularSolve.jl was slower than this method, so we left it as is

ext/LinearSolveRecursiveFactorizationExt.jl
Shreyas-Ekanathan: as in make `U, V, F, out` a struct?

ChrisRackauckas: yes

Project.toml
```diff
@@ -120,7 +120,7 @@ PrecompileTools = "1.2"
 Preferences = "1.4"
 Random = "1.10"
 RecursiveArrayTools = "3.37"
-RecursiveFactorization = "0.2.23"
+RecursiveFactorization = "0.2.25"
```
```suggestion
RecursiveFactorization = "0.2.26"
```

ChrisRackauckas:
Core LTS tests fail.

Shreyas-Ekanathan:
yeah you're right, sorry. hopefully that's fixed now

ext/LinearSolveRecursiveFactorizationExt.jl
```diff
@@ -105,4 +104,27 @@ function SciMLBase.solve!(
...
 
+function SciMLBase.solve!(cache::LinearSolve.LinearCache, alg::ButterflyFactorization;
+        kwargs...)
+    A = cache.A
+    A = convert(AbstractMatrix, A)
+    b = cache.b
+    M, N = size(A)
+    if cache.isfresh
+        @assert M==N "A must be square"
+        ws = RecursiveFactorization.🦋workspace(A, b)
```
oscardssmith: I think we can reuse the ws even when cache is fresh. As long as the size is the same, I think we can just update `A`, and `b` and everything works.

ext/LinearSolveRecursiveFactorizationExt.jl
```diff
@@ -105,4 +105,37 @@ function SciMLBase.solve!(
...
+            workspace = RecursiveFactorization.🦋workspace(cache_A, b)    
+            cache.cacheval = (workspace)
+        end
+        cache.isfresh = false
+    end
+    workspace = cache.cacheval
+    (;A, b, ws, U, V, out, tmp, n) = workspace
+    thread = alg.thread
+    RecursiveFactorization.🦋mul!(A, ws)
+    F = RecursiveFactorization.lu!(A, Val(false), thread)
```
oscardssmith: Aren't these lines only needed if `cache.isfresh`?

ext/LinearSolveRecursiveFactorizationExt.jl
```diff
@@ -107,4 +107,40 @@ function SciMLBase.solve!(
...
+        F = RecursiveFactorization.lu!(A, Val(false), thread)
+        cache.cacheval = (workspace, F)
+        cache.isfresh = false
+    end
+
+    workspace, F = cache.cacheval
+    (;A, b, ws, U, V, out, tmp, n) = workspace
+    mul!(tmp, U', b)
+    TriangularSolve.ldiv!(F, tmp, thread)
+    mul!(b, V, tmp)
```
ChrisRackauckas, line 136: Mutating `b` will make it incorrect for the second solve. It seems like you need another temporary vector to do this right?

ext/LinearSolveRecursiveFactorizationExt.jl
```diff
@@ -130,6 +130,9 @@ function SciMLBase.solve!(cache::LinearSolve.LinearCache, alg::ButterflyFactoriz
 
     workspace, F = cache.cacheval
     (;A, b, ws, U, V, out, tmp, n) = workspace
+    for i in 1:M
+        @inbounds b[i] = cache_b[i]
+    end
```
oscardssmith:
```suggestion
    b .= cache_b
```

ext/LinearSolveRecursiveFactorizationExt.jl
```diff
@@ -111,15 +111,15 @@ function SciMLBase.solve!(cache::LinearSolve.LinearCache, alg::ButterflyFactoriz
         kwargs...)
     cache_A = cache.A
     cache_A = convert(AbstractMatrix, cache_A)
-    b = cache.b
+    cache_b = cache.b
```
ChrisRackauckas, line 114: They alias the same vector though? I don't see how this would make a difference.

ext/LinearSolveRecursiveFactorizationExt.jl
```diff
@@ -130,6 +130,9 @@ function SciMLBase.solve!(cache::LinearSolve.LinearCache, alg::ButterflyFactoriz
 
     workspace, F = cache.cacheval
     (;A, b, ws, U, V, out, tmp, n) = workspace
```
ChrisRackauckas, line 132: or is workspace.b not an alias? It's a separate vector?

Shreyas-Ekanathan: It should be a separate vector since we pad `b`

Issue #790, opened 2025-10-01
Precompilation with 3.40.2 fails


svchb:
using CUDA
using LinearSolve

fails with the following error:

WARNING: Method definition usecuda() in module LinearSolve at LinearSolve.jl:440 overwritten in module LinearSolveCUDAExt at ext/LinearSolveCUDAExt.jl:14.\nERROR: Method overwriting is not permitted during Module precompilation.

PR #792, opened 2025-10-02
Fix precompilation of usecuda, usemetal, and useblis

ChrisRackauckas:
Fixes https://github.com/SciML/LinearSolve.jl/issues/790 and https://discourse.julialang.org/t/linearsolvecudaext-fails-to-compile/131693/10


topolarity:
Sorry about the regression folks. This looks good, thanks!

PR #796, opened 2025-10-05
Give a very clear and strong error message for 64-bit builds with 32-bit BLAS

ChrisRackauckas:
This should make https://github.com/SciML/LinearSolve.jl/issues/790 a lot less likely

### DIFF
--- a/src/LinearSolve.jl
+++ b/src/LinearSolve.jl
@@ -46,7 +46,7 @@ import Krylov
 
 const CRC = ChainRulesCore
 
-if Int === Int64 && Base.USE_BLAS64
+if Int === Int64 && !Base.USE_BLAS64
    error("Invalid installation of Julia detected.\n\n Detected that Julia was built in 64-bit version but with a 32-bit BLAS. This gives issues" *
          " in LinearAlgebra.jl and LinearSolve.jl which can be unrecoverable and are thus not supported. Most likely this is due to a bad build" *
          " of Julia, with the common reasons being an incorrect build script in the NixOS and ArchLinux package managers (and old versions of homebrew)." *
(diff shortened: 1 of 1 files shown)


src/LinearSolve.jl
```diff
@@ -46,6 +46,17 @@ import Krylov
 
 const CRC = ChainRulesCore
 
+if Int === Int64 && Base.USE_BLAS64
+   error("Invalid installation of Julia detected.\n\n Detected that Julia was built in 64-bit version but with a 32-bit BLAS. This gives issues" *
+         " in LinearAlgebra.jl and LinearSolve.jl which can be unrecoverable and are thus not supported. Most likely this is due to a bad build" *
+         " of Julia, with the common reasons being an incorrect build script in the NixOS and ArchLinux package managers (and old versions of homebrew)." *
+         " To fix this issue, and many other potentially small issues that may be undetected, use get a valid version of Julia with the correct BLAS and" *
+         " LLVM versions by either installing via juliaup (recommended), or downloading the appropriate binary from https://julialang.org/install/" *
+         " If using a Unix machine with a bash terminal, `curl -fsSL https://install.julialang.org` | sh will install juliaup and `juliaup add latest` will"
```
ChrisRackauckas:
```suggestion
         " If using a Unix machine with a bash terminal, `curl -fsSL https://install.julialang.org` | sh will install juliaup and `juliaup add latest` will" *
```

src/LinearSolve.jl
```diff
@@ -46,6 +46,17 @@ import Krylov
 
 const CRC = ChainRulesCore
 
+if Int === Int64 && Base.USE_BLAS64
```
ChrisRackauckas:
```suggestion
if Int === Int64 && !Base.USE_BLAS64
```

PR #798, opened 2025-10-07
support LinearSolve on no-gpl builds

oscardssmith:
when building julia without GPL libs, UMFPACK doesn't exist. For example julia +pr59699


ChrisRackauckas:
https://github.com/SciML/LinearSolve.jl/pull/797

oscardssmith:
That isn't sufficient.

PR #799, opened 2025-10-07
Fix nogpl precompile

oscardssmith:
KLU is LGPL so it should be allowed, and don't try to precompile UMFPack if nogpl


ext/LinearSolveSparseArraysExt.jl
```diff
@@ -11,7 +11,10 @@ using LinearAlgebra: LinearAlgebra, I, Hermitian, Symmetric, cholesky, ldiv!, lu
 using SparseArrays: SparseArrays, AbstractSparseArray, AbstractSparseMatrixCSC,
                     SparseMatrixCSC,
                     nonzeros, rowvals, getcolptr, sparse, sprand
+
+@static if Base.USE_GPL_LIBS
```
asinghvi17, line 15:
```suggestion
@static if isdefined(Base, :USE_GPL_LIBS) && Base.USE_GPL_LIBS
```

oscardssmith: no. This has always been defined.

asinghvi17: Ah I was not aware of that - makes sense

PR #382, opened 2023-09-25
Fix KrylovJL_GMRES with Enzyme

ChrisRackauckas:

PR #803, opened 2025-10-20
fix resid type

oscardssmith:
In LinearSolve@v3, `resid` for Krylov methods somehow ended up as a Ref{Float64} rather than a Float64. I assume this was unintentional. (found while debugging Thunderbolt)


oscardssmith:
@ChrisRackauckas actually it looks like this was intentional? https://github.com/SciML/LinearSolve.jl/pull/382 but something seems really off here.

PR #819, opened 2025-11-13
Add API to update iterative solver tolerances

termi-official:
## Checklist

### DIFF
--- a/test/basictests.jl
+++ b/test/basictests.jl
@@ -77,7 +77,7 @@ function test_interface(alg, prob1, prob2)
 end
 
 function test_tolerance_update(alg, prob, u)
-    cache = init(prob, alg; verbose=LinearVerbosity(; error_control=SciMLLogging.WarnLevel(), numerical=SciMLLogging.WarnLevel()))
+    cache = init(prob, alg)
     LinearSolve.update_tolerances!(cache; reltol = 1e-2, abstol=1e-8)
     u1 = copy(solve!(cache).u)
 


ChrisRackauckas:
@jClugstor

jClugstor:
It only hits that message if the retcode is also being set to Failure, which only happens if the solver stats `stats.solved` is false. So the message is doing what it's supposed to 

https://github.com/SciML/LinearSolve.jl/blob/5df1b2967226dd974901b8a68986b2de7d3922ad/src/iterative_wrappers.jl#L311

src/iterative_wrappers.jl
```diff
@@ -338,3 +338,5 @@ function SciMLBase.solve!(cache::LinearCache, alg::KrylovJL; kwargs...)
     return SciMLBase.build_linear_solution(alg, cache.u, Ref(resid), cache;
         iters = stats.niter, retcode, stats)
 end
+
+update_tolerances_internal!(cache, alg::KrylovJL, atol, rtol) = nothing
```
ChrisRackauckas, line 342: is this not supported? Since this is the only one people actually use.

termi-official: All Krylov packages are supported. This was was in particular easy, because you pass the tolerance into the solve call. The relevant lines for the transfer are these

https://github.com/SciML/LinearSolve.jl/blob/d53ba71ea3196f6f33977664eefecf1c520c29d3/src/iterative_wrappers.jl#L268-L269

https://github.com/SciML/LinearSolve.jl/blob/d53ba71ea3196f6f33977664eefecf1c520c29d3/src/iterative_wrappers.jl#L290-L291

https://github.com/SciML/LinearSolve.jl/blob/d53ba71ea3196f6f33977664eefecf1c520c29d3/src/iterative_wrappers.jl#L297

PR #820, opened 2025-11-14
Add trim tests for LUFactorization, MKLLUFactorization, and RFLUFactorization

ChrisRackauckas-Claude:
## Summary
Add trimming tests to LinearSolve.jl, similar to PR #734 in NonlinearSolve.jl.

This PR adds tests for Julia's trimming feature (available in Julia 1.12+) to verify that LinearSolve can be successfully trimmed when using pre-initialized caches with different LU factorization algorithms.

## Changes
- Add `test/trim/` directory with infrastructure for testing trimming
- Add trim tests for three factorization methods:
  - `LUFactorization` - Julia's built-in LU factorization
  - `MKLLUFactorization` - MKL-based LU factorization  
  - `RFLUFactorization` - RecursiveFactorization-based LU factorization
- Add "Trim" test group to CI configuration
- Tests only run on Julia 1.12+ (when trimming was introduced)

## Test Structure
The tests verify:
1. Each factorization works correctly with pre-initialized cache
2. Each implementation can be successfully compiled with `juliac --trim`

Each test creates a module with:
- `const` pre-initialized problem, algorithm, and cache
- A function that reuses the cache for solving with different parameters
- A main entry point for `juliac` compilation

## Notes
- Trim tests only run on Julia 1.12+
- CI excludes trim tests on LTS and pre-release versions
- Tests use a custom environment in `test/trim/` to avoid conflicts
- Each factorization is tested separately to ensure independent trimming

## Related
- Similar to NonlinearSolve.jl PR #734
- Addresses trimming support for LinearSolve algorithms

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/test/trim/runtests.jl
+++ b/test/trim/runtests.jl
@@ -11,7 +11,7 @@ end
     using LinearAlgebra
     using LinearSolve
     # Only test if MKL is available
-    if LinearSolve.usemkl(nothing)
+    if LinearSolve.usemkl
         include("linear_mkl.jl")
         sol = TestMKLLUFactorization.solve_linear(1.0)
         @test sol.u ≈ [0.09090909090909091, 0.6363636363636364]
@@ -59,7 +59,7 @@ end
         ("main_lu.jl", true),
         ("main_rf.jl", true)
     ]
-    if LinearSolve.usemkl(nothing)
+    if LinearSolve.usemkl
         push!(test_files, ("main_mkl.jl", true))
     end
 


ChrisRackauckas-Claude:
Fixed: Removed JuliaFormatter from dependencies (it was incorrectly added during formatting and should not be a package dependency)

ChrisRackauckas-Claude:
Fixed: Changed to return the solution object from `solve!(cache)` instead of returning the cache itself

ChrisRackauckas-Claude:
Fixed: Changed `LinearSolve.usemkl(nothing)` to `LinearSolve.usemkl` - `usemkl` is a constant boolean, not a function. This fixes the "MethodError: objects of type Bool are not callable" error.

PR #822, opened 2025-11-14
Trim allocations in dual linear problem

hersle:
I am differentiating ODEs with ForwardDiff in my package.
```julia
using SymBoltz, ProfileCanvas
...
@time ForwardDiff.jacobian(logP, logθ) # ForwardDiff-differentiates ~20 stiff ODEs with ~50x50 Jacobians
@profview_allocs ForwardDiff.jacobian(logP, logθ)
```
```
  1.537143 seconds (1.17 M allocations: 906.513 MiB, 18.37% gc time)
```
image

There are a lot of allocations from `nodual_value(val)` and `partial_vals(val)` in LinearSolve's ForwardDiff extension. They set some cache field with a new reallocated array every time. This seems unnecessary to me.

In this PR I attempt to instead use `map!` and a custom `setproperty!` dispatch to set the elements `nodual_value(val[i])` and `partial_vals(val[i])` without having to reallocate the arrays. This gives me a much shorter runtime and lower memory usage:
```
  0.823024 seconds (218.36 k allocations: 18.415 MiB)
```
image
Now the LinearSolve allocations in the first flamegraph are gone.

## Checklist

### DIFF
--- a/ext/LinearSolveForwardDiffExt.jl
+++ b/ext/LinearSolveForwardDiffExt.jl
@@ -308,31 +308,54 @@ function SciMLBase.solve!(
     )
 end
 
-# If setting A or b for DualLinearCache, put the Dual-stripped versions in the LinearCache
-function Base.setproperty!(dc::DualLinearCache, sym::Symbol, val::AbstractArray)
+function setA!(dc::DualLinearCache, A)
+    # Put the Dual-stripped versions in the LinearCache
+    prop = nodual_value!(getproperty(dc.linear_cache, :A), A) # Update in-place
+    setproperty!(dc.linear_cache, :A, prop) # Does additional invalidation logic etc.
+
+    # Update partials
+    setfield!(dc, :dual_A, A)
+    partial_vals!(getfield(dc, :partials_A), A) # Update in-place
+
+    # Invalidate cache (if setting A or b)
+    setfield!(dc, :rhs_cache_valid, false)
+end
+function setb!(dc::DualLinearCache, b)
+    # Put the Dual-stripped versions in the LinearCache
+    prop = nodual_value!(getproperty(dc.linear_cache, :b), b) # Update in-place
+    setproperty!(dc.linear_cache, :b, prop) # Does additional invalidation logic etc.
+
+    # Update partials
+    setfield!(dc, :dual_b, b)
+    partial_vals!(getfield(dc, :partials_b), b) # Update in-place
+
+    # Invalidate cache (if setting A or b)
+    setfield!(dc, :rhs_cache_valid, false)
+end
+function setu!(dc::DualLinearCache, u)
+    # Put the Dual-stripped versions in the LinearCache
+    prop = nodual_value!(getproperty(dc.linear_cache, :u), u) # Update in-place
+    setproperty!(dc.linear_cache, :u, prop) # Does additional invalidation logic etc.
+
+    # Update partials
+    setfield!(dc, :dual_u, u)
+    partial_vals!(getfield(dc, :partials_u), u) # Update in-place
+end
+
+function Base.setproperty!(dc::DualLinearCache, sym::Symbol, val)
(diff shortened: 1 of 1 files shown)


ext/LinearSolveForwardDiffExt.jl
```diff
@@ -383,7 +383,7 @@ function update_partials_list!(partial_matrix::AbstractVector{T}, list_cache) wh
     p = eachindex(first(partial_matrix))
     for i in p
         for j in eachindex(partial_matrix)
-            list_cache[i][j] = partial_matrix[j][i]
+            @inbounds list_cache[i][j] = partial_matrix[j][i]
```
ChrisRackauckas, line 400: are these needed?

src/common.jl
```diff
@@ -141,6 +141,14 @@ function Base.setproperty!(cache::LinearCache, name::Symbol, x)
     setfield!(cache, name, x)
 end
 
+# Dispatch for setting a property to f.(x) without allocating
+function Base.setproperty!(cache::LinearCache, name::Symbol, x::AbstractArray, f::Function)
```
ChrisRackauckas: Is this used? Also, it's a pun so probably should be changed.

hersle: Yes, it's used here:

https://github.com/SciML/LinearSolve.jl/blob/3c4784494b30a0e9a9dde07cbc93a8502a4bb202/ext/LinearSolveForwardDiffExt.jl#L315

Hmm, what do you mean by it's a pun? 😅

hersle: It's a "trick" to call `setproperty!` while doing an in-place of the array.

ext/LinearSolveForwardDiffExt.jl
hersle: Not required, but they gave me a non-negligible speedup, as I saw `checkbounds` showing up in profiling.

hersle: I added it so the user should not suffer a performance penalty from library internals. I can remove them if you want to keep it safe.

This function is just shuffling data around. The optimal solution would be to avoid it altogether, but I am not sure if it's easily possible.

test/forwarddiff_overloads.jl
```diff
@@ -165,11 +165,10 @@ cache.b = new_b
 linu = [ForwardDiff.Dual(0.0, 0.0, 0.0), ForwardDiff.Dual(0.0, 0.0, 0.0),
     ForwardDiff.Dual(0.0, 0.0, 0.0)]
 cache.u = linu
+@test linu == cache.u
```
hersle: I have to move this before the `solve!` to make this test pass. Is that fine or am I overlooking something?

src/common.jl
ChrisRackauckas: But why use the name of `setproperty`? It's not actually a `setproperty!` call

test/forwarddiff_overloads.jl
ChrisRackauckas: No that's not good.

src/common.jl
hersle: Yeah that's true. After understanding everything better, I came up with a simpler version that just uses standard `setproperty!` calls and new in-place `nodual_value!` and `partial_vals!` functions.

test/forwarddiff_overloads.jl
hersle: Sorry this was never a problem, my local state messed up the testing. Removing this.

hersle:
@ChrisRackauckas Refactored to use just standard `setproperty!` calls and new in-place `nodual_value!` and `partial_vals!` functions. I think this is cleaner. Local test passes:
```julia-repl
julia> include("test/forwarddiff_overloads.jl")
Test Passed
```

hersle:
Hold on, there is a small type instability in

image

Seems related to the (somewhat complex) branching/recursion on `sym` in `setproperty!`.

PR #842, opened 2025-12-05
Add overloads for overdetermined system derivatives 

jClugstor:
## Checklist

### DIFF
--- a/ext/LinearSolveForwardDiffExt.jl
+++ b/ext/LinearSolveForwardDiffExt.jl
@@ -63,11 +63,10 @@ LinearSolve.@concrete mutable struct DualLinearCache{DT}
 end
 
 function linearsolve_forwarddiff_solve!(cache::DualLinearCache, alg, args...; kwargs...)
-    # Check if we're solving an overdetermined system (more rows than columns in A)
+    # Check if A is square - if not, use the non-square system path
     A = cache.linear_cache.A
-    is_overdetermined = size(A, 1) > size(A, 2)
 
-    if is_overdetermined
+    if !issquare(A)
         # For overdetermined systems, differentiate the normal equations: A'Ax = A'b
         # Taking d/dθ of both sides:
         # dA'/dθ · Ax + A' · dA/dθ · x + A'A · dx/dθ = dA'/dθ · b + A' · db/dθ
--- a/test/forwarddiff_overloads.jl
+++ b/test/forwarddiff_overloads.jl
@@ -284,6 +284,9 @@ backslash_large = A_large_dual \ b_large_dual
 # Test primal values match
 @test ForwardDiff.value.(sol_large.u) ≈ ForwardDiff.value.(backslash_large)
 
+@test A_large_dual' * A_large_dual * sol_large.u ≈ A_large_dual' * b_large_dual
+@test A_large_dual' * A_large_dual * backslash_large ≈ A_large_dual' * b_large_dual
+
 # Test partials match
 @test ForwardDiff.partials.(sol_large.u) ≈ ForwardDiff.partials.(backslash_large)
 


ChrisRackauckas:
tests fail

jClugstor:
@ChrisRackauckas only failure I see that could come from this is the adjoint test for lts, but as far as I can tell those are all square systems, so this PR shouldn't affect those.

ext/LinearSolveForwardDiffExt.jl
```diff
@@ -63,6 +63,84 @@ LinearSolve.@concrete mutable struct DualLinearCache{DT}
...
+                rhs_list[i] .+= A_list[i]' * residual
+                # Subtract A' · dA/dθ · x
+                temp = A_list[i] * u
+                rhs_list[i] .-= A' * temp
+            end
+        end
+
+        # Solve A'A · dx/dθ = rhs for each partial
+        # Create a cache for the normal equations and reuse the factorization
+        AtA = A' * A
```
ChrisRackauckas: This is numerically unstable. Instead just solve `A * dx/dθ = rhs`, the methods like QR factorization and GMRES will naturally do this without squaring the eigenvalues

jClugstor: The number of rows don't match in that case though? 

```julia 
infil> rhs_list[1]
3-element Vector{Float64}:
 -0.49033292918119054
 -0.4863208414656224
 -0.49187865599734093

infil> A
10×3 Matrix{Float64}:
 0.8753     0.651302   0.9616
 0.636808   0.304259   0.0872156
 0.542556   0.359615   0.943729
 0.0258567  0.0241303  0.070481
 0.275383   0.282139   0.866992
 0.0623616  0.338967   0.687406
 0.481159   0.0197438  0.996632
 0.905349   0.555788   0.215757
 0.0858424  0.944214   0.506554
 0.452954   0.033606   0.38146
```

ChrisRackauckas: no it's just `A' \ rhs_list[i]`, which then solves via QR/GMRES

jClugstor: Oh I think I get it. Is it better now?

ext/LinearSolveForwardDiffExt.jl
```diff
@@ -63,6 +63,79 @@ LinearSolve.@concrete mutable struct DualLinearCache{DT}
 end
 
 function linearsolve_forwarddiff_solve!(cache::DualLinearCache, alg, args...; kwargs...)
+    # Check if we're solving an overdetermined system (more rows than columns in A)
+    A = cache.linear_cache.A
+    is_overdetermined = size(A, 1) > size(A, 2)
+
+    if is_overdetermined
```
ChrisRackauckas: what is the difference between this and the other branch? Can't they just be the same?

jClugstor: I wasn't getting correct derivatives from the other branch. 

The original branch is based off of `A(p)x(p) = b(p)`. You can take the derivative of both sides, use the product rule, and rearrange to get a linear solve for dx/dp. 

But when we do a `\` or a solve a LinearProblem that's overdetermined/underdetermined, we're not really solving `Ax=b`, because you get an `x` that doesn't satisfy the equation. 

In those cases we're actually getting an answer for `A'(p)A(p)x(p) = A'(p)b(p)`. So we need to do the derivative calculation based on that, so you get a different linear system to solve to find dx/dp.

ChrisRackauckas: This branch should give the same answer for both.

jClugstor: I've changed it to just check `!issquare(A)` now

ChrisRackauckas: Why not just use this branch?

jClugstor: Oh, because the not square branch has to do two linear solves for each partial column, but that's unnecessary for the square case.

PR #846, opened 2025-12-10
Remove UnPack

devmotion:
Simple regex search and replace.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "LinearSolve"
 uuid = "7ed4a6bd-45f5-4d41-b270-4a48e9bafcae"
 authors = ["SciML"]
-version = "3.48.1"
+version = "3.49.0"
 
 [deps]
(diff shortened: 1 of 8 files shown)


Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "LinearSolve"
 uuid = "7ed4a6bd-45f5-4d41-b270-4a48e9bafcae"
 authors = ["SciML"]
-version = "3.48.1"
+version = "3.49.0"
```
oscardssmith, line 4: Arguably a patch version, but I don't really care.

devmotion: It's not a bug fix though.

oscardssmith: It's not a bugfix, but I think it is a patch. It introduces no new features to the public api which IMO is the important difference between a minor and patch release.

devmotion: ColPrac disagrees: https://docs.sciml.ai/ColPrac/stable/#Changing-dependency-compatibility

oscardssmith: good to know!

PR #865, opened 2025-12-30
Fix SVDFactorization CUDA failure by deferring algorithm selection

### LINKED ISSUE #488: Don't select an algorithm for SVDFactorization
Selecting an algorithm for SVDFactorization will cause it to fail for CUDA. Possible solutions:

1. Keep it as `nothing` and use the default versions used in the downstream package

ArnavTomar18:
This PR removes the hard-coded LinearAlgebra.DivideAndConquer()
algorithm from SVDFactorization so that downstream svd implementations
(e.g. CuArray / cuSOLVER) can select a compatible default algorithm.

This fixes failures when running SVD on CUDA arrays.

Closes #488

### DIFF
--- a/src/factorization.jl
+++ b/src/factorization.jl
@@ -455,13 +455,12 @@ end
 ## SVDFactorization
 
 """
-`SVDFactorization(full=false,alg=LinearAlgebra.DivideAndConquer())`
+    SVDFactorization(full=false, alg=nothing)
+Julia's built-in `svd`. Equivalent to `svd!(A)`.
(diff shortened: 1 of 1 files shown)


arnav-tomar:
@ChrisRackauckas thanks! I’ve pushed a fix that removes the hard-coded DivideAndConquer() default in SVDFactorization() and lets the backend svd choose the algorithm, which should avoid the CUDA failure.
Let me know if you want any changes or tests added.

PR #868, opened 2026-01-07
refactor: create new `LinearProblem` instead of `remake` in ForwardDiff ext

AayushSabharwal:


ChrisRackauckas:
What does this actually fix?

PR #871, opened 2026-01-16
Add ReactantCore.jl support with @trace macros for Reactant AD

ChrisRackauckas-Claude:
## Summary

- Add ReactantCore.jl as a dependency to enable Reactant compatibility
- Import and use `@trace` macro from ReactantCore for data-dependent control flow in SimpleGMRES solver
- Add comprehensive Reactant AD test suite mirroring the existing Enzyme tests

### Dependencies
- Added `ReactantCore = "a3311ec8-5e00-46d5-b541-4f83e724a433"` to `[deps]` in Project.toml
- Added `ReactantCore = "0.1"` to `[compat]` in Project.toml
- Added `Reactant` to test/nopre/Project.toml for testing

### Code Changes
- `src/LinearSolve.jl`: Import `@trace` macro from ReactantCore
- `src/simplegmres.jl`: Added `@trace` to data-dependent while loops in both `SimpleGMRESCache{false}` and `SimpleGMRESCache{true}` solve functions for Reactant compatibility

### Tests
- Created `test/nopre/reactant.jl` with comprehensive tests for:
  - LUFactorization with Reactant AD
  - DefaultLinearSolver with Reactant AD  
  - RFLUFactorization with Reactant AD
  - Cache reuse with Reactant AD
  - Forward mode differentiation
  - OperatorAssumptions with Reactant AD

## Notes

The `@trace` macro from ReactantCore enables Reactant.jl to properly trace data-dependent control flow (while loops whose conditions depend on computed values). This is necessary for Reactant's compilation to work correctly with iterative solvers like GMRES.

Note: `@trace` cannot wrap blocks containing `return` statements, so early return checks remain as regular `if` statements.

## Test plan

- [x] Package compiles successfully
- [ ] Run existing test suite to verify no regressions
- [ ] Run new Reactant AD tests (requires Reactant.jl which needs Julia 1.11 or earlier)


ChrisRackauckas:
This doesn't handle the fact that we effectively don't want to trace through here. 

@avik-pal

Issue #8, opened 2021-07-16
PETSc wrappers


ChrisRackauckas:
https://github.com/JuliaParallel/PETSc.jl

PR #894, opened 2026-02-20
feat: add/fix PETSc KSP solver wrapper

Iddingsite:
Add ksp() support from PETSc.jl with the new version of the wrapper (0.4.x, https://github.com/JuliaParallel/PETSc.jl/pull/215). Closes https://github.com/SciML/LinearSolve.jl/issues/8. I believe Claude was mostly hallucinating, because this PR https://github.com/SciML/LinearSolve.jl/pull/873 was not working. On Master, this is failing:

```
using PETSc, MPI, SparseArrays, LinearSolve, LinearAlgebra

A = sprand(100, 100, 0.1) + 10I
b = rand(100)
prob = LinearProblem(A, b)

# Basic GMRES
sol = solve(prob, PETScAlgorithm(:gmres))

# CG with ILU preconditioner
sol = solve(prob, PETScAlgorithm(:cg; pc_type = :ilu))
```
but works with this PR. 

## What is currently supported

- [x] All PETSc KSP types (:gmres, :cg, :bcgs, :richardson, :preonly, …)
- [x] All PETSc preconditioners (:jacobi, :ilu, :lu, :gamg, :hypre, …)
- [ ] Serial and MPI-distributed solves via the comm keyword
- [x] Null-space specification (:constant or :custom) for singular systems
- [x] Transposed solves via `transposed = true`
- [x] Arbitrary PETSc Options Database flags via `ksp_options::NamedTuple` (for instance ksp_options = (ksp_monitor = "", ksp_view = "")).

What I am not sure/not happy with: 
- With PETSc.jl, there is a need to explicitly free the cache after use for the memory. Right now, I recommend to use `PETScExt.cleanup_petsc_cache!(cache)` to clear it after usage (works on `solve()` or `integ()` object). Most people will clear the memory by closing the Julia session, but I think it makes sense to expose such a function.
- ~~Currently, the MPI implementation is a bit clunky/naive because every rank holds the full Julia matrix in memory and inserts its locally owned rows into PETSc. I could not come up with a better way because of the way LinearSolve.jl expect the input from the users. I am happy to remove the MPI support if that is not satisfying enough. Also needs some proper benchmarks.~~ I have removed the MPI support for more processes than 1 because the PR was getting too big.
- ~~The use of PETSc in general is mostly based on Sparse matrix for large system. Because of that, the handling of for dense matrices is a bit clunky and I could not make the reuse of ksp() work with how it is communicating with PETSc, and it currently has to rebuild it everytime `solve!()` is called again. However, this works correctly for sparse matrix. Could potentially be improved in another PR.~~ This works well for dense matrix now
- ~~Not sure where to put the test files. There is `petsctests_mpi.jl` and `petsctests.jl` right now.~~ Removed `petsctests_mpi.jl` for now

## Checklist

AI Disclosure: Claude was used to help writting the code and the comments + docstrings

### DIFF
--- a/test/petsctests.jl
+++ b/test/petsctests.jl
@@ -4,7 +4,6 @@ using MPI
 using PETSc
 using SparseArrays
 using Random
-using Statistics
 using Test
 
 MPI.Init()


Iddingsite:
@ChrisRackauckas I think this is ready for review but nothing urgent. Everything works for dense and sparse matrices with all the solvers and preconditionners, and with support for null-space. I dropped the MPI support for now because the PR was getting too big to review, so it only works with 1 process for now. I am not sure anyway LinearSolve.jl is fit for that. 

The big limitation currently is that PETSc expect sparse matrices with CSR format, whereas Julia is using CSC. So, when reusing the cache for large scale problems, the time is spent on converting from CSR to CSC format. That is also what is done currently in the PETSc.jl package when initialising the first time using (see [here](https://github.com/JuliaParallel/PETSc.jl/blob/943bd1441e483e5519bd43dd926e59324cc98134/src/mat.jl#L49C5-L68C4)). I could not come up with something cleaner, but happy to take suggestions.

That currently makes it hard to use it as a stand alone for a non-linear solver where we want to reuse the same cache, and people should just use SNES for that I guess. 

The PR needs https://github.com/JuliaParallel/PETSc.jl/pull/227 so we need at least 0.4.6 for PETSc.jl.

PR #898, opened 2026-02-22
Add Ginkgo.jl wrapper (GinkgoJL, GinkgoJL_CG)

### LINKED ISSUE #486: Ginkgo wrapper
https://github.com/youwuyou/Ginkgo.jl

AJ0070:
## Checklist

## Additional context

Closes #486 

Adds a LinearSolve.jl extension wrapping [Ginkgo.jl](https://github.com/youwuyou/Ginkgo.jl) (Julia bindings for the [Ginkgo](https://ginkgo-project.github.io/) HPC sparse linear algebra library).

New public API: `GinkgoJL` and `GinkgoJL_CG`, supporting `:omp`, `:cuda`, and `:reference` executors. Currently wraps CG only, as that is the only solver exposed by Ginkgo.jl. Input sparse matrices are automatically converted to `SparseMatrixCSC{Float32, Int32}` and vectors to `Float32` as required by Ginkgo.jl.

Ginkgo is registered as a `[weakdeps]`-only entry (not in `[extras]`/`[targets]`) since `ginkgo_jll` lacks artifacts for Windows and 32-bit platforms. Tests in `basictests.jl` follow the existing `AlgebraicMultigrid` pattern and are skipped gracefully when Ginkgo is unavailable.

### DIFF
--- a/ext/LinearSolveGinkgoExt.jl
+++ b/ext/LinearSolveGinkgoExt.jl
@@ -103,7 +103,7 @@ function LinearSolve.init_cacheval(
         verbose::Union{LinearVerbosity, Bool}, assumptions::OperatorAssumptions
     )
     exec = Ginkgo.create(alg.executor)
-    return (; exec = exec.ptr)
+    return (; exec)
 end
 
 function SciMLBase.solve!(cache::LinearCache, alg::GinkgoJL; kwargs...)


ext/LinearSolveGinkgoExt.jl
```diff
@@ -0,0 +1,164 @@
...
+
+# -----------------------------------------------------------------------
+# Copy a Dense{Float32} result back into a Julia vector
+# -----------------------------------------------------------------------
+
+function _copy_gko_dense_to!(dest::AbstractVector, src::Ginkgo.Dense{Float32})
+    n = length(dest)
+    for i in 1:n
+        dest[i] = src[i, 1]
+    end
```
ChrisRackauckas: ?

ext/LinearSolveGinkgoExt.jl
```diff
@@ -0,0 +1,164 @@
...
+    # ---- Convert A -------------------------------------------------
+    gko_A = _to_gko_csr(cache.A, exec)
+
+    # ---- Convert b and initial guess x ----------------------------
+    gko_b = _to_gko_dense(cache.b, exec)
+    gko_x = _to_gko_dense(cache.u, exec)
+
+    # ---- Dispatch to solver ----------------------------------------
+    if alg.KrylovAlg === :cg
+        Ginkgo.cg!(
```
ChrisRackauckas: gmres?

src/extension_algs.jl
```diff
@@ -722,6 +722,65 @@ A wrapper over the IterativeSolvers.jl MINRES.
...
+including OpenMP, CUDA, HIP, and SYCL, making it suitable for both CPU and GPU computation.
+
+!!! note
+
+    Using this solver requires adding the package Ginkgo.jl, i.e. `using Ginkgo`
+
+## Keyword Arguments
+
+  - `KrylovAlg`: The Ginkgo solver to use. Currently supported:
+    - `:cg` (default): Conjugate Gradient — for symmetric positive definite systems
```
ChrisRackauckas: why choose cg as the default? Not safe.

ext/LinearSolveGinkgoExt.jl
```diff
@@ -0,0 +1,164 @@
...
+# Convert a Julia AbstractVector to Dense{Float32}
+# -----------------------------------------------------------------------
+
+function _to_gko_dense(v::AbstractVector, exec)
+    tmpfile = tempname() * ".mtx"
+    try
+        _write_dense_vec_mtx(v, tmpfile)
+        return Ginkgo.Dense{Float32}(tmpfile, exec)
+    finally
+        rm(tmpfile, force = true)
```
ChrisRackauckas: Seems really slow to write it all to a file. There's no way to keep it in memory?

src/extension_algs.jl
AJ0070: changing it to gmres

ext/LinearSolveGinkgoExt.jl
AJ0070: It has no in-memory constructor

AJ0070: what I can do is I can build full context of MTX writers in an IObuffer first and flush with single write call instead of println per element. would it be okay?

ChrisRackauckas: none at all? Even in the jll directly?

ChrisRackauckas: It still seems like a very slow thing to do. Is the performance here not awful?

AJ0070: After looking into the Ginkgo.jl C API, disk I/O can be avoided for sparse matrices but not for dense vectors. For CSR matrices, `ginkgo_matrix_csr_f32_i32_create` allocates internal storage and the `get_const_{values,col_idxs,row_ptrs}` accessors provide writable pointers, allowing data from a Julia CSC→CSR conversion to be copied directly into Ginkgo’s buffers using `unsafe_copyto!`, with no file involved; this matches how `Ginkgo.nonzeros()`, `rowptr()`, and `colvals()` operate internally via `unsafe_wrap`. In contrast, dense matrices lack a writable pointer interface: although `ginkgo_matrix_dense_f32_create` creates uninitialized storage, there is no accessor to populate it from a Julia array, `ginkgo_matrix_dense_f32_at` only returns values, `fill` sets a constant, and `setindex!` is unimplemented in `Dense.jl`. As a result, temporary files are required for `b` and `x`, with disk overhead minimized by formatting MTX data in an `IOBuffer` so each vector triggers only a single `write()` call.

ChrisRackauckas: we really just want the sparse matrices here

ext/LinearSolveGinkgoExt.jl
```diff
@@ -0,0 +1,191 @@
...
+# The get_const_* accessors return raw writable pointers; unsafe_copyto! fills
+# Ginkgo's internal buffers directly — same pattern as Ginkgo.nonzeros() etc.
+
+"""
+    _csc_to_csr_arrays(A)
+
+Convert a `SparseMatrixCSC{Float32,Int32}` to the three CSR arrays required
+by Ginkgo (all 0-indexed): `rowptr` (length m+1), `col_idxs` (length nnz),
+`vals` (length nnz).
+"""
```
ChrisRackauckas: Is this not in SparseArrays?

PR #900, opened 2026-02-23
Add LinearSolvePyAMG sub-library with PyAMG integration and tests

AJ0070:
## Checklist

- [x] Appropriate tests were added
- [x] Any code changes were done in a way that does not break public API
- [x] All documentation related to code changes were updated
- [x] The new code follows the
  [contributor guidelines](https://github.com/SciML/.github/blob/master/CONTRIBUTING.md), in particular the [SciML Style Guide](https://github.com/SciML/SciMLStyle) and
  [COLPRAC](https://github.com/SciML/COLPRAC).
- [x] Any new documentation only uses public API
  
### Changes
- lib/LinearSolvePyAMG/ : new sub-library (Project.toml, CondaPkg.toml, src/, test/)
- docs/src/basics/Preconditioners.md : replaced stale cortner/PyAMG.jl entry
- docs/src/solvers/solvers.md : added `### LinearSolvePyAMG.jl` section
- test/basictests.jl : added opt-in `@static if isdefined(...)` test block
- .gitignore : added `.CondaPkg/`

Implemented as a sub-library (like `lib/LinearSolveAutotune/`) rather than a weak dependency because PyAMG is not in the Julia General registry.

Resolves #372 

AI disclosure: used claude and gemini to help in replicating structure of LinearSolveAutotune in LinearSolvePyAMG, adding comments and refactoring documentation.

PR #907, opened 2026-03-01
Remove ignored link for linkcheck in make.jl

AJ0070:
Removed a link from the linkcheck_ignore list in the documentation configuration.

cleanup for #900


AJ0070:
@ChrisRackauckas

AJ0070:
thank you for the quick merge!

PR #925, opened 2026-03-24
Add KLU init_cacheval for ComplexF64 sparse matrices

ChrisRackauckas-Claude:
## Summary

Fixes SciML/OrdinaryDiffEq.jl#2892

The vendored KLU module wraps both real (`klu_d_*`) and complex (`klu_z_*`) SuiteSparse functions, but `LinearSolveSparseArraysExt` only had `init_cacheval` methods for `Float64` sparse matrices. Complex sparse matrices fell through to the generic `AbstractArray` fallback which returns `nothing`, causing `solve!` to crash with `type Nothing has no field nzval`.

- Added `init_cacheval` for `AbstractSparseArray{ComplexF64, Int64}` (with preallocated constant)
- Added `init_cacheval` for `AbstractSparseArray{ComplexF64, Int32}`
- Added tests for complex sparse KLU (direct solve, cached solve, refactorization)

This fixes FIRK solvers (RadauIIA3/5/9, AdaptiveRadau) which need complex linear systems for their Butcher tableau eigenvalue pairs when used with `KLUFactorization()` and sparse Jacobians.

## Test plan

- [x] Complex KLU solve, cache, and refactorization tests pass
- [x] Verified OrdinaryDiffEq.jl issue #2892 MRE works with this fix (all 4 Radau variants)
- [x] Existing sparse_vector.jl tests still pass

### DIFF
--- a/ext/LinearSolveSparseArraysExt.jl
+++ b/ext/LinearSolveSparseArraysExt.jl
@@ -324,20 +324,18 @@ function LinearSolve.init_cacheval(
     return PREALLOCATED_KLU
 end
 
-const PREALLOCATED_KLU_COMPLEX = KLU.KLUFactorization(
-    SparseMatrixCSC(
-        0, 0, [1], Int[],
-        ComplexF64[]
-    )
-)
-
+# KLU supports Float64 and ComplexF64 (KLUTypes)
 function LinearSolve.init_cacheval(
-        alg::KLUFactorization, A::AbstractSparseArray{ComplexF64, Int64}, b, u, Pl, Pr,
+        alg::KLUFactorization, A::AbstractSparseArray{T, Int64}, b, u, Pl, Pr,
         maxiters::Int, abstol,
         reltol,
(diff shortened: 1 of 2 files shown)


ext/LinearSolveSparseArraysExt.jl
```diff
@@ -324,6 +324,22 @@ function LinearSolve.init_cacheval(
...
 
+const PREALLOCATED_KLU_COMPLEX = KLU.KLUFactorization(
+    SparseMatrixCSC(
+        0, 0, [1], Int[],
+        ComplexF64[]
+    )
+)
+
+function LinearSolve.init_cacheval(
+        alg::KLUFactorization, A::AbstractSparseArray{ComplexF64, Int64}, b, u, Pl, Pr,
```
oscardssmith: Should we also support ComplexF32?

oscardssmith: also why are we preallocating a global for this?

ChrisRackauckas: Nope SuiteSparse doesn't support it.

Issue #914, opened 2026-03-07
Dual number overloads fail with  dual numbers with no partials.


j-fu:
See the discussion in https://discourse.julialang.org/t/automatic-differentiation-of-function-using-linearproblem/136063

ChrisRackauckas:
Dual numbers without partials are ill-formed. Is there a reason to actually support this?

j-fu:
True, it probably doesn't make sense. But may be a clearer error message could be helpful.

AJ0070:
This no longer reproduces on current main (Julia 1.12.6). I checked zero-partial `Dual{T,V,0}` values across ten surfaces: Dual A with plain b, Dual A with Dual b, plain A with Dual b, explicit LUFactorization, GenericLUFactorization, QRFactorization and KrylovJL_GMRES, `init` plus `solve!`, cache reuse with a new b, matrix right hand sides, nested zero-partial Duals, and a mixed case of zero-partial A against a 2-partial b. All of them solve and return the expected values, so the clearer error message discussed here is no longer needed. Ready to close.

AJ0070:
Correcting my own comment above: it tested a reconstruction with dense matrices, but the reproducer in the linked Discourse thread is sparse and goes through KLU, which is the path that actually threw `BoundsError: attempt to access Tuple{} at index [1]`. I have now run that exact code on current main (Julia 1.12.6):

```julia
using SparseArrays, LinearSolve, ForwardDiff
A = ForwardDiff.Dual.(sprand(4, 4, 0.75))   # eltype Dual{Nothing, Float64, 0}
b = ForwardDiff.Dual.(rand(4))
sol = solve(LinearProblem(A, b), KLUFactorization())
```

It returns a solution with no error. I also checked a nonsingular sparse variant of the same problem through KLUFactorization, UMFPACKFactorization, LUFactorization and the default algorithm, all of which solve and agree with each other. So the zero-partial case is handled on the path this issue was actually about.

PR #927, opened 2026-03-26
Improve ForwardDiff Dual N=0 handling with clear ArgumentError and regression test

AJ0070:
## Checklist
- [x] The new code follows the contributor guidelines, in particular the SciML Style Guide and COLPRAC  

## What changed
- Added an early check in the ForwardDiff extension init path to detect zero-partial Dual inputs from A or b.
- Added a targeted `ArgumentError` message to make the failure mode explicit and actionable.
- Added a regression test in the ForwardDiff overload test file to lock in this behavior.

## Additional context
This follows discussion on #914: zero-partial Duals are ill-formed for the intended AD path, but users should still get a clear error instead of an internal failure.

AI Disclosure: Used codex


ext/LinearSolveForwardDiffExt.jl
```diff
@@ -303,6 +303,22 @@ function SciMLBase.init(
...
+_npartials(partials::Partials) = length(partials)
+_npartials(partials::AbstractArray{<:Partials}) = isempty(partials) ? 0 : length(first(partials))
+
+function _check_supported_forwarddiff_partials!(∂_A, ∂_b)
+    nA = _npartials(∂_A)
+    nb = _npartials(∂_b)
+    if nA === 0 || nb === 0
+        throw(ArgumentError(
+            "LinearSolve does not support ForwardDiff.Dual values with zero partials (N = 0). " *
+            "Use primal values with ForwardDiff.value(...) or construct Dual numbers with at least one partial."
```
devmotion: `ForwardDiff.value` is internal, but more importantly it's also generally unsafe and hence usually one should use `ForwardDiff.value(T, x)` instead (which is internal as well though, so I don't think it belongs into a user-facing error message).

AJ0070: Then I’ll just replace it with a note advising not to use non-Dual numbers.

ChrisRackauckas: Yeah I don't think we should ever tell users to use ForwardDiff.value 😅 😅 😅 😅 😅 😅 😅 😅 😅 😅 😅 😅 😅 😅 😅 😅 😅 😅 😅 😅 😅  that's just asking for punishment.

AJ0070:
@ChrisRackauckas Any suggestions on this one?

ChrisRackauckas:
I'm not convinced we actually want to do this.

Issue #911, opened 2026-03-03
Mass matrix ODE not solving anymore after LinearSolve.jl v3.62


JoshuaLampert:
**Describe the bug 🐞**

Since LinearSolve.jl v3.62, I am observing a strange behavior solving some mass-matrix `ODEProblem`s with `Rodas5P` (or other DAE solvers from OrdinaryDiffEqRosenbrock.jl) from my package KernelInterpolation.jl, see, e.g. this PR https://github.com/JoshuaLampert/KernelInterpolation.jl/pull/159. Unfortunately, I have not been able to reproduce the behavior on a simpler problem involving dependencies. The same problem with LinearSolve.jl v3.61 and nothing else changed runs successfully and gives a reasonable solution.
The solution returns without any warning and error, but the resulting object only contains the initial condition. It seems like all time steps get rejected. The MRE below uses adaptive time stepping. I also tried fixed time stepping with low `dt = 1e-6` (this was the lowest time step the solver needed before the update of LinearSolve.jl to v3.62), but then I get an "Instability detected" warning.
I am not sure if I am doing something wrong or if this is a bug in LinearSolve.jl, but I would like to understand the issue and know how to fix it properly.

**Expected behavior**

The MRE below should given output similar to
```julia
#timesteps:     10 │ Δt: 1.1812e-06 │ sim. time: 8.5923e-06 (0.001%)   │ run time: 1.2088e-01 s
#timesteps:     20 │ Δt: 2.5277e-06 │ sim. time: 2.4782e-05 (0.002%)   │ run time: 1.4642e-01 s
#timesteps:     30 │ Δt: 7.6793e-06 │ sim. time: 7.5318e-05 (0.008%)   │ run time: 5.0613e-01 s
#timesteps:     40 │ Δt: 3.6031e-05 │ sim. time: 2.7623e-04 (0.028%)   │ run time: 5.3264e-01 s
#timesteps:     50 │ Δt: 2.0187e-04 │ sim. time: 1.2842e-03 (0.128%)   │ run time: 5.5868e-01 s
#timesteps:     60 │ Δt: 8.7454e-04 │ sim. time: 6.0639e-03 (0.606%)   │ run time: 5.8321e-01 s
#timesteps:     70 │ Δt: 2.2177e-03 │ sim. time: 1.8483e-02 (1.848%)   │ run time: 6.0866e-01 s
[...]

PR #942, opened 2026-04-08
Fix QR fallback truncation for ill-conditioned matrices

ChrisRackauckas-Claude:
## Summary

- Julia's `QRPivoted \ b` applies rank-revealing truncation with threshold `rcond = n * eps`. For ill-conditioned (but full-rank) matrices where `cond(A) > 1/(n*eps)`, this discards meaningful components and produces residuals orders of magnitude worse than LU (e.g., 338 vs 0.32 for a 241×241 matrix with cond ≈ 5e13, detected rank 89/241).
- After the existing truncated QR solve, the fallback now also tries a non-truncated solve (`R\(Q'b)` + permute) using the same factorization — just one extra triangular back-substitution + one matvec for the residual check. Returns whichever of {LU, truncated QR, non-truncated QR} has the lowest residual.
- For `:lu_failure` (genuinely rank-deficient), truncated QR is returned directly as before (non-truncated would divide by near-zero R diagonals).
- Retcode now honestly reflects solution quality: `Success` if the best candidate meets tolerance, `APosterioriSafetyFailure` if not.

## Root cause analysis

The issue reported in #911 was traced to Julia 1.12's `ldiv!` for `QRPivoted` (in `LinearAlgebra/src/qr.jl:568`), which uses `LAPACK.laic1!` for incremental condition estimation with `rcond = n * eps(T)`. For n=241, this gives a condition cutoff of ~1.87e13. The reporter's DAE Jacobian had cond ≈ 5e13, causing Julia to detect rank 89/241 and truncate 152 columns. The truncated minimum-norm solution had residual 338 vs LU's 0.32 — a 1000x regression that caused the ODE solver to blow up.

The non-truncated column-pivoted QR actually gives the *best* residual of all methods (0.16), confirming the matrix was full-rank and the truncation was harmful.

| Method | Residual | Notes |
|--------|----------|-------|
| QR full (non-truncated) | **0.16** | Best for ill-conditioned full-rank |
| LU | 0.32 | |
| QR NoPivot | 0.37 | |
| SVD | 6.09 | Also truncates, but less aggressively |
| **QR truncated (old fallback)** | **338** | Rank detected as 89/241 |

## Test plan

- [x] Retcodes test: rank-deficient matrices still get `Success` via QR fallback
- [x] Default algs test: near-singular matrices get `APosterioriSafetyFailure` (updated from `Success` — more honest retcode)
- [x] Cache reuse test: subsequent solves with same A still use QR, solution matches fresh QR reference
- [x] Well-conditioned matrices: no fallback triggered, `Success` as before
- [x] Ill-conditioned (cond ≈ 5e13): residual matches LU quality (~0.32) instead of 338

### DIFF
--- a/src/LinearSolve.jl
+++ b/src/LinearSolve.jl
@@ -365,7 +365,7 @@ function is_algorithm_available(alg::DefaultAlgorithmChoice.T)
 end
 
 """
-    DefaultLinearSolver(;safetyfallback=true, residualsafety=safetyfallback)
+    DefaultLinearSolver(;safetyfallback=true, residualsafety=false)
 
 The default linear solver. This is the algorithm chosen when `solve(prob)`
 is called. It's a polyalgorithm that detects the optimal method for a given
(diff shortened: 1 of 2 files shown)


ChrisRackauckas-Claude:
## Updated scope

Replaced the dual-solve approach with a simpler fix after discussion:

**Changes:**
- `residualsafety` defaults to `false` (was `true`). LU with partial pivoting always achieves backward error ≈ eps — large forward residuals reflect problem conditioning, not algorithm failure. No fallback can improve on it.
- `safetyfallback` now includes an O(n) NaN/Inf check on the solution, catching near-singular matrices where LU produces non-finite values despite `issuccess=true`.

**Safety layers:**
1. Zero pivot → QR fallback *(existing)*
2. NaN/Inf in solution → QR fallback *(new)*
3. Residual check → opt-in via `residualsafety=true` *(now opt-in)*

The dual-solve comparison (truncated vs non-truncated QR) is preserved in draft PR #943 for reference, but not needed with `residualsafety=false` as default.

Companion PR: SciML/OrdinaryDiffEq.jl#3366 (passes ODE reltol to LinearSolve tolerance for users who opt in to `residualsafety=true`).

PR #948, opened 2026-04-10
fix: fix failures on CI

AJ0070:
## Checklist

## Additional context

This PR fixes Trim test failures on environments where juliac cannot find a system compiler (gcc or clang).

### What changed
- Updated `runtests.jl` to check for a supported C compiler before running juliac-based assertions.
- If no compiler is found, the compile/run portion of the Trim test is skipped instead of failing.

### Why
CI failures were caused by runner environment limitations (missing compiler), not by LinearSolve behavior. This change keeps Trim coverage on capable runners while avoiding false-negative failures on compiler-less runners.


ChrisRackauckas:
This doesn't seem good: if it can't compile, we should error? Which runner is having issues?

AJ0070:
I think its deepsea3-63 (Run trim: Error During Test at /home/chrisrackauckas/github-runners/deepsea3-63/_work/LinearSolve.jl/LinearSolve.jl/test/trim/runtests.jl:85)

ChrisRackauckas:
what was the failure? url?

AJ0070:
1. https://github.com/SciML/LinearSolve.jl/actions/runs/24230391475/job/70740537719?pr=922
2. https://github.com/SciML/LinearSolve.jl/actions/runs/24230391475/job/70740538562?pr=922
3. https://github.com/SciML/LinearSolve.jl/actions/runs/24205956493/job/70661719321

AJ0070:
we can fix these failures by installing C++11 on those machines or we can downgrade python version since 3.14 is new so wheel does not exist and it tries to build it from source(I think we should follow the downgrading python version approach):
1. https://github.com/SciML/LinearSolve.jl/actions/runs/24233724005/job/70751057518
2. https://github.com/SciML/LinearSolve.jl/actions/runs/24233724005/job/70751056103

ChrisRackauckas:
Yes, there's many machines and you cannot assume anything is installed on them

Issue #953, opened 2026-04-13
Precompilation hangs on 3.73?


hersle:
Can anyone reproduce this hang in a fresh environment (on Julia 1.10)?

<img width="1622" height="1070" alt="Image" src="https://github.com/user-attachments/assets/ef1bc2b0-0503-488d-9e16-6c4f1049b644" />

ChrisRackauckas:
What CPU and operating system? Also, if you remove STRUMPACK does it still happen?

ChrisRackauckas:
oh `LinearSolveSTRUMPACKExt = "SparseArrays"` that needs a hotfix to require the JLL @AJ0070 ASAP

hersle:
```julia-repl
julia> versioninfo()
Julia Version 1.10.11
Commit a2b11907d7b (2026-03-09 14:59 UTC)
Build Info:
  Official https://julialang.org/ release
Platform Info:
  OS: Linux (x86_64-linux-gnu)
  CPU: 20 × 12th Gen Intel(R) Core(TM) i7-12800H
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-15.0.7 (ORCJIT, alderlake)
Threads: 1 default, 0 interactive, 1 GC (on 20 virtual cores)
```

ChrisRackauckas:
It's at least hotfixed, needs a Yggdrasil setup in order to fully release it.

AJ0070:
I am currently working on this and will try to fix this as soon as possible

PR #954, opened 2026-04-13
fix: hotfix to stop hang

AJ0070:
temporary fix for #953


ChrisRackauckas:
why not require the JLL?

AJ0070:
I couldn’t find a resolvable STRUMPACK_jll in General from my environment, so I avoided adding an uncertain dep/UUID that could break resolution. For the immediate fix, I removed the SparseArrays-only trigger so LinearSolveSTRUMPACKExt no longer auto-loads and hangs precompile.

AJ0070:
@ChrisRackauckas Should I create STRUMPACK_jll and register it?

ChrisRackauckas:
Yes that needs a Yggdrasil.jl PR

PR #960, opened 2026-04-19
run trim test only on version: "1", arch: x64, os: ubuntu-latest

### LINKED ISSUE #959: Trim CI: Enforce hard-fail policy for missing compiler and define required toolchain target(s)
## Trim CI Failure Policy Discussion

The Trim test group currently fails on CI across macOS, Ubuntu, and Windows with Juliac reporting:
``` could not find a compiler, looked for gcc and clang ```

This raises a question: should missing compiler be treated as a skip condition, or as a hard failure?

---

## Why

- Trim is validating executable compilation behavior, not just runtime correctness.
- Skipping on missing compiler can hide real breakage in CI setup.
- A failing preflight gives faster, actionable signal to maintainers.

---

## Suggested Approach

- Add a compiler preflight for Trim jobs and fail early with a clear message if no compiler is available.
- Define at least one required canonical Trim target (e.g., Linux x64 with pinned toolchain).
- Keep broader cross-platform Trim coverage as optional/non-blocking until toolchain provisioning is stable on all runners.

---

## Open Questions

- Which Trim target(s) should be required for merge gating?
- Should non-canonical OS Trim jobs remain required, optional, or temporarily disabled?
- Do we want explicit compiler pinning in CI, or rely on runner defaults + preflight checks?

AJ0070:
fix #959

### DIFF
--- a/.github/workflows/Tests.yml
+++ b/.github/workflows/Tests.yml
@@ -29,6 +29,7 @@ jobs:
           - "pre"
         arch:
           - x64
+          - x86
         group:
           - "Core"
           - "DefaultsLoading"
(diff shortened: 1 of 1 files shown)


AJ0070:
will ping for review when ready! opened it to see CI behaviour on the changes.

AJ0070:
Sorry for this PR being stale. I am unwell, but I will complete it by the end of the week.

AJ0070:
getting back to this.

AJ0070:
@ChrisRackauckas, the trim test now runs only on the Ubuntu LTS version. It is passing successfully on CI, and I will keep an eye on CI to ensure there are no failures.

AJ0070:
it's still flaky! fixing it

AJ0070:
@ChrisRackauckas now trim test is consistently passing! sorry for the previous ping

PR #964, opened 2026-04-29
Add preconditioning support for LSMR/LSQR/LSLQ/CGLS/CRLS

nathanaelbosch:
## Checklist

## Additional context

I ran into this when using LSMR. Making this PR was faster than writing a thorough issue as this seems like an obvious fix, but **I have not yet thoroughly tested this yet!** If more examples and tests are wanted I can turn this into a proper PR by the end of next week.

I could also imagine that the same issue holds for other methods too but I also didn't carefully check this yet.

EDIT: I added tests and extended the contribution to the other least-squares methods.


ChrisRackauckas:
Yes it would be nice to get a test on this.

PR #968, opened 2026-05-02
Fix PETSc CSR precompile crash by moving CSR hooks into LinearSolvePETScExt

AJ0070:
## Checklist

### Issue
CI precompile failed with `FieldError: type Nothing has no field is_sparse_petsc` when loading the CSR PETSc extension.

### Root Cause
A sibling extension (`LinearSolvePETScCSRExt`) attempted to extend the PETSc hooks before the main PETSc extension was initialized, creating a fragile load-order dependency.

### Change
- Moved CSR-specific PETSc hooks into `LinearSolvePETScExt.jl`
- Registered `SparseMatricesCSR` under the `LinearSolvePETScExt` extension in `Project.toml`
- Implemented CSR conversions and in-place update helpers directly in the main PETSc extension:
  - `is_sparse_petsc`
  - `to_petsc_mat`
  - `store_sparse_pattern!`
  - `check_pattern_changed`
  - `update_sparse_values!`
  
 AI disclosure: Used copilot to figure out the root cause of the failure and verify the correctness of the fix.


AJ0070:
investigating PETSc failures

AJ0070:
since the major code diff is related to PETSc failure, pivoting PR title and description according to it.

AJ0070:
@ChrisRackauckas what do you think about the fixing approach?

Project.toml
```diff
@@ -83,8 +83,7 @@ LinearSolveKernelAbstractionsExt = "KernelAbstractions"
 LinearSolveKrylovKitExt = "KrylovKit"
 LinearSolveMetalExt = "Metal"
 LinearSolveMooncakeExt = "Mooncake"
-LinearSolvePETScCSRExt = ["PETSc", "SparseArrays", "SparseMatricesCSR"]
-LinearSolvePETScExt = ["PETSc", "SparseArrays"]
+LinearSolvePETScExt = ["PETSc", "SparseArrays", "SparseMatricesCSR"]
```
ChrisRackauckas, line 86: If we do this, then `SparseMatricesCSR` is required. That is fine, but it should get a mention in the error messages because people are likely to not know that `using PETSc` is not sufficient.

AJ0070:
@ChrisRackauckas updated it!

ChrisRackauckas (approved):

ChrisRackauckas:
It would be good to figure out what's going on in those Pres so we can relay this to the Julia team for the upcoming v1.13 release.

PR #970, opened 2026-05-05
PETSc: complete GSoC weeks 1–4 for multi-rank SparseMatrixCSC support

AJ0070:
## Checklist

## Additional context

This PR completes the GSoC proposal weeks 1–4 PETSc deliverable for plain `SparseMatrixCSC` inputs on a multi-rank communicator.

Included in this PR:
- extend `PETScCache` with `rstart` and `rend`, initialized to `-1`
- remove the serial-only communicator guard for standard Julia matrices
- add a row-owned MPI AIJ assembly path for replicated `SparseMatrixCSC` inputs
- create matrix-compatible distributed PETSc vectors for this path with `MatCreateVecs`
- copy each rank's owned RHS slice into PETSc before `KSPSolve`
- gather the full PETSc solution back onto every rank with `VecScatterCreateToAll`
- keep the existing `PSparseMatrix` / `PVector` PETSc MPI path working
- add MPI tests covering ownership, assembly, basic distributed solve, and repeated solve with a new RHS
- revert unrelated HYPRE and PartitionedSolvers scaffolding changes from this branch
- update the GSoC progress report to reflect the current implementation state

AI Disclosure: Used Codex 5.3 to  get review of my code, write test and PR description in markdown.

### DIFF
--- a/ext/LinearSolvePETScExt.jl
+++ b/ext/LinearSolvePETScExt.jl
@@ -580,31 +580,27 @@ function run_ksp!(pcache, petsclib, alg, b::AbstractVector, u::AbstractVector)
     end
 end
 
-function run_ksp_mpi!(pcache, petsclib, alg, b::AbstractVector, u::AbstractVector)
-    petsc_x = pcache.petsc_x
-    petsc_b = pcache.petsc_b
-    local_range = (pcache.rstart + 1):pcache.rend
-
-    local_b = LibPETSc.VecGetArray(petsclib, petsc_b)
+function copyto_local_petsc_vec!(petsclib, petsc_vec, values::AbstractVector)
+    local_values = LibPETSc.VecGetArray(petsclib, petsc_vec)
     try
-        copyto!(local_b, view(b, local_range))
+        copyto!(local_values, values)
     finally
-        LibPETSc.VecRestoreArray(petsclib, petsc_b, local_b)
+        LibPETSc.VecRestoreArray(petsclib, petsc_vec, local_values)
     end
+    return nothing
+end
 
-    local_x = LibPETSc.VecGetArray(petsclib, petsc_x)
+function copyfrom_local_petsc_vec!(dest::AbstractVector, petsclib, petsc_vec)
+    local_values = LibPETSc.VecGetArray(petsclib, petsc_vec)
     try
-        copyto!(local_x, view(u, local_range))
+        copyto!(dest, local_values)
     finally
-        LibPETSc.VecRestoreArray(petsclib, petsc_x, local_x)
+        LibPETSc.VecRestoreArray(petsclib, petsc_vec, local_values)
     end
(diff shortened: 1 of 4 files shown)


AJ0070:
@ChrisRackauckas

ext/LinearSolvePETScExt.jl
```diff
@@ -576,6 +580,56 @@ function run_ksp!(pcache, petsclib, alg, b::AbstractVector, u::AbstractVector)
...
+        copyto!(local_x, view(u, local_range))
+    finally
+        LibPETSc.VecRestoreArray(petsclib, petsc_x, local_x)
+    end
+
+    if alg.transposed
+        LibPETSc.KSPSolveTranspose(petsclib, pcache.ksp, petsc_b, petsc_x)
+    else
+        LibPETSc.KSPSolve(petsclib, pcache.ksp, petsc_b, petsc_x)
+    end
```
ChrisRackauckas, line 599: These try/finally worry me. What is our error handling here? We should be checking if the solution is right and setting retcodes.

AJ0070:
@ChrisRackauckas fixed it

ChrisRackauckas:
see the comment

AJ0070:
@ChrisRackauckas I want to make sure I’m interpreting your comment correctly. Do you mean that for this new PETSc MPI solve path we should add explicit LinearSolve-level post-solve validation and `retcode` handling after `KSPSolve`, rather than just doing the PETSc array cleanup and relying on PETSc convergence state? If so, should “checking if the solution is right” here mean using the standard LinearSolve residual-safety check on the returned solution, or did you have a different validation/retcode pattern in mind for PETSc?

ChrisRackauckas:
You're try/catching to mask errors. This means the result of solve may not actually be solve. The errors need to be caught and handled, not just masked.

AJ0070:
@ChrisRackauckas Should I do the same for PVector/PSparseMatrix extension path?

ChrisRackauckas:
yes

AJ0070:
@ChrisRackauckas

ChrisRackauckas:
The try/catch/finally handling?

AJ0070:
@ChrisRackauckas Sorry for the late response, my laptop was at the repair shop since my keyboard was not working properly. The solve path is not using `catch` to mask errors. What remains is `try/finally` for PETSc resource cleanup only: restoring arrays obtained from `VecGetArray`, and destroying temporary PETSc `Vec` / scatter objects. If `KSPSolve` or the MPI gather path throws, that exception still propagates out of `solve!`; it is not converted into a normal `LinearSolution`. The explicit `retcode` handling only happens in the normal non-exception path after PETSc returns control. The only swallowed `catch` left is in cache cleanup/finalizer code.

AJ0070:
Thanks!

Issue #972, opened 2026-05-10
DualLinearCache: MethodError on cache reuse with sparse Dual A + Float64 b


briochemc:
## Summary

`LinearSolveForwardDiffExt.setb!` errors with `MethodError: no method matching map!(::partial_vals, ::Nothing, ::Vector{Float64})` when a `DualLinearCache` constructed from sparse Dual `A` + Float64 `b` is reused by mutating `cache.b` to another Float64 vector — the exact pattern Newton-style solvers (e.g. `NonlinearSolve.NewtonRaphson`) use between iterations.

The first `solve!` succeeds (the splitting only iterates `partials_A_list`, since `partials_b_list` is `nothing`). The next `setb!` then unconditionally calls `partial_vals!(getfield(dc, :partials_b), b)` at [`ext/LinearSolveForwardDiffExt.jl:485`](https://github.com/SciML/LinearSolve.jl/blob/master/ext/LinearSolveForwardDiffExt.jl#L485), i.e. `partial_vals!(::Nothing, ::Vector{Float64})` → `map!(partial_vals, nothing, …)` → MethodError.

This was hit in the wild via `NonlinearSolve.NewtonRaphson(linsolve = UMFPACKFactorization())` on a `NonlinearProblem` with sparse Float64 `jac_prototype` under `ForwardDiff.hessian` (nested Duals on `prob.p`, which makes the inner Jacobian Dual-typed against a Float64 residual).

## MWE

```julia
using LinearSolve, ForwardDiff, SparseArrays, LinearAlgebra

function f(p)
    A = sparse(Diagonal(p))         # A becomes Dual under ForwardDiff
    b1 = [1.0, 2.0]                 # Float64 (no partials)
    prob = LinearProblem(A, b1)
    cache = init(prob)
    u1 = copy(solve!(cache).u)      # first solve OK
    cache.b = [3.0, 4.0]            # ← swap b (Float64 → Float64)
    u2 = copy(solve!(cache).u)      # ← MethodError fires here
    return u1 .+ u2
end

@show f([1.0, 2.0])                 # primal call works
J = ForwardDiff.jacobian(f, [1.0, 2.0])
@show J
```

Output:

```julia
f([1.0, 2.0]) = [4.0, 3.0]
ERROR: MethodError: no method matching map!(::typeof(LinearSolveForwardDiffExt.partial_vals), ::Nothing, ::Vector{Float64})
[...]

PR #975, opened 2026-05-11
Guard setu! against TypeError on typed dual_u field (#974)

### LINKED ISSUE #974: DualLinearCache.setu!: TypeError on typed dual_u field, sibling to #972
## Summary

Follow-up to #972 / #973. After v3.76.0 fixed the `partial_vals!(::Nothing, …)` failure in `LinearSolveForwardDiffExt.setb!`, the same family of bug still fires in `setu!` on the nested-Dual (Hessian) path: the unconditional `setfield!(dc, :dual_u, u)` at [`ext/LinearSolveForwardDiffExt.jl:503`](https://github.com/SciML/LinearSolve.jl/blob/v3.76.0/ext/LinearSolveForwardDiffExt.jl#L503) rejects a `Vector{Float64}` `u` when `dc.dual_u` is statically typed `Vector{<:ForwardDiff.Dual}` from an outer Hessian tag.

This is hit by `NonlinearSolve.NewtonRaphson(linsolve = UMFPACKFactorization())` on a `NonlinearProblem` with sparse Float64 `jac_prototype` under `ForwardDiff.hessian` (nested Duals on `prob.p`). The single-Dual gradient case works fine (the cache fields end up typed `Vector{Float64}`), only the Hessian's outer-Dual layer triggers the mismatch.

Concretely: `NonlinearSolveBase.set_lincache_u!` → `setproperty!(dc, :u, u)` → `setu!(dc, u)` → `setfield!(dc, :dual_u, ::Vector{Float64})` → `TypeError`.

## MWE

```julia
using LinearSolve, NonlinearSolve, ForwardDiff, SparseArrays, LinearAlgebra

# Out-of-place residual + populated sparse Float64 `jac_prototype` are what
# trigger it; an in-place residual or empty-prototype path goes through
# different code and does not hit this.
const Tmat = sparse(Diagonal(ones(2)))
f(u, p)   = (Tmat + I / p[1]) * u .- p[2]
jac(u, p) = Tmat + I / p[1]

function obj(p)
    jp = jac(zeros(2), ForwardDiff.value.(p))    # populated sparse Float64
    fn = NonlinearFunction(f; jac = jac, jac_prototype = jp)
    prob = NonlinearProblem(fn, zeros(2), p)
    sol  = solve(prob, NewtonRaphson(linsolve = UMFPACKFactorization()); abstol = 1e-12)
    return sum(sol.u)
end

@show obj([1.0, 2.0])                            # primal OK
@show ForwardDiff.gradient(obj, [1.0, 2.0])      # OK in v3.76.0
@show ForwardDiff.hessian(obj, [1.0, 2.0])       # ← TypeError here
```

Output:

```
TypeError: in setfield!, expected Vector{ForwardDiff.Dual{ForwardDiff.Tag{typeof(obj), Float64}, Float64, 2}}, got a value of type Vector{Float64}
  [1] setu!(dc::DualLinearCache, u::Vector{Float64})
      @ LinearSolveForwardDiffExt ext/LinearSolveForwardDiffExt.jl:503
  [2] setproperty!(dc::DualLinearCache, sym::Symbol, val::Vector{Float64})
      @ LinearSolveForwardDiffExt ext/LinearSolveForwardDiffExt.jl:549
  [3] set_lincache_u!(cache, u)
      @ NonlinearSolveBase src/linear_solve.jl:117
```

## Suggested fix

Guard `setu!` analogously to how #973 guarded `setb!`'s `partial_vals!` — but here the unconditional call that fires is the `setfield!` on `:dual_u`, so the guard belongs on both `setfield!` and `partial_vals!`:

```julia
function setu!(dc::DualLinearCache, u)
    prop = nodual_value!(getproperty(dc.linear_cache, :u), u)
    setproperty!(dc.linear_cache, :u, prop)
    get_dual_type(u) === nothing && return nothing   # ← skip when u carries no partials
    setfield!(dc, :dual_u, u)
    return partial_vals!(getfield(dc, :partials_u), u)
end
```

Note that the same `setfield!(dc, :dual_X, X)` pattern still sits unguarded in `setb!` at [line 488](https://github.com/SciML/LinearSolve.jl/blob/v3.76.0/ext/LinearSolveForwardDiffExt.jl#L488) and `setA!` at [line 473](https://github.com/SciML/LinearSolve.jl/blob/v3.76.0/ext/LinearSolveForwardDiffExt.jl#L473). They don't fire in the cases I've hit (the residual `b` and operator `A` already carry the outer-Dual tag by the time those setters run), but the same fix-shape would harden them. Probably worth applying uniformly.

## Versions

- LinearSolve 3.76.0 (with #973)
- NonlinearSolve 4.19.0
- NonlinearSolveBase 2.x
- ForwardDiff 1.3.3
- Julia 1.12.6

Note this issue was submitted by Claude.

ChrisRackauckas-Claude:
> ⚠️ Please ignore this PR until reviewed by @ChrisRackauckas.

## Summary

Fixes #974.

Follow-up to #972 / #973. After #973 fixed the `partial_vals!(::Nothing, …)` failure path, the same family of bug still fires in `setu!` on the nested-Dual (Hessian) layer: the unconditional `setfield!(dc, :dual_u, u)` rejects a `Vector{Float64}` `u` when `dc.dual_u` is statically typed `Vector{<:Dual}` from an outer Hessian tag.

This is hit by `NonlinearSolve.NewtonRaphson(linsolve = UMFPACKFactorization())` on a `NonlinearProblem` with sparse Float64 `jac_prototype` under `ForwardDiff.hessian`: `NonlinearSolveBase.set_lincache_u!` → `setproperty!(dc, :u, u)` → `setu!(dc, u)` → `setfield!(dc, :dual_u, ::Vector{Float64})` → `TypeError`.

## Fix

Move the `get_dual_type(u) === nothing` early-return ahead of `setfield!` in `setu!`, so that calls with a non-Dual `u` no-op cleanly instead of TypeError-ing on the typed `dual_u` field. `dual_u` is rewritten in place by `linearsolve_dual_solution!` on the next `solve!`, so the skipped assignment loses no information.

## Test

Adds a regression test in `test/forwarddiff_overloads.jl` (`@testset \"DualLinearCache setu! with non-Dual u (#974)\"`) that:

1. Builds a `DualLinearCache` from sparse Dual `A` + Float64 `b` so `dual_u` is statically typed `Vector{<:Dual}`.
2. Triggers the exact path hit by `NonlinearSolveBase.set_lincache_u!` (`cache.u = Vector{Float64}`).
3. Confirms a subsequent `solve!` still produces the correct dual solution.

I confirmed the new test reproduces the `TypeError` against the unpatched `setu!` (1 errored) and passes after the fix (3/3, 61/61 in `forwarddiff_overloads.jl`).

The Hessian MWE in the issue progresses past `setu!` after this fix but then hits a separate `set_du!` type-mismatch inside `NonlinearSolveBase.NewtonDescent` (the descent cache's `δu::Vector{Float64}` rejecting a `Vector{<:Dual}` returned from the linear solve). That is a NonlinearSolveBase concern and out of scope here.

## Test plan

- [x] `forwarddiff_overloads.jl` passes locally on Julia 1.12.6 (61/61 total; the new testset is 3/3).
- [x] Confirmed the new test errors on the unpatched `setu!` (stashed the fix, re-ran — 1 errored as expected).
- [x] No changes to behavior when `u` carries Duals.
- [x] Runic-formatted.

### DIFF
--- a/ext/LinearSolveForwardDiffExt.jl
+++ b/ext/LinearSolveForwardDiffExt.jl
@@ -498,10 +498,12 @@ function setu!(dc::DualLinearCache, u)
     prop = nodual_value!(getproperty(dc.linear_cache, :u), u) # Update in-place
     setproperty!(dc.linear_cache, :u, prop) # Does additional invalidation logic etc.
 
-    # Update partials only when u actually carries Duals; otherwise there is
-    # nothing to extract and the partials slot may be unallocated.
-    setfield!(dc, :dual_u, u)
+    # Skip dual_u when u carries no Duals: dual_u is statically typed
+    # Vector{<:Dual} from construction, so setfield! with a Vector{Float64}
+    # would TypeError. dual_u is rewritten in place by linearsolve_dual_solution!
+    # on the next solve!, so the assignment is unnecessary in that case.
     get_dual_type(u) === nothing && return nothing
+    setfield!(dc, :dual_u, u)
     return partial_vals!(getfield(dc, :partials_u), u)
 end
 
--- a/ext/LinearSolveForwardDiffExt.jl
+++ b/ext/LinearSolveForwardDiffExt.jl
@@ -508,9 +508,7 @@ function setu!(dc::DualLinearCache{DT}, u) where {DT}
         # `linearsolve_dual_solution!`.
         dual_u_field = getfield(dc, :dual_u)
         if dual_u_field isa AbstractArray
-            @inbounds for i in eachindex(dual_u_field, u)
-                dual_u_field[i] = DT(u[i])
-            end
+            dual_u_field .= DT.(u)
         else
             setfield!(dc, :dual_u, DT(u))
         end
--- a/ext/LinearSolveForwardDiffExt.jl
+++ b/ext/LinearSolveForwardDiffExt.jl
@@ -493,16 +493,32 @@ function setb!(dc::DualLinearCache, b)
     # Invalidate cache (if setting A or b)
     return setfield!(dc, :rhs_cache_valid, false)
 end
-function setu!(dc::DualLinearCache, u)
+function setu!(dc::DualLinearCache{DT}, u) where {DT}
     # Put the Dual-stripped versions in the LinearCache
     prop = nodual_value!(getproperty(dc.linear_cache, :u), u) # Update in-place
     setproperty!(dc.linear_cache, :u, prop) # Does additional invalidation logic etc.
 
-    # Skip dual_u when u carries no Duals: dual_u is statically typed
-    # Vector{<:Dual} from construction, so setfield! with a Vector{Float64}
-    # would TypeError. dual_u is rewritten in place by linearsolve_dual_solution!
-    # on the next solve!, so the assignment is unnecessary in that case.
-    get_dual_type(u) === nothing && return nothing
+    if get_dual_type(u) === nothing
+        # `u` is primal-only (e.g. the Vector{Float64} iterate handed in by
+        # `NonlinearSolveBase.set_lincache_u!` during Newton iterations under
+        # an outer Hessian tag), while `dual_u` is statically typed
+        # Vector{<:Dual}. Promote element-wise to `DT` with zero partials so
+        # the field invariant is preserved without dropping derivatives — the
+        # next solve! will rewrite the partials from the Dual A / b via
+        # `linearsolve_dual_solution!`.
+        dual_u_field = getfield(dc, :dual_u)
+        if dual_u_field isa AbstractArray
+            @inbounds for i in eachindex(dual_u_field, u)
+                dual_u_field[i] = DT(u[i])
+            end
+        else
+            setfield!(dc, :dual_u, DT(u))
+        end
+        pu = getfield(dc, :partials_u)
+        pu === nothing || fill!(pu, zero(eltype(pu)))
+        return nothing
+    end
+
     setfield!(dc, :dual_u, u)
     return partial_vals!(getfield(dc, :partials_u), u)
 end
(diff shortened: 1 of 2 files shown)


ChrisRackauckas-Claude:
Good catch — pushed an update that switches to type-promotion.

**How the case arises:** `NonlinearSolveBase.set_lincache_u!` (called between Newton iterations) passes the *primal* iterate down with `cache.u = primal_u`. Under outer `ForwardDiff.hessian`, the outer-tag Dual specialises `DualLinearCache.dual_u` against `Vector{Dual{outer_tag, …}}`, but the outer NonlinearSolve cache's iterate type is `Vector{Float64}` (since `NonlinearProblem(fn, zeros(2), p)` was constructed with a Float64 `u0`). So `setu!` receives a `Vector{Float64}` against a `Vector{<:Dual}` field.

**You're right that skipping was wrong.** Even though `solve!` would later overwrite `dual_u` in place via `linearsolve_dual_solution!` so the final `sol.u` carries the correct duals, between `setu!` and the next `solve!` the field invariant was broken and the message was "drop the partials" rather than "promote u to DT." Replaced the early-return with element-wise `dual_u[i] = DT(u[i])` (zero partials, since a Float64 iterate has none to extract). `partials_u` is zeroed for consistency.

**Test extended** to cross-check `sol.u`'s partials against `Diagonal(p) \ b` after the `cache.u = …` mutation, not just the values (`≈` on Dual only compares the `.value`). Confirmed 66/66 passing locally on Julia 1.12.6, and that the new partials assertion would have caught the dropped-derivatives case.

Force-pushed after rebasing on the latest main; only my own commits, no other contributors on the branch.

PR #982, opened 2026-05-13
HYPRE: Tests and Docs (GSoC week 9)

AJ0070:
## Checklist

## Additional context

Included in this PR:
- add a 2-rank MPI test path for automatic distributed construction from plain Julia sparse matrices and vectors
- add explicit HYPRE non-convergence `retcode` handling
- add failure-mode coverage for both the serial HYPRE path and the communicator-based MPI path
- update the `HYPREAlgorithm` docstring and the main solver docs to describe the serial and communicator-based HYPRE workflows

AI Disclosure: Used GPT 5.4

### DIFF
--- a/docs/src/solvers/solvers.md
+++ b/docs/src/solvers/solvers.md
@@ -409,55 +409,6 @@ For the MPI workflow, `sol.u` is a distributed `HYPREVector` holding the owned l
 each rank. Unlike the PETSc `SparseMatrixCSC` MPI path, the full Julia solution is not
 replicated back onto every rank automatically.
 
-#### Basic usage
-
-```julia
-using LinearSolve, HYPRE, SparseArrays
-
-HYPRE.Init()
-
-n = 8
-A = spdiagm(0 => collect(2.0:9.0))
-b = collect(2.0:9.0)
-
-prob = LinearProblem(A, b)
-sol = solve(prob, HYPREAlgorithm(HYPRE.PCG))
-```
-
-#### MPI auto-construction from plain Julia arrays
-
-```julia
-using LinearSolve, HYPRE, MPI, SparseArrays
-
-MPI.Init()
-HYPRE.Init()
-
-comm = MPI.COMM_WORLD
-n = 20
-A = spdiagm(0 => collect(1.0:n))
-b = 2.0 .* collect(1.0:n)
-
-prob = LinearProblem(A, b)
-sol = solve(prob, HYPREAlgorithm(HYPRE.PCG; comm = comm))
-
-# sol.u is a distributed HYPREVector over the owned local rows on each rank.
-local_x = Vector{Float64}(undef, sol.u.iupper - sol.u.ilower + 1)
-copy!(local_x, sol.u)
-```
-
-If the matrix structure is unchanged and only the right-hand side changes, reuse the cached
-distributed HYPRE objects:
-
-```julia
-using SciMLBase
-
-cache = SciMLBase.init(prob, HYPREAlgorithm(HYPRE.PCG; comm = comm))
-sol = solve!(cache)
-
-b2 = 3.0 .* collect(1.0:n)
-cache.b = b2
-sol = solve!(cache)
-```
 
 ```@docs
 HYPREAlgorithm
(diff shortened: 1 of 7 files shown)


docs/src/solvers/solvers.md
```diff
@@ -384,6 +384,80 @@ KrylovKitJL
...
 
+!!! note
+
+    Initialize HYPRE before solving:
+    ```julia
+    using HYPRE
+    HYPRE.Init()
+    ```
+
+[HYPRE](https://computing.llnl.gov/projects/hypre-scalable-linear-solvers-multigrid-methods)
```
AJ0070: Is this link fine? or should I add https://github.com/fredrikekre/HYPRE.jl ?

AJ0070:
@ChrisRackauckas If you think a focused tutorial would be helpful here, I can add one here as well.

docs/src/solvers/solvers.md
ChrisRackauckas: both

ChrisRackauckas:
Well you have to open Julia with `mpirun` and all of that right? So it would be good to have a whole tutorial that shows this, and even a short video to put to julialang would be helpful since it requires a non-standard execution model.

AJ0070:
@ChrisRackauckas That makes sense. I'll add a focused HYPRE MPI tutorial that shows the full launch, setup and solve workflow with `mpirun`, and I can prepare a demo video going through the same tutorial.

docs/src/solvers/solvers.md
AJ0070: added both!

test/hypretests_mpi.jl
```diff
@@ -123,3 +123,14 @@ sol = solve!(cache)
 @test sol.iters > 0
 copy!(local_sol, sol.u)
 @test local_sol ≈ fill(3.0, local_size)
+
+# Automatic distributed construction failure retcode
+A_fail = spdiagm(-1 => -ones(19), 0 => 2.0 .* ones(20), 1 => -ones(19))
+b_fail = ones(20)
+sol = solve(
+    LinearProblem(A_fail, b_fail), HYPREAlgorithm(HYPRE.PCG; comm = comm);
```
ChrisRackauckas: why would this fail?

AJ0070: The failure is intentional from the iteration limit, not from a bad matrix. I tightened the test to make that explicit: it now uses a larger SPD 1D Laplacian and `maxiters = 1`, so the expected outcome is that PCG does not converge in a single iteration and we return a non-success `retcode` instead of throwing.

ChrisRackauckas: Oh I missed that it was PCG

AJ0070:
@ChrisRackauckas If tutorial looks good to you, I'll start working on the video.

docs/src/solvers/solvers.md
```diff
@@ -384,6 +384,81 @@ KrylovKitJL
...
+  LinearSolve construct `HYPREMatrix` / `HYPREVector` values automatically.
+- MPI auto-construction: pass a communicator with `comm = MPI.COMM_WORLD` (or another MPI
+  communicator), and LinearSolve will split a plain Julia matrix/vector into contiguous local
+  row blocks before constructing distributed HYPRE objects.
+
+For the MPI workflow, `sol.u` is a distributed `HYPREVector` holding the owned local rows on
+each rank. Unlike the PETSc `SparseMatrixCSC` MPI path, the full Julia solution is not
+replicated back onto every rank automatically.
+
+#### Basic usage
```
ChrisRackauckas: too much

docs/src/solvers/solvers.md
```diff
@@ -384,6 +384,81 @@ KrylovKitJL
...
+```julia
+using SciMLBase
+
+cache = SciMLBase.init(prob, HYPREAlgorithm(HYPRE.PCG; comm = comm))
+sol = solve!(cache)
+
+b2 = 3.0 .* collect(1.0:n)
+cache.b = b2
+sol = solve!(cache)
+```
```
ChrisRackauckas:
```suggestion
```

If the matrix structure is unchanged and only the right-hand side changes, reuse the cached
distributed HYPRE objects:

```julia
using SciMLBase

cache = SciMLBase.init(prob, HYPREAlgorithm(HYPRE.PCG; comm = comm))
sol = solve!(cache)

b2 = 3.0 .* collect(1.0:n)
cache.b = b2
sol = solve!(cache)
```
```

AJ0070:
Thanks!

PR #983, opened 2026-05-13
Fix CUDSS extension for CUDA v6.x

albertomercurio:
## Checklist

## Additional context

Given the recent updates of CUDA.jl and CUDSS.jl, we need these small changes in the extension here.


ytdHuang:
By the way, we also need a version release after this PR.

ChrisRackauckas:
```
Test with CUDSS loaded: Test Failed at /home/chrisrackauckas/github-runners/demeter4-2/_work/LinearSolve.jl/LinearSolve.jl/test/gpu/cuda.jl:51
  Expression: LinearSolve.cudss_loaded(A_gpu_csr)
```

interesting, it didn't load?

ytdHuang:
@albertomercurio @ChrisRackauckas 

It seems like the GPU test is still using `CUDA v5.11.2` but with `CUDSS v0.7.0`.

So when we call :

```julia
using CUDA.CUSPARSE # from CUDA v5

using CUDSS
CUDSS.cuSPARSE # from cuSPARSE v6
```

it is indeed different I think

ytdHuang:
By the way, would it be better that the `Project.toml` is setup as:

```
[extensions]
LinearSolveCUDAExt = ["CUDACore", "cuSPARSE"]
```

instead of `"CUDA"` ?

ChrisRackauckas:
Possibly, I haven't looked too much at the recent changes.

ytdHuang:
I tried it locally, but the compat version cannot be resolved.

Because `CUSOLVERRF.jl` only support until `CUDA v5`.

ChrisRackauckas:
So we need to drop or update that? Is CUSOLVERRF.jl still maintained?

ytdHuang:
I opened an issue in `CUSOLVERRF` today: https://github.com/exanauts/CUSOLVERRF.jl/issues/27

But the last version release for `CUSOLVERRF` was Apr. 18, 2025.

===

I think the best solution for now is to drop `CUSOLVERRF`, and use `cuSOLVER` directly instead of `CUDA`.

===

I tried to fix all `CUDA v6`/`CUDSS v0.7` related stuffs in PR #986

PR #985, opened 2026-05-16
PartitionedSolvers: Add PartitionedSolversAlgorithm scaffold (GSoC week 10 and 11)

AJ0070:
## Checklist

## Additional context

Included in this PR:
- add `PartitionedSolvers` as a weak dependency with compat and extension wiring
- add `PartitionedSolversAlgorithm` to `extension_algs.jl` and export it
- add `ext/LinearSolvePartitionedSolversExt.jl`
- add a cache scaffold for the future PartitionedSolvers-backed solve path
- validate that `PartitionedSolversAlgorithm` is only used with `PSparseMatrix` / `PVector`
- provide an explicit stub `solve!` boundary that marks the full solve path as week 12–13 work
- add a direct structural test file covering extension loading, input validation, and the stub-solve boundary
- add a solver-docs entry describing the current scaffold status

Update on tutorial: I have recorded it and currently editing it.

AI Disclosure: Used GPT 5.4


AJ0070:
@ChrisRackauckas I looked into the `NoPre` failure, and my current understanding is that it is indirect rather than a regression in the solver logic itself: the failing test is `test/nopre/static_arrays.jl`, but the stderr warning is coming from the `Mooncake` / `LLVM` stack (`lookup(jljit::JuliaOJIT, name, external_jd_only = false) is deprecated ...`). In `test/runtests.jl`, the `NoPre` job activates `test/nopre`, then does `Pkg.develop(PackageSpec(path = dirname(@__DIR__)))` and `Pkg.instantiate()`, so after the root `Project.toml` changes in this branch, that environment gets re-resolved against the current root project, which appears to move it to a newer `Mooncake` / `LLVM` combination that emits this warning; `static_arrays.jl` then fails because it currently treats any stderr output as a test failure even though the solve itself still works. So I think this branch can still be the trigger, but through dependency re-resolution rather than because the StaticArrays path is broken. The approaches I see are: keep the `NoPre` environment pinned more strictly so root project changes do not perturb the `Mooncake` / `LLVM` stack, intentionally accept the updated `NoPre` resolution and patch `static_arrays.jl` to ignore just this specific external LLVM deprecation while still failing on any other stderr output, or tighten compat so the problematic `Mooncake` / `LLVM` combination is avoided entirely. I currently lean toward the second option as the narrowest fix, but I wanted your opinion before pushing that approach.

AJ0070:
@ChrisRackauckas

docs/src/solvers/solvers.md
```diff
@@ -414,6 +414,25 @@ replicated back onto every rank automatically.
...
+
+!!! note
+
+    Using this integration requires loading `LinearSolve.jl`, `PartitionedArrays.jl`, and the
+    `PartitionedSolvers` subproject:
+    ```julia
+    using LinearSolve, PartitionedArrays, PartitionedSolvers
+    ```
+
+This integration is currently in the week 10--11 scaffold stage. The
```
ChrisRackauckas: ?

AJ0070: In this PR I have added stubs for Partitioned solvers, So this will be updated it in the next PR

AJ0070: removed it.

ChrisRackauckas:
Dig into the hypre test failure

AJ0070:
@ChrisRackauckas I investigated this and it turned out to be a test flakiness in the `PCG + ParaSails` path rather than a regression in the distributed-construction code. The failure is coming from the cached `solve!` branch in `test_interface`, and I was able to reproduce the same `HYPRE_ParCSRPCGSetup returned error code 1` locally on the current generic SPD matrix family used by `generate_probs` (`sprand(..., 0.01) + 3I; A = A' * A`). I then stress-checked that combination across multiple seeds and saw the same ParaSails setup failure recur. When I switched the ParaSails case to a denser, more strongly regularized SPD family matching what HYPRE.jl’s own ParaSails test uses (`sprand(..., 0.05); A = A' * A + 5I`), the failures disappeared. While tracing it, I also found that the direct `solve(...)` branch in `test_interface` was hardcoding `Pl = HYPRE.BoomerAMG`, so the direct and cached branches were not actually exercising the same preconditioner. I fixed both issues in `test/hypretests.jl`: the ParaSails case now uses a ParaSails-friendly matrix family, and the direct solve path now uses the same requested `Pl` as the cached path instead of hardcoding BoomerAMG. It passes locally now.

AJ0070:
still getting failure on CI, figuring it out

ChrisRackauckas:
okay well this PR seems fine, so isolate the HYPRE issue in another PR.

AJ0070:
Okay!

PR #986, opened 2026-05-16
Support `CUDA v6`/`CUDSS v0.7` and use `cuSOLVER` as dependency instead of `CUDA`

ytdHuang:
## Checklist

This PR directly set `cuSOLVER` as dependency for the extension instead of the entire `CUDA` package.

It still works if users load `CUDA v6`, since it internally loads `cuSOLVER`, too.

This PR also includes the changes in PR #983


ytdHuang:
Note that the admins haven't decide whether to drop the extension of [`CUSOLVERRF.jl`](https://github.com/exanauts/CUSOLVERRF.jl) or not yet.

So currently, I only remove `CUSOLVERRF` in `test/gpu/Project.toml`

PR #994, opened 2026-05-27
fix HYPRE test failure

AJ0070:
Fix test failures.

NOTE: I opened this to see the CI behavior and will ping once it's ready to review.


AJ0070:
@ChrisRackauckas I traced this to HYPRE ParaSails setup itself rather than the LinearSolve wrapper: the SPD ParaSails path calls LP64 LAPACK symbols like `dpotrf_`/`dpotrs_`, and even with `Sym = 0` the nonsymmetric path can call `dgels_`; on the failing CI images those LP64 symbols are not backed through libblastrampoline, so HYPRE errors before LinearSolve behavior is tested. I changed the ParaSails test to use a deterministic SPD matrix and explicit ParaSails settings, then added a small direct HYPRE capability probe that runs the LinearSolve ParaSails interface tests only when HYPRE itself can set up PCG+ParaSails in the current process, otherwise it skips just those ParaSails cases. Does this seem like the right test-level handling, or would you prefer forcing an LP64 LAPACK backend in CI instead?

PR #1004, opened 2026-06-05
Add PureUMFPACKFactorization extension (PureUMFPACK.jl)

ChrisRackauckas-Claude:
## Summary

Adds an opt-in `PureUMFPACKFactorization <: AbstractSparseFactorization` that wraps the pure-Julia UMFPACK port [PureUMFPACK.jl](https://github.com/SciML/PureUMFPACK.jl), mirroring the existing SuiteSparse `UMFPACKFactorization` and the pure-Julia `PureKLUFactorization`.

- `PureUMFPACK` is a **weak dependency**; the algorithm is provided by a new `LinearSolvePureUMFPACKExt` extension and is only available after `using PureUMFPACK`.
- The SuiteSparse `UMFPACKFactorization` and the default polyalgorithm are **unchanged** — this is purely additive/opt-in.

## Details

- `src/factorization.jl`: `PureUMFPACKFactorization(; reuse_symbolic = true, check_pattern = true)` struct + docstring + generic `init_cacheval(...) = nothing` stub, placed next to `KLUFactorization`/`PureKLUFactorization`.
- `src/LinearSolve.jl`: export `PureUMFPACKFactorization`.
- `Project.toml`: `PureUMFPACK` added to `[weakdeps]`, `[compat]`, `[extras]`, the `test` target, and the new extension `LinearSolvePureUMFPACKExt = ["PureUMFPACK", "SparseArrays"]`.
- `ext/LinearSolvePureUMFPACKExt.jl`: `init_cacheval` (returns an empty typed `PureLU` for sparse inputs so the cache field type can hold the real factor) and `SciMLBase.solve!` modeled on the `UMFPACKFactorization`/`PureKLUFactorization` methods.

### Caveat on caching semantics

PureUMFPACK exposes `splu` (full factorization) and `solve`/`\`; it has **no** in-place numeric-refactorization (`lu!`-style) API and no symbolic-reuse API. So each *fresh* solve (`cache.isfresh`) refactorizes via `splu`. `reuse_symbolic`/`check_pattern` are accepted for API parity with `UMFPACKFactorization` but cannot skip the symbolic step here; this is documented in the docstring and code comments. Cached factorizations are still reused across solves that only change the right-hand side. Singular factors (a zero on `U`'s diagonal under `check = false`) are mapped to `ReturnCode.Infeasible`.

### Registration

PureUMFPACK.jl is **not yet registered** in the General registry, so `test/pureumfpack.jl` adds it via `Pkg.add(url = "https://github.com/SciML/PureUMFPACK.jl.git")` before running.

## Tests

`test/pureumfpack.jl` (wired into the `Core`/`All` group of `runtests.jl`) verifies the extension loads (`Base.get_extension(...) !== nothing`), a square sparse solve, caching + value-only refactor, `reuse_symbolic = false`, and singular -> `Infeasible`. Run locally on Julia 1.10, all passing.

---


ChrisRackauckas:
This should also get an opt-out https://github.com/SciML/LinearSolve.jl/pull/1023

ChrisRackauckas:
No opt out since it uses blas

ChrisRackauckas (approved):

PR #1005, opened 2026-06-05
Add SpecializingFactorizations.jl package extension

ChrisRackauckas-Claude:

Adds an opt-in package extension wrapping [SpecializingFactorizations.jl](https://github.com/SciML/SpecializingFactorizations.jl) as two LinearSolve algorithms. This is a weakdep extension only — no hard dependency and no change to any default algorithm.

## What SpecializingFactorizations.jl provides

It exposes two type-stable, single-workspace dense factorizations that detect the property a matrix actually has and dispatch to the specialized solve, all behind one concrete workspace type (the structure/rank choice is a runtime enum field, not a Julia type, so the pipeline infers and the warm path is allocation-free):

- `specializinglu` / `SpecializedLU` — a **square** solver that cheaply scans `A` for structure (diagonal, bidiagonal, tridiagonal, banded, triangular, symmetric PD, symmetric/Hermitian indefinite) and uses the matching specialized factorization instead of a general `O(n^3)` LU.
- `specializingqr` / `SpecializedQR` — a **rectangular / rank-deficient** least-squares solver. Column-pivoted rank-revealing QR (`geqp3`) that returns the least-squares / minimum-norm solution for any shape including singular/rank-deficient, never throwing.

## Wrapped algorithms

- `SpecializedLUFactorization` — wraps `specializinglu`; for square systems.
- `SpecializedQRFactorization` — wraps `specializingqr`; for rectangular / rank-deficient least-squares.

Both wrapped because they cover complementary problem classes (square structured solves vs. rectangular/rank-deficient least-squares) and both expose the `ldiv!` + in-place re-factor (`specializinglu!` / `specializingqr!`) interface the `LinearCache` warm-solve path wants.

## Implementation

- `src/extension_algs.jl`: two `AbstractDenseFactorization` structs with docstrings + examples; exported from `src/LinearSolve.jl`.
- `ext/LinearSolveSpecializingFactorizationsExt.jl`: `init_cacheval` builds the factorization; `SciMLBase.solve!` re-factors in place on `cache.isfresh` (reusing the workspace via `specializing*!`), guards `issuccess`, and returns `build_linear_solution` with the appropriate `ReturnCode`.
- `Project.toml`: `SpecializingFactorizations` added to `[weakdeps]`, `[extensions]`, `[compat]` (`"0.1"`), `[extras]`, and the `test` target. No `SparseArrays` trigger — it operates on dense matrices only.
- `test/specializing_factorizations.jl`: focused test set (extension load, square/structured/complex/float32 LU, overdetermined and rank-deficient QR, cache reuse across re-factor for both), wired into the Core test group.

## Registration

SpecializingFactorizations.jl is registered in the General registry (v0.1.0).

## Caveats

- `SpecializedLUFactorization` is square-only by construction; rectangular systems should use `SpecializedQRFactorization`.
- Structure detection is structural (`!iszero`), not numerical/tolerant.
- Cache reuse re-factors into the existing workspace; a different matrix *size* on a warm cache regrows buffers (handled by `specializing*!`).

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -63,6 +63,7 @@ STRUMPACK_jll = "86fbd0b9-476f-557c-b766-62c724b42d8c"
 SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
 SparseMatricesCSR = "a0a7dd2c-ebf4-11e9-1f05-cf50bc540ca1"
 Sparspak = "e56a9233-b9d6-4f03-8d0f-1825330902ac"
+SpecializingFactorizations = "fa08b7a1-13d3-4faf-875d-5cbc1520e3f3"
 blis_jll = "6136c539-28a5-5bf0-87cc-b183200dce32"
 
 [extensions]
@@ -98,6 +99,7 @@ LinearSolveRecursiveFactorizationExt = "RecursiveFactorization"
 LinearSolveSTRUMPACKExt = ["SparseArrays", "STRUMPACK_jll"]
 LinearSolveSparseArraysExt = "SparseArrays"
 LinearSolveSparspakExt = ["SparseArrays", "Sparspak"]
+LinearSolveSpecializingFactorizationsExt = "SpecializingFactorizations"
 
 [compat]
 AMD = "0.5"
@@ -163,6 +165,7 @@ SparseArrays = "1.10"
 SparseColumnPivotedQR = "2"
 SparseMatricesCSR = "0.6"
 Sparspak = "0.3.9"
+SpecializingFactorizations = "0.1"
 StaticArrays = "1.9"
 StaticArraysCore = "1.4.3"
 Test = "1.10"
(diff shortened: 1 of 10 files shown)


ChrisRackauckas:
This should also get an opt-out https://github.com/SciML/LinearSolve.jl/pull/1023

ChrisRackauckas-Claude:
Rebased onto current main (was conflicting after the test-suite restructure in #1022 and the test-target slimming).

Conflict resolution:
- `Project.toml`: kept main's slimmed compat/extras/test target and added `SpecializingFactorizations` to each (it stays a base test dep since the test runs in the Core group).
- `test/runtests.jl`: took main's new grouped structure and re-added the test under the Core group as `core/specializing_factorizations.jl` (file moved from `test/specializing_factorizations.jl` to match the new `test/<group>/` layout).

Verified locally on Julia 1.12: the `LinearSolveSpecializingFactorizationsExt` extension precompiles and all 18 tests in `test/core/specializing_factorizations.jl` pass. Runic reports no formatting changes on the touched files.

ChrisRackauckas-Claude:
Rebased on main and added the ForwardDiff opt-out per the review comment (same pattern as #1023):

- `_use_direct_dual_solve` in `ext/LinearSolveForwardDiffExt.jl` now returns `true` for `SpecializedLUFactorization` and `SpecializedQRFactorization`, so Dual-number problems are solved directly instead of through the primal/partials splitting path. This is valid because SpecializingFactorizations.jl has generic-element fallbacks for non-`BlasFloat` types (generic dense LU, generic column-pivoted QR). It also matters for correctness on the QR side: the split-path math (`∂x = A⁻¹(∂b − ∂A·x)`) is square-only, while the direct path handles Dual rectangular least-squares.
- Added tests in `test/core/forwarddiff_overloads.jl`: square Dual systems through both algorithms plus a rectangular Dual least-squares solve through `SpecializedQRFactorization`, all checked against `A \\ b` / `qr(A) \\ b`. Note the tests compare values and partials separately: when the primal diff is exactly zero but partials differ by an ulp, `isapprox` on Dual vectors returns `false` because `norm(diff)` evaluates to `Dual(NaN,…)` (sqrt has infinite slope at 0), so the plain `≈(sol, ref)` pattern only passes for bitwise-identical results.

Ran locally on the rebased branch: `test/core/forwarddiff_overloads.jl` (full file) and `test/core/specializing_factorizations.jl` both pass; Runic reports no formatting changes.

Note for whichever of this and #1023 merges second: both touch the same line of `_use_direct_dual_solve`, so there will be a trivial conflict.

ChrisRackauckas-Claude:
CI status on `3bd3d2a8`: everything this PR touches is green — **Core (julia 1 / lts / pre), QA, Spell Check, Runic, format-check, Sublibrary CI** all pass (the earlier Core failure was resolve.jl's subtype sweep hitting the new algs without the extension loaded; fixed by loading SpecializingFactorizations there, which also gives both algorithms re-solve coverage).

Remaining red is not from this PR:
- tests/LinearSolveHYPRE — pre-existing on main, fix in #1030
- tests/LinearSolvePETSc — pre-existing on main (PETSc.jl 0.4.10 API change), fix in #1031
- Documentation — pre-existing NonstructuralZeros @ref fatal error, fix in #1026
- Downgrade / Downgrade Sublibraries — pre-existing, fixes in #1025 / #1029
- tests/NoPre (julia lts) — new upstream Enzyme/LLVM.jl breakage (`MethodError: LLVM.ConstantInt(::LLVM.VectorType, ::Int64)`), reproduces on main run 27251663412; being investigated separately
- tests/NoPre (julia 1), Pardiso (julia 1), ParU (julia 1), GPU — self-hosted runner infra deaths (job ran 0 steps, no logs, died at ~10 min); need a rerun from someone with admin rights, or the next push will retry them

PR #1016, opened 2026-06-07
Add PersistentDropFactorization for sparse solves with wobbling stored zeros

ChrisRackauckas-Claude:
> [!NOTE]
> Draft — please **ignore until reviewed by @ChrisRackauckas**. Opened by an agent for review.

## What

Adds `PersistentDropFactorization`, an opt-in sparse meta-algorithm for sequences of solves where the matrix keeps a **fixed stored sparsity pattern** but carries many numerically-zero stored entries whose positions *wobble* between solves (ODE/DAE Jacobians, `W = I - γJ` built from a conservative symbolic sparsity pattern).

## Why

Stored zeros are not free: they join the fill-reducing ordering and symbolic factorization as if they were real nonzeros, inflating the factor (~2× for KLU-style orderings), so every numeric refactor and triangular solve is more expensive. The obvious fix — `dropzeros` before each solve — is a trap: the dropped pattern changes whenever a zero wobbles, forcing a fresh symbolic analysis nearly every step, which costs far more than it saves.

`PersistentDropFactorization` instead drops only the entries that have been zero in **every** solve so far — the complement of the running **union** of ever-nonzero positions. That union only grows, so the reduced pattern is always a valid superset of the current nonzeros: the inner symbolic analysis stays valid and is reused, widening (re-analyzing) only the handful of times a previously-dead entry first activates. Steady state is one refactor + one solve per step on the small pattern.

## How it composes with the default LU → QR fallback

The reduction is applied **above** the inner solver: the reduced matrix becomes the operand handed to the inner algorithm. `alg = nothing` (default) wraps the default sparse polyalgorithm via a nested `LinearCache` on the reduced matrix. Because the inner cache solves the reduced matrix, the default's **singular LU → column-pivoted-QR fallback factorizes the same reduced matrix** — the fill reduction carries through to the QR fallback automatically, with no solver-specific threading. (Verified in tests: on singular steps the inner reports `fell_back_to_qr` and `nnz(inner.A) == length(keep) < nnz(A)`.)

Works with `alg = nothing` (default), `KLUFactorization()`, `PureKLUFactorization()`, `UMFPACKFactorization()`, etc.

## Results

On a 198×198 mechanism-DAE Jacobian sequence (30% stored zeros, dropzeros pattern changing nearly every step, ~2× fill reduction):

| strategy (steady state) | µs/step |
|---|--:|
| default polyalgorithm (no drop) | 48.7 |
| `PersistentDropFactorization()` | **26.0** (1.9×) |
| `PersistentDropFactorization(PureKLUFactorization())` | **25.1** |

Per-step `dropzeros` (naive) on the same data is ~3× *slower* than no-drop (constant re-analysis); `PersistentDropFactorization` avoids that by keying on the stable union pattern.

## Tests

`test/persistent_drop.jl` (79 assertions): correctness vs dense reference for default/PureKLU/UMFPACK inner algorithms, union convergence (bounded re-analyses, none after warmup), the reduced operand reaching the singular QR fallback, and enforcement of the constant-stored-pattern contract. Existing `default_algs.jl` and `qa.jl` (Aqua/ExplicitImports) pass; Runic-formatted.


ChrisRackauckas-Claude:
Superseded by the assumptions-driven design (#new). Instead of a wrapper type — which a runtime heuristic can't select without breaking type stability — the persistent-nonstructural-zero reduction is now an `OperatorAssumptions(...; persistent_nonstructural_zeros=...)` knob backed by shared type-stable helpers, wired into the default solver so one reduced matrix is shared across the KLU/UMFPACK + QR-fallback slots. Closing in favor of that.

PR #1017, opened 2026-06-07
persistent_nonstructural_zeros operator assumption + default sparse persistent-zero reduction

ChrisRackauckas-Claude:
> [!NOTE]
> Draft — please **ignore until reviewed by @ChrisRackauckas**. Supersedes #1016 (wrapper-type approach, closed).

## What

Adds an `OperatorAssumptions(...; persistent_nonstructural_zeros=...)` knob and a shared, type-stable reduction that drops *persistently* numerically-zero stored entries from sparse matrices before factorization, wired into the default solver.

## Why

ODE/DAE Jacobians and `W = I - γJ` operators built from a conservative symbolic sparsity pattern store ~30% entries that are numerically zero at positions that stay zero across a solve sequence. Those stored zeros join the fill-reducing ordering + symbolic factorization as if real, inflating the factor ~2× (for KLU-style orderings), so every refactor and triangular solve is more expensive. Per-step `dropzeros` is a trap (the dropped pattern wobbles, forcing constant re-analysis).

The reduction keeps only the **running union of ever-nonzero positions** — a pattern that only grows, so it's always a valid superset of the current nonzeros. The inner symbolic factorization stays valid and is reused; it widens (re-analyzes) only the handful of times a dead entry first activates, then refactors forever.

## Design (addresses the type-stability + sharing constraints)

- **`OperatorAssumptions` gains `persistent_nonstructural_zeros::Union{Nothing,Bool}`** — a runtime field (like `condition`), *no new type parameter*. `nothing` = auto-detect from the starting matrix's stored-zero fraction (`PERSISTENT_ZERO_FRACTION_THRESHOLD`, 0.1); `true`/`false` = force. **Force-off is bit-identical to today with no detection overhead.**
- **Shared helpers** `init_sparse_reduction` / `reduce_operand!` (generic no-op fallbacks in core; `SparseMatrixCSC` methods in the SparseArrays ext). `init_sparse_reduction` returns a **concrete** `SparseReduction` whose `active` is a *runtime field* — so the cache type never depends on whether the reduction fires, and the **algorithm type never switches**. This is the key fix vs the closed wrapper-type approach (#1016), which couldn't be heuristic-selected without losing inference.
- **Default solver shares one reduced matrix.** `DefaultLinearSolverInit` gains a `TR` param + `sparse_reduction` field. The sparse `solve!` builds the reduced operand once and `setfield!`-swaps `cache.A` to it (no `isfresh`/`A_backup` side effects) around the *whole* sub-dispatch — so the KLU/UMFPACK attempt **and** the singular LU → column-pivoted-QR fallback factorize the same reduced matrix. Placed so the upcoming default-cache-sharing PR can own it.

## Results / verification (all run locally, Julia 1.10)

- **Type-stable:** default `solve!` return type stays concrete (the `default_algs.jl` `@inferred` check passes); `init` returns the *same concrete cache type* for auto/on/off (the reduction adds no instability; `reduce_operand!` is `@inferred`).
- **~1.9× steady-state** vs the un-reduced default on a 198×198 mechanism-DAE Jacobian sequence (30% stored zeros): 25.4 vs 49.1 µs/step. Single solves with stored zeros also win (fill reduction pays off the first time); tight matrices (no stored zeros) auto-stay-inactive (~0 overhead).
- **`test/persistent_zeros.jl` (47 assertions):** correctness vs dense reference; auto/force-on/force-off; bounded widenings + zero further re-analysis once converged; the reduced operand reaching the singular QR fallback; constant-pattern contract enforced.
- `default_algs.jl` and `qa.jl` (Aqua + ExplicitImports) pass; Runic-formatted.

## Follow-ups (kept in mind, not in this PR)

- **Standalone sparse-solver wiring** (KLU/UMFPACK/Sparspak/SPQR used directly, not via the default) reuses the same helpers; it lands with the planned **default-cache-sharing** PR, since both want the reduction state held uniformly across solver cachevals.
- The cleaner upstream fix where controllable (MTK / sparse AD) is to tighten the symbolic sparsity pattern so the stored zeros never appear; this assumption is the robust runtime safety net.

### DIFF
--- a/ext/LinearSolveHYPREExt.jl
+++ b/ext/LinearSolveHYPREExt.jl
@@ -198,10 +198,10 @@ function SciMLBase.init(
     cache = LinearCache{
         typeof(A), typeof(b), typeof(u0), typeof(p), typeof(alg), Tc,
         typeof(Pl), typeof(Pr), typeof(reltol), typeof(verb_spec),
-        typeof(__issquare(assumptions)), typeof(sensealg),
+        typeof(__issquare(assumptions)), typeof(sensealg), Nothing,
     }(
         A, b, u0, p, alg, cacheval, isfresh, precsisfresh, Pl, Pr, abstol, reltol,
-        maxiters, verb_spec, assumptions, sensealg
+        maxiters, verb_spec, assumptions, sensealg, nothing
     )
     return cache
 end
--- a/ext/LinearSolveSparseArraysExt.jl
+++ b/ext/LinearSolveSparseArraysExt.jl
@@ -988,4 +988,97 @@ LinearSolve.PrecompileTools.@compile_workload begin
     end
 end
 
+# --- Persistent-nonstructural-zero reduction (shared helper for sparse solvers) ---
+#
+# Drops stored entries that have been numerically zero in every solve so far: the
+# kept set is the running union of ever-nonzero positions, which only grows, so
+# `reduced` is always a valid superset of the current matrix's nonzeros and the
+# inner symbolic factorization stays reusable. `active` is a runtime field (not a
+# type), so `init_sparse_reduction` returns a concrete type regardless of whether
+# the reduction fires — keeping the caller type-stable.
+mutable struct SparseReduction{Tv, Ti}
+    active::Bool
+    colptr::Vector{Ti}            # full stored pattern (reference, for validation)
+    rowval::Vector{Ti}
+    mask::Vector{Bool}            # over full nnz positions: kept (ever-nonzero)?
+    keep::Vector{Int}             # kept positions into the full nzval (CSC order)
+    reduced::SparseMatrixCSC{Tv, Ti}
+    nanalyze::Int
+    nrefactor::Int
+end
+
+function _persistent_reduced(
+        colptr::Vector{Ti}, rowval, n, mask, nzval::AbstractVector{Tv}
+    ) where {Ti, Tv}
+    keep = Int[]
+    rcolptr = Vector{Ti}(undef, n + 1)
+    rrowval = Ti[]
+    rnzval = Tv[]
+    @inbounds for j in 1:n
+        rcolptr[j] = length(rrowval) + 1
+        for p in colptr[j]:(colptr[j + 1] - 1)
+            if mask[p]
+                push!(keep, p)
+                push!(rrowval, rowval[p])
+                push!(rnzval, nzval[p])
+            end
+        end
+    end
+    rcolptr[n + 1] = length(rrowval) + 1
+    return keep, SparseMatrixCSC(n, n, rcolptr, rrowval, rnzval)
+end
+
+function LinearSolve.init_sparse_reduction(
+        A::AbstractSparseMatrixCSC{Tv, Ti}, assumptions
+    ) where {Tv, Ti}
+    pnz = LinearSolve.__persistent_nonstructural_zeros(assumptions)
+    nz = nonzeros(A)
+    active = if pnz === true
+        true
+    elseif pnz === false
+        false
+    else
+        !isempty(nz) &&
+            count(iszero, nz) / length(nz) >= LinearSolve.PERSISTENT_ZERO_FRACTION_THRESHOLD
+    end
+    n = size(A, 1)
+    colptr = copy(getcolptr(A))
+    rowval = copy(rowvals(A))
+    if active
+        mask = Bool[!iszero(v) for v in nz]
+        keep, reduced = _persistent_reduced(colptr, rowval, n, mask, nz)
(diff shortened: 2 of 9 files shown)


ChrisRackauckas-Claude:
Addressed review (commit `ef77730`):

- **No `try`/`finally`** in the default sparse `solve!`: `cache.A` is restored by normal control flow after the sub-solve. The sparse solvers report status by return code (not exceptions), so the swap doesn't need exception safety.
- **Documented `condition`** (and `issquare`) in the `OperatorAssumptions` docstring.
- **Inconsistent (non-persistent) zeros → auto `dropzeros` without caching.** When the union of ever-nonzero positions bloats past `UNION_BLOAT_FRACTION` (0.9) in auto mode, the zeros aren't persistent, so the reduction stops maintaining the union and drops each matrix's own zeros per solve — the inner solver re-analyzes on pattern change. Explicit `persistent_nonstructural_zeros = true` pins union caching and never switches. New test covers the switch; both reduction paths stay `@inferred`-clean and `default_algs` type-stability check still passes. 61 assertions green.

ChrisRackauckas-Claude:
Update (commit `c33fd2f`):

**Expanded to standalone sparse solvers.** `LinearCache` now has a `sparse_reduction` field, set at `init` for standalone `AbstractSparseFactorization` algorithms (`nothing` for the default solver — which keeps its own in `DefaultLinearSolverInit` — and for dense/iterative). `KLUFactorization`, `PureKLUFactorization`, and `UMFPACKFactorization` `solve!` now drop persistent nonstructural zeros via the shared `reduce_operand!` (a no-op when the field is `nothing`, so the default path is unchanged). Verified: standalone PureKLU/KLU/UMFPACK with the reduction match the dense reference on the 200-step Jacobian sequence; `reinit!` (which mutates fields) preserves the reduction state; dense/iterative paths and the default are unaffected.

**Clearer bloat threshold.** The auto union→per-solve switch now triggers on the **fraction of the *starting* zeros that have become nonzero** (`NONPERSISTENT_ZERO_FRACTION = 0.5`), independent of matrix density — rather than a fraction of the whole stored pattern (which drifted with density). I.e. it answers "how many of the originally-zero entries turned nonzero," not "what fraction of all stored entries."

**Tested on a 200-matrix ODE Jacobian set:** the zeros are persistent there (only ~2% of starting zeros ever activate), so the union stays small (686/977) and the reduction stays in union-caching mode.

95 assertions green (incl. standalone-solver cases); `default_algs` type-stability check, Aqua, ExplicitImports all pass. A separate timing run comparing this HEAD vs the original branch commit on the persistent case (to confirm no regression) is in progress.

ChrisRackauckas-Claude:
Update (commit `4bd6e66`) — re: the no-zero-dropping default sparse-QR branch:

The default's non-LU sparse branch (sparse QR for non-square, and CHOLMOD) did no zero dropping. Now:
- **Sparse column-pivoted QR** (non-square / least-squares) drops persistent nonstructural zeros via the same operand-swap as the sparse-LU branch. `_persistent_reduced` and the reduction state were generalized to carry separate row/column extents, so the reduced operand keeps the rectangular `m×k` shape (it previously assumed a square `n×n` result). Standalone `SparseColumnPivotedQRFactorization` reduces too.
- **CHOLMOD is intentionally left alone** — dropping numerical zeros asymmetrically would break the symmetric Cholesky structure, and its input is a `Symmetric` wrapper so the reduction is inactive there regardless.

Verified on a 14×9 sparse least-squares sequence (reduced keeps shape, drops the persistent zeros, matches the dense least-squares solution); square LU, CHOLMOD, `default_algs`, and `nonsquare.jl` all unchanged. 104 assertions green.

ChrisRackauckas-Claude:
Update (commits `7119404`,`+qa fix`) — renamed + enum, per your suggestion:

`persistent_nonstructural_zeros::Union{Nothing,Bool}` → **`nonstructural_zeros::NonstructuralZeros.T`**, an exported EnumX enum that declares how the operator's nonstructural (stored-but-numerically-zero) entries behave:

| value | meaning | reduction |
|---|---|---|
| `Auto` (default) | detect from the starting matrix, adapt | cached union → per-solve if non-persistent |
| `None` | none worth dropping | off (bit-identical) |
| `Persistent` | present, stable positions | cached-union |
| `Present` | present, positions may vary | per-solve dropzeros |

`Present` makes the previously auto-only per-solve mode explicitly selectable (the open question from before). `Auto` still detects and, when the zeros turn out non-persistent, switches to per-solve on its own. Test file renamed to `test/nonstructural_zeros.jl`; 114 assertions (incl. a `Present` case) green; default_algs, Aqua, ExplicitImports pass.

ChrisRackauckas (approved):

PR #1023, opened 2026-06-09
Opt PureKLU and RFLU out of split dual AD path

ChrisRackauckas:

PR #1025, opened 2026-06-10
Add LogExpFunctions lower-bound compat to fix Downgrade CI

ChrisRackauckas-Claude:
**Note: this PR should be ignored until reviewed by @ChrisRackauckas.**

## Problem

The **Downgrade / Downgrade Tests - Core** job has failed on every `main` run since 2026-06-07 (and on all PRs since, e.g. #1023). Root cause, bisected by registry timeline rather than git history since no LinearSolve commit triggered it:

- ForwardDiff **1.4.0** was registered 2026-06-07 18:03 UTC — exactly between the last green Downgrade run (16:00–17:28) and the first red one (23:44).
- ForwardDiff 1.4.0 is the first version allowing **LogExpFunctions 1.x**. The downgrade resolver pins *direct* deps to their lower bounds but *maximizes* indirect deps, so it now pins LogExpFunctions **1.0.1** into the downgraded manifest (reached via the strong dep PureKLU → ForwardDiff → LogExpFunctions).
- The test sandbox then cannot add the Zygote test dep with `allow_reresolve = false`: every Zygote 0.7.x requires LogExpFunctions 0.3.x →
  ```
  ERROR: Unsatisfiable requirements detected for package Zygote [e88e6eb3]:
  ...restricted by compatibility requirements with LogExpFunctions [2ab3a3ac] to versions: 0.1.0-0.6.40 or uninstalled — no versions left
  ```

## Fix

Add LogExpFunctions to `[extras]` with compat `"0.3"`. This anchors the downgraded manifest at LogExpFunctions 0.3.x so the Zygote test dep resolves. The normal (non-downgrade) test environment is already held at 0.3.x by Zygote's own compat, so nothing changes there. The bound should be dropped once Zygote supports LogExpFunctions 1.x.

## Verification (all run locally on Julia 1.10.11, same procedure as the CI job: `julia-downgrade-compat` with stdlib skip + `mode=deps`, then `Pkg.test(allow_reresolve=false)` with `GROUP=Core`)

- **Unmodified `main`**: reproduces the exact CI failure (`Unsatisfiable requirements ... Zygote ... LogExpFunctions`).
- **With this change, downgraded**: manifest pins `LogExpFunctions 0.3.29`, direct deps stay at their lower bounds, the test sandbox resolves, and the Core test suite launches (`Testing Running tests...`).
- **With this change, normal mode** (no downgrade): resolution unaffected, test suite launches.

Caveat on local verification: the sandbox machine reaps long-running Julia processes (external SIGTERM ~10 min in), so the full downgraded Core suite could not be run to completion locally — resolution success and test startup were verified, which is precisely the failure mode of the CI job. The Downgrade CI run on this PR provides the full-suite confirmation.


ChrisRackauckas-Claude:
Independent verification (separate agent, full local run before noticing this PR existed):

Reproduced the failure on unmodified `main` with the exact CI flow on Julia 1.10.11: ran julia-downgrade-compat's `downgrade.jl` (`--min=@deps --julia=1.10`, stdlibs skipped) and then `Pkg.test(; julia_args=["--check-bounds=yes"], force_latest_compatible_version=false, allow_reresolve=false)` — got the identical `Unsatisfiable requirements detected for package Zygote` / `LogExpFunctions restricted to 1.0.1` error. The downgraded manifest indeed lands `LogExpFunctions 1.0.1` via maximized ForwardDiff 1.4.0/DiffRules.

With `LogExpFunctions` in `[extras]` + compat capped to 0.3.x (I tested `0.3.29`, this PR's `0.3` resolves identically since the resolver maximizes non-minimized packages to 0.3.29), the downgrade resolver produces `LogExpFunctions v0.3.29`, the test sandbox resolves `Zygote v0.7.10`, and the **full GROUP=Core test suite passes** on the downgraded manifest with Julia 1.10.11:

```
     Testing Running tests...
...
Test Summary: | Pass  Total  Time
Resize        |   70     70  2.3s
```

Also confirmed no Zygote release supports LogExpFunctions 1.x (Zygote master still has `LogExpFunctions = "0.3.29"`), so a Zygote-side bound bump is not currently possible.

PR #1029, opened 2026-06-10
Fix Downgrade Sublibraries CI: numeric julia-version and resolvable sublibrary lower bounds

ChrisRackauckas-Claude:

## Problem

`Downgrade Sublibraries / test (lib/LinearSolveAutotune)` and `test (lib/LinearSolvePyAMG)` fail on every run of `main` (the workflow has never passed). The visible CI error is:

```
[ERROR] Invalid compat version spec: --julia=lts
```

julia-downgrade-compat's Resolver-based `resolve.jl` only accepts registry compat syntax for `julia_version`, not setup-julia channel aliases. Fixing that locally exposed four more layers of breakage, all reproduced and fixed here. **Full downgraded test suites of both sublibraries now pass locally on Julia 1.10.11** (exact CI flow: `downgrade.jl` with `--min=@alldeps --julia=1.10` and stdlibs+sublib names skipped, then `Pkg.build` and `Pkg.test(; julia_args=["--check-bounds=yes"], force_latest_compatible_version=false, allow_reresolve=false)` with `LINEARSOLVE_TEST_GROUP=Core`).

## Fixes

1. **Workflow**: `julia-version: "lts"` → `"1.10"` (numeric works for both setup-julia and the downgrade resolver). The reusable workflow's *default* is also `lts`, reported upstream as SciML/.github#83 (OrdinaryDiffEq.jl already passes a numeric `"1.11"`).

2. **`Metal = "1"` → `"1.5"` (Autotune)**: Metal 0.4–1.4 depend on Python_jll, and Python_jll 3.8.x carries a registry compat entry for `LibMPDec_jll` on versions where it is not a dep — Resolver.jl crashes with `KeyError: key "LibMPDec_jll" not found` (`u = deps_uuids[name]` in `bin/Registries.jl`; Pkg's own resolver tolerates this registry pattern). Metal ≥ 1.5 drops the Python_jll dep, removing it from the closure.

3. **Floors consistent with the path-sourced parent**: the downgrade action removes the `[sources]`-path `LinearSolve` before resolving, then the `Pkg.test` sandbox grafts the real parent back in with `allow_reresolve=false` — so frozen minimized versions must satisfy the parent's compat. Aligned: `FastLapackInterface ≥ 2.0.4`, `RecursiveFactorization ≥ 0.2.26` (parent floors); `Plots ≥ 1.40` (old Plots freezes GeometryBasics 0.3 → StructArrays 0.6.21, unsatisfiable against the parent's `RecursiveArrayTools ≥ 3.37`); `PrettyTables ≥ 2.4` (needed by `SymbolicIndexingInterface ≥ 0.3.42`, required by RecursiveArrayTools 3.37+); `SciMLBase ≥ 2.148` in **LinearSolvePyAMG** (parent floor — its downgraded manifest froze SciMLBase 2.0.x otherwise).

4. **`LAPACK_jll = "3"` → `"3.12"` (Autotune)**: minimized builds are genuinely broken — 3.9.0+0 cannot load (`libblastrampoline.so: cannot open shared object file`, unversioned LBT link) and 3.10.0+0 **segfaults** in OpenBLAS `idamax_k_ZEN` during the autotune benchmarks (LinearSolveBLISExt ccalls `@blasfunc(dgetrf_)` into it). 3.12.0+0 is the lowest build verified to pass the full suite.

5. **`ProgressMeter = "1"` → `"1.6"` (Autotune)**: 1.0.0 lacks `Progress(n; desc, barlen, showspeed)` used by `src/benchmarking.jl` — exactly the 3 test errors observed (`MethodError: no method matching Progress(::Int64; desc::String, ...)`). 1.6.0 is the lowest version supporting the full keyword API used (verified 1.4.0/1.5.0 fail, 1.6.0 works).

Patch bumps: LinearSolveAutotune 1.11.0 → 1.11.1, LinearSolvePyAMG 1.1.0 → 1.1.1.

## Local verification output

LinearSolveAutotune (downgraded manifest, Julia 1.10.11):
```
Test Summary:             | Pass  Total     Time
LinearSolveAutotune Tests |  151    151  4m30.9s
     Testing LinearSolveAutotune tests passed
```

LinearSolvePyAMG (downgraded manifest, Julia 1.10.11):
```
Test Summary:          | Pass  Total   Time
LinearSolvePyAMG Tests |   19     19  11.2s
     Testing LinearSolvePyAMG tests passed
```

## Upstream issues found (not filed — outside SciML)

- **julia-actions/julia-downgrade-compat**: re-adds `[sources]` path packages to the manifest without a `deps` list and excludes them (and their dep closure) from resolution; the outer environment cannot precompile (`Package LinearSolve does not have PrecompileTools in its dependencies`) — only the `Pkg.test` sandbox re-resolve makes it work.
- **StefanKarpinski/Resolver.jl**: `KeyError` on registry compat entries naming packages absent from `deps` for that version range (Python_jll 3.8.x / LibMPDec_jll).


.github/workflows/DowngradeSublibraries.yml
```diff
@@ -21,7 +21,10 @@ jobs:
     uses: "SciML/.github/.github/workflows/sublibrary-downgrade.yml@v1"
     secrets: "inherit"
     with:
-      julia-version: "lts"
+      # julia-downgrade-compat's resolver rejects channel aliases like "lts"
+      # ("Invalid compat version spec: --julia=lts"); pass the numeric
+      # major.minor instead. Bump when the LTS advances.
+      julia-version: "1.10"
```
ChrisRackauckas:
```suggestion
      julia-version: "lts"
```

PR #1036, opened 2026-06-12
Fix Sparspak symbolic reuse invalidated by per-solve dropzeros

ChrisRackauckas-Claude:
## Summary

The `nonstructural_zeros` feature introduced in v3.85 (per-solve `dropzeros` mode, `cache_union=false`) can return sparse matrices with different `nnz` counts across successive Newton iterations. Sparspak's `sparspaklu!` reuses the symbolic factorization from the previous call without checking whether the sparsity structure has changed — unlike KLU/UMFPACK, which both guard against this. When `nnz` changes, `sparspaklu!` produces a corrupt factorization, leading to wrong linear-solve results and stalled nonlinear (e.g. BVP) iterations.

- **`ext/LinearSolveSparseArraysExt.jl`**: Add `reduced_nnz::Int` and `structure_changed::Bool` to `SparseReduction`. Update `reduce_operand!` to track whether the per-solve dropzeros changed the matrix structure.
- **`ext/LinearSolveSparspakExt.jl`**: Before calling `sparspaklu!`, check `structure_changed` via the `DefaultLinearSolverInit` cache. If the structure is unstable, skip symbolic reuse and fall back to a full `sparspaklu`.

Fixes the BigFloat BVP stalling regression reported in SciML/BoundaryValueDiffEq.jl#509 (comment from @ChrisRackauckas identifying FIRK test failures).

## Test plan

- [x] `test/core/nonstructural_zeros.jl` — 122/122 pass
- [x] `BigFloat` + `SparspakFactorization` direct solve — correct results
- [x] `BigFloat` + `DefaultLinearSolver` (Sparspak path) — correct results
- [ ] CI on full test suite

> **Note**: Please ignore until reviewed by @ChrisRackauckas.

### DIFF
--- a/ext/LinearSolveSparseArraysExt.jl
+++ b/ext/LinearSolveSparseArraysExt.jl
@@ -13,7 +13,7 @@ using ArrayInterface: ArrayInterface
 using LinearAlgebra: LinearAlgebra, I, Hermitian, Symmetric, cholesky, ldiv!, lu, lu!
 using SparseArrays: SparseArrays, AbstractSparseArray, AbstractSparseMatrixCSC,
     SparseMatrixCSC,
-    nonzeros, rowvals, getcolptr, sparse, sprand, dropzeros
+    nonzeros, rowvals, getcolptr, sparse, sprand, dropzeros, nnz
 using SciMLLogging: @SciMLMessage
 
 @static if Base.USE_GPL_LIBS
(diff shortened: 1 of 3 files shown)


ChrisRackauckas-Claude:
Root-cause companion: #1037 makes PureKLU the default sparse LU for generic (non-BLAS) element types, so the default path for BigFloat no longer selects Sparspak at all. This PR remains valuable as defense-in-depth for users who explicitly request `SparspakFactorization()`. The resolve.jl PureUMFPACKFactorization test fix is shared between both PRs (identical patch, resolves cleanly when either merges).

PR #1048, opened 2026-06-17
Specialize split dual AD path on sparse matrix in ForwardDiff extension

hersle:
## Checklist

## Additional context

In my package I need to forward-differentiate stiff ODEs with sparse Jacobians. Before this PR, the [generic dual AD path through linear solves](https://github.com/SciML/LinearSolve.jl/blob/53a1799c96bde2607f793cfae839629cb77d6016/ext/LinearSolveForwardDiffExt.jl#L639) in the ODE has been a big bottleneck, as it does not specialize properly on the sparse matrix, but iterates through all dense $m \times n$ entries (including zeros) and goes through slow binary searches to look up the sparse matrix for each $(i, j)$.

With this PR, I try to specialize the expensive `partials_to_list` and `update_partials_list` functions for sparse matrices in order to remove this bottleneck.


hersle:
Example timings from my package:
```julia
using SymBoltz, ForwardDiff
M = ΛCDM()
p= parameters_Planck18(M)
prob = CosmologyProblem(M, p; jac = true, sparse = true) # stiff ODE with 82 unknowns and sparse+analytical Jacobian

vary = [M.c.Ω₀, M.b.Ω₀]
probgen = parameter_updater(prob, vary)
k = logrange(1e0, 1e3; length = 100)
P(θ) = spectrum_matter(probgen(θ), k; thread = false) # depends on solution of 100x of the above ODEs
θ₀ = [p[par] for par in vary]
@time ∂P_∂θ = ForwardDiff.jacobian(P, θ₀)
@profview ∂P_∂θ = ForwardDiff.jacobian(P, θ₀)
```
```
Before this PR: 18.477147 seconds (379.61 k allocations: 113.941 MiB, 0.60% gc time)
After this PR:   2.796430 seconds (383.78 k allocations: 102.552 MiB, 0.48% gc time)
```
Before this PR (note the big `update_partials_list!`):
bilde


After this PR (note the small `update_partials_list!`):
bilde

hersle:
Will fix formating.

Is it OK to move SparseArrays to deps? That is the cleanest solution I could think of. Or do you see another way?

ChrisRackauckas:
Which linear solver? It shouldn't make a difference on PureKLU, and that would likely be pretty fast in a lot cases like this.

ChrisRackauckas:
Can do an extension that is SparseArrays+ForwardDiff for these.

hersle:
> Which linear solver? It shouldn't make a difference on PureKLU, and that would likely be pretty fast in a lot cases like this.

Using `KLUFactorization` for now. I saw `PureKLUFactorization` and tried it, but did not see any immediate improvement yet. I could compare the two after getting this improvement in, to make the comparison fair.

> Can do an extension that is SparseArrays+ForwardDiff for these.

Won't that end up duplicating a lot of the machinery in [`ext/LinearSolveForwardDiffExt.jl`](https://github.com/SciML/LinearSolve.jl/blob/53a1799c96bde2607f793cfae839629cb77d6016/ext/LinearSolveForwardDiffExt.jl)? I went with making SparseArray a dep to avoid that mess, and since it is in the stdlib. Is that not an option?

ChrisRackauckas:
Oh wait, you want to do it with https://github.com/SciML/LinearSolve.jl/pull/1023

hersle:
Wasn't #1041 merged in favor of #1023? Should not #1023 be closed?

ChrisRackauckas:
Oh yes, I was confused why I still saw that PR around. So wait, with the forwarddiff opt out you still don't see an advantage to PureKLU with forwarddiff? That's wild because it's a huge difference.

hersle:
No, I don't see any improvement with it. I tried the same code as above, now with `PureKLUFactorization` (on LinearSolve v3.85.2), and I get:
```
 16.977270 seconds (88.85 M allocations: 35.636 GiB, 22.30% gc time)
```
bilde

This is after warm-up. Seems like its doing a lot of allocations, that could perhaps explain parts of it.

ChrisRackauckas:
The A matrix is dual?

hersle:
Yes it should be, its the $W = I - \gamma J$ matrix inside `Rodas5P`, and $J$ should be dual.

ChrisRackauckas:
That's really odd.

hersle:
Yeah I was also expecting it to perform well. Here is `@profview_allocs` on the same code with `PureKLUFactorization`. Maybe throwing Claude at it could eliminate the allocations?

bilde

PR #1054, opened 2026-06-20
Use separate cache flags for A and b in ForwardDiff extension

hersle:
Avoids updating A if only b changes (e.g. in an ODE) and vice versa.


ChrisRackauckas:
This needs tests.

ChrisRackauckas:
Rebased and finished in https://github.com/SciML/LinearSolve.jl/pull/1062, just needed to update but basically was taken in as-is with your commit.

hersle:
That's ok, thanks a lot for following it through! I appreciate it, as my time to follow up is a bit limited :)

Issue #1064, opened 2026-06-28
ForwardDiff autodiff with PureKLUFactorization is very slow through implicit ODE solver


hersle:
Follow-up to https://github.com/SciML/LinearSolve.jl/issues/1052 with a simple reproducible example:
```julia
using ModelingToolkit, OrdinaryDiffEqRosenbrock, LinearSolve, ForwardDiff
using ModelingToolkit: t_nounits as t, D_nounits as D

N = 20
@parameters α
@variables u(t)[1:N]
eqs = [D(u[i]) ~ α * ((i > 1 ? u[i-1] : 0) - 2u[i] + (i < N ? u[i+1] : 0)) for i in 1:N] # 1D heat equation
@mtkbuild sys = ODESystem(eqs, t)
prob0 = ODEProblem(sys, [[u[i] => 1.0 for i in 1:N]; α => 1.0], (0.0, 1.0); jac=true, sparse=true)

setter = ModelingToolkit.setsym_oop(prob0, α)

function usumend(α, alg)
    u0, p = setter(prob0, α)
    prob = remake(prob0; u0, p)
    sol = solve(prob, alg; abstol=1e-10, reltol=1e-10)
    return sum(sol.u[end])
end
usumend_klu(α) = usumend(α, Rodas5P(linsolve=KLUFactorization()))
usumend_pureklu(α) = usumend(α, Rodas5P(linsolve=PureKLUFactorization()))

@time ForwardDiff.derivative(usumend_klu, 1.0) # 0.000826 seconds (1.98 k allocations: 305.623 KiB, 0.05% compilation time: 100% of which was recompilation)
@time ForwardDiff.derivative(usumend_pureklu, 1.0) # 0.006095 seconds (54.58 k allocations: 12.772 MiB, 50.12% gc time, 0.01% compilation time: 100% of which was recompilation)
```
PureKLU is significantly slower and allocates much more. This is a very simple example; I believe it's possible to make it much worse by differentiating more parameters, for example.

On latest LinearSolve and
```julia-repl
julia> versioninfo()
Julia Version 1.12.6
Commit 15346901f0 (2026-04-09 19:20 UTC)
Build Info:
  Official https://julialang.org release
Platform Info:
  OS: Windows (x86_64-w64-mingw32)
  CPU: 12 × AMD Ryzen 5 7600 6-Core Processor              
  WORD_SIZE: 64
  LLVM: libLLVM-18.1.7 (ORCJIT, znver4)
[...]

PR #1065, opened 2026-06-28
Cache inner LinearCache in DualLinearCache in direct dual path

hersle:
A suggested fix to https://github.com/SciML/LinearSolve.jl/issues/1064. This is largely generated with Claude. I admittedly don't understand the internals well enough to know whether this is the right fix to make. Any help is appreciated :)

It fixes the original example in https://github.com/SciML/LinearSolve.jl/issues/1064#issue-4761973882, which now gives:
```julia
@time ForwardDiff.derivative(usumend_klu, 1.0) # 0.000808 seconds (1.98 k allocations: 300.891 KiB, 0.09% compilation time: 100% of which was recompilation)
@time ForwardDiff.derivative(usumend_pureklu, 1.0) # 0.000636 seconds (1.93 k allocations: 315.426 KiB, 0.05% compilation time: 100% of which was recompilation)
```
Now PureKLU is faster than KLU (as it should be?).

### DIFF
--- a/ext/LinearSolveForwardDiffExt.jl
+++ b/ext/LinearSolveForwardDiffExt.jl
@@ -63,6 +63,9 @@ LinearSolve.@concrete mutable struct DualLinearCache{DT}
     dual_A
     dual_b
     dual_u
+
+    # Cached LinearCache for direct Dual path (nothing for split path algorithms)
+    dual_linear_cache
 end
 
 function linearsolve_forwarddiff_solve!(cache::DualLinearCache, alg, args...; kwargs...)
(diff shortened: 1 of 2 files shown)


hersle:
This is my best attempt for now. I think the problem was that `A` was being refactored at every single `solve!` even when `A` is unchanged between the solves. Is this PR a reasonable way to fix that, or is there a better way that achieves the same?

The changes here make PureKLU keep using the direct AD path and makes it faster than KLU. https://github.com/SciML/LinearSolve.jl/pull/1066 instead moves PureKLU to the split AD path, which makes it only as fast as KLU.

Issue #1052, opened 2026-06-20
Major performance regression after opting RFLUFactorization out of split dual AD path


hersle:
@ChrisRackauckas I think something is wrong with https://github.com/SciML/LinearSolve.jl/pull/1041. Found when solving ODEs:
```julia
using OrdinaryDiffEqRosenbrock, LinearSolve, ForwardDiff, RecursiveFactorization

f!(du, u, p, _) = du .= u.^2 .+ p.*u
prob = ODEProblem(f!, randn(500), (0.0, 0.1), [1.0])
alg = Rodas5P(; linsolve = LinearSolve.RFLUFactorization())
loss(p) = solve(remake(prob; p), alg; save_everystep=false, abstol=1e-6, reltol=1e-6).u[end]

@time ForwardDiff.derivative(loss, 1.0);
```

Before c0544d29 (i.e. LinearSolve 3.85.1):
```
  0.035590 seconds (268 allocations: 15.641 MiB)
```

After c0544d29 (i.e. LinearSolve 3.85.2):
```
  1.574712 seconds (1.05 k allocations: 169.178 MiB, 1.27% gc time)
```
This is with `] dev`-ed LinearSolve and `git checkout c0544d29` vs. `git checkout c0544d29~`.

hersle:
Profiling after https://github.com/SciML/LinearSolve.jl/commit/c0544d29757f4b0e09f6c6a9ec35d0bd23fd959b:
```julia
@profview ForwardDiff.derivative(loss, 1.0);
```

<img width="1926" height="1148" alt="Image" src="https://github.com/user-attachments/assets/8623d635-f706-4952-b590-fa0f65024b91" />

hersle:
Also worried that `PureKLUFactorization` may be affected, which was also opted out in the same PR. Could explain why it performs so bad for the example in https://github.com/SciML/LinearSolve.jl/pull/1048.

ChrisRackauckas:
I cannot reproduce it being slower in isolation. There must be some other explanation for your case.

hersle:
Thanks for checking! I will see if I am able to narrow it down.

oscardssmith:
what cpu do you have?

hersle:
@ChrisRackauckas @oscardssmith See https://github.com/SciML/LinearSolve.jl/issues/1064 with an actual reproducible example using PureKLU. CPU is listed there.

PR #1066, opened 2026-06-28
Keep PureKLUFactorization off the direct dual AD path (fixes #1064)

### LINKED ISSUE #1064: ForwardDiff autodiff with PureKLUFactorization is very slow through implicit ODE solver

ChrisRackauckas-Claude:
Fixes #1064.

## Problem

PR #1041 added `PureKLUFactorization` to the direct dual solve path (`_use_direct_dual_solve`), taken whenever `A` carries duals. For ForwardDiff over an implicit ODE solve (e.g. `Rodas5P`), the Rosenbrock `W` matrix carries duals, so every solve factorized `W` in dual arithmetic via PureKLU's pure-Julia KLU.

This is the **same** problem #1052/#1057 fixed for `RFLUFactorization`, which was opted out at the same time. PureKLU was explicitly left on the direct path ("PureKLU are unaffected"), but #1064 shows it is affected identically.

## Fix

Drop `PureKLUFactorization` from `_use_direct_dual_solve` so it stays on the split primal/partials path, which factorizes the primal `W` once in fast `Float64` arithmetic and reuses that factorization across the partial back-solves — instead of re-running the whole numeric factorization in `(2N+1)`-wide scalar dual arithmetic. Adds a regression test asserting PureKLU is not routed to the direct path (mirrors the #1052 RFLU test).

## Local verification (Julia 1.12, Linux)

Exact issue reproducer (tridiagonal heat equation, `N=20`, `Rodas5P`, single derivative):

| | time | alloc |
|---|---|---|
| PureKLU **direct** (before) | 3.79 ms | 7.93 MiB |
| PureKLU **split** (this PR) | 1.56 ms | 227 KiB |
| C-backed `KLUFactorization` (always split) | 1.49 ms | 222 KiB |

Identical derivatives; the split path matches the trusted C-KLU path. `test/Core/forwarddiff_overloads.jl` passes end-to-end (including the existing PureKLU correctness tests for duals in A / in b / both, and the new #1064 regression test). `Runic --check` clean.

## Scope note (the tradeoff is real, not universal)

The split path is the better **default** for the sparse-Jacobian / AD-through-implicit-ODE workload PureKLU+ForwardDiff is actually used in. I swept direct-vs-split across sparsity patterns and dual chunk sizes:

- **High fill** (general sparse / PDE stencils / random sparse): split wins, by a margin that *grows* with fill and chunk size (e.g. `rand-1%`, `n=500`, `chunk=8`: 27.9 ms → 7.7 ms).
- **Very low fill** (tridiagonal / narrow band) with a **large** dual chunk: the direct path can edge ahead in a single shot (~2×), because the factorization is so cheap the split path's per-solve overhead dominates. This is not the regime PureKLU+AD hits.

There is also a **separate latent inefficiency** I found while digging: `_solve_direct_dual!` calls `__init` on a fresh cache every `solve!`, re-doing KLU symbolic analysis and re-allocating the whole cache each time — that is the real source of the 7.93 MiB in the issue. Fixing that is out of scope here; I'll file it separately.

---

### DIFF
--- a/ext/LinearSolveForwardDiffExt.jl
+++ b/ext/LinearSolveForwardDiffExt.jl
@@ -393,15 +393,21 @@ end
 # Check if the algorithm should use the direct dual solve path
 # (algorithms that can work directly with Dual numbers without the primal/partials separation)
 function _use_direct_dual_solve(alg)
-    # NOTE: RFLUFactorization is intentionally *not* on the direct path. Even when
-    # A carries duals, its fast Float64 factorization is BLAS/SIMD-grade, and routing
-    # the Dual problem through it falls back to generic scalar dual arithmetic, losing
-    # that speedup entirely (~40x slower, see issue #1052). The split path keeps the
-    # fast primal factorization and reuses it across the partial back-solves.
+    # NOTE: RFLUFactorization and PureKLUFactorization are intentionally *not* on the
+    # direct path. The split path factorizes the *primal* A once (in fast Float64
+    # arithmetic) and reuses that factorization across the partial back-solves, instead
+    # of re-running the whole numeric factorization in (2N+1)-wide scalar dual
+    # arithmetic. For RFLU this is decisive because its Float64 factorization is
+    # BLAS/SIMD-grade (~40x slower on the direct path, see #1052). For PureKLU it wins
+    # whenever the factorization carries non-trivial fill (general sparse Jacobians,
+    # PDE stencils) and for the small-chunk AD-through-implicit-ODE workload these are
+    # actually used in — see #1064, where the direct path was ~2.4x slower with ~35x
+    # the allocation through a Rosenbrock solve, matching the trusted C-KLU split path
+    # once routed here. (For a single solve of a very-low-fill matrix with a large dual
+    # chunk the direct path can edge ahead, but that is not the regime PureKLU+AD hits.)
     return alg isa GenericLUFactorization ||
         alg isa LinearSolve.SpecializedLUFactorization ||
-        alg isa LinearSolve.SpecializedQRFactorization ||
-        alg isa LinearSolve.PureKLUFactorization
+        alg isa LinearSolve.SpecializedQRFactorization
 end
 
 function _use_direct_dual_solve(alg::DefaultLinearSolver)
--- a/test/Core/forwarddiff_overloads.jl
+++ b/test/Core/forwarddiff_overloads.jl
@@ -280,6 +280,16 @@ plain_A = ForwardDiff.value.(A)
 prob = LinearProblem(sparse(plain_A), b)
 @test ≈(solve(prob, PureKLUFactorization()), plain_A \ b, rtol = 1.0e-9)
 
+# Regression test for #1064: PureKLUFactorization must stay on the split
+# primal/partials path and NOT take the direct dual solve. Like RFLU (#1052),
+# its fast Float64 (KLU) factorization is far cheaper than factorizing the Dual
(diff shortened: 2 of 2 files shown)


hersle:
Is this the right fix to make? With this, PureKLU would effectively just do the same split path as KLU, so I don't think PureKLU would be used to its full potential. Also there is no performance gain.

With the different changes in https://github.com/SciML/LinearSolve.jl/pull/1065, I am able to make PureKLU continue using the direct path, but fix the unnecessary factorizations. This makes PureKLU faster than KLU, so it makes sense to have both.

ChrisRackauckas:
This isn't, no. But I'm also a bit skeptical of your PR. I think it needs to branch at a higher level and deal with non-promoting ldiv! in PureKLU itself.

hersle:
Yeah I understand that. Please improve it when you find a better way :) At least it makes PureKLU perform (close) to its full potential through ForwardDiff, so any future improvements can check that performance against the reference.

ChrisRackauckas:
Yeah your PR had nothing wrong really, but it's the fact that it exposed it wasn't caching was like "oh, wait, if we need to work around that... that means we should just make the opt out path be one dispatch higher... and that just means we just need to not promote..." etc. etc., so a bit of a yak shave but it would solve it at its root cause and should be much better.

Issue #1060, opened 2026-06-25
Better documentation on how to use reuse_symbolic


lijas:
I think the documentation on how to use `reuse_symbolic` for some of the direct solvers is lacking. If I understand everything correctly, I think you have to manually trigger the isfresh field:

```
linsolve = LS.init(prob, solver)
sol1 = LS.solve!(linsolve)
# Update matrix but not pattern
# linsolve.A = 2 * K #BAD: This triggers new refactorization
linsolve.A.nzval .= 2 .* K.nzval #This does not update isfresh to true
linsolve.isfresh = true #Manually trigger isfresh
sol2 = LS.solve!(linsolve)
```

As far I can see, this is not described in the documentation. I would suggest adding a how-to here: https://docs.sciml.ai/LinearSolve/stable/tutorials/caching_interface/

ChrisRackauckas:
I don't understand what you're trying to do here. If you change A, it will check if the pattern changes and symbolic refactorize when necessary (unless reuse_symbolic=false, in which case it will always retrigger that) and then it will do the factorization. If you don't change A... then you don't need to refactorize anyways so I don't see what's going on here?

lijas:
In my FE-code, I have the same sparsity pattern in A for each time step (only the entries changes). As I understand, I should be able to get slighly better performance in the linear solve by reusing the symbolic factorization from the first linear solve, but in order to get this gain, I have to manually set isfresh=true (which is not documented).

I hope this example makes my issue clearer:
```
using SparseArrays, LinearAlgebra
import LinearSolve 

const LS = LinearSolve

function f()
    N = 100
    A1 = sprand(N, N, 0.01) + 10.0*I
    A1 = 0.5*(A1 + A1')
    b = rand(N)

    prob = LS.LinearProblem(A1, b)
    solver = LS.UMFPACKFactorization(; reuse_symbolic=true, check_pattern=false)
    linsolve = LS.init(prob, solver)

    #Solve 1
    print("First solve: ")
    @time sol1 = LS.solve!(linsolve)
    u_sol1 = copy(sol1.u)

    #New timestep: Update values of matrix (but use same pattern)
    linsolve.A.nzval .*= 2

    #Solve2 (reuses full factorization which is not what I want)
    print("Second solve: ")
    @time sol2 = LS.solve!(linsolve)
    @show norm(u_sol1) == norm(sol2.u) #True

    #Solve3 (this only reuses the symbolic factorization):
[...]

PR #1067, opened 2026-06-28
docs: document symbolic factorization reuse for sparse matrices (#1060)

ChrisRackauckas-Claude:

## What

Addresses #1060 (and the follow-up [comment](https://github.com/SciML/LinearSolve.jl/issues/1060#issuecomment-4825411788)). Adds a **"Reusing the Symbolic Factorization with Sparse Matrices"** section to `docs/src/tutorials/caching_interface.md`. The existing tutorial only covered changing `b` and replacing a dense `A`; `reuse_symbolic` was not mentioned anywhere in it.

The new section covers:

- What `reuse_symbolic` / `check_pattern` do for `UMFPACKFactorization` / `KLUFactorization`, and why reusing the symbolic part matters (symbolic analysis is the expensive part; numeric refactor is cheap) — the common FE/Newton/implicit-timestepping case of fixed pattern, changing values.
- The gotcha from the issue: **in-place edits to `A.nzval` are not seen by the cache**, so the next `solve!` silently reuses the stale factorization and returns a wrong result.
- The supported fix: signal the change with `linsolve.A = A` (reassigning the same mutated object is fine and still reuses the symbolic factorization when the pattern is unchanged). A note documents the internal `isfresh` flag as an equivalent-but-not-preferred escape hatch.

## Verification

Ran the new `@example` block locally against this branch:

```
doc example OK; residual = 1.6e-15, changed = 0.24
```

i.e. the post-update solve is correct (tiny residual) and the solution actually changed after the in-place value edit + `linsolve.A = A` (confirming the numeric refactorization fired while the symbolic factorization was reused).

Docs-only change; no source modified.

### DIFF
--- a/docs/src/tutorials/caching_interface.md
+++ b/docs/src/tutorials/caching_interface.md
@@ -73,17 +73,19 @@ time steps of an implicit ODE/PDE integrator) is that `A` keeps the **same
 sparsity pattern** across solves while only its stored values change. In that
 case, the symbolic factorization can be reused and only the numeric
 factorization needs to be redone. This is controlled by the `reuse_symbolic`
-keyword (default `true`) on the sparse factorization algorithms:
+keyword on the sparse factorization algorithms, which **defaults to `true`** —
+so you get this behavior automatically and do not need to set it:
 
 ```julia
-solver = LS.UMFPACKFactorization(; reuse_symbolic = true)
+solver = LS.UMFPACKFactorization()   # reuse_symbolic = true by default
 ```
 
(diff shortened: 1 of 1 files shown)


lijas:
@ChrisRackauckas 
Is it really true that: "For sparse factorizations such as `UMFPACKFactorization` and `KLUFactorization`,
the most expensive part of a solve is often the *symbolic* factorization"? I dont really see much speed up for my case.

ChrisRackauckas:
No depends. I usually see it's like 20-25%

PR #1069, opened 2026-06-29
Route duals-in-b PureKLU solves to a native LinearCache (uses PureKLU 1.1 mixed ldiv!)

### LINKED ISSUE #1064: ForwardDiff autodiff with PureKLUFactorization is very slow through implicit ODE solver

ChrisRackauckas-Claude:
Adds native handling of **duals-only-in-`b`** PureKLU solves: a `DualBLinearProblem` (A primal, b dual) is routed to a plain `LinearCache` and solved with PureKLU's primal-factor / Dual-RHS `ldiv!`, instead of the split `DualLinearCache`. Came out of the #1064 / #1065 discussion (cc @hersle).

Depends on **PureKLU 1.1.0** (SciML/PureKLU.jl#67), now **merged and registered** — the `PureKLU = "1.1"` compat resolves directly.

## What's here

`init(::DualBLinearProblem, ::PureKLUFactorization) = __init(prob, alg)` — routes duals-in-`b` to a native `LinearCache`. Type-stable: dispatch is purely on the problem subtype (A primal / b dual) + alg type, so `init` always returns a `LinearCache` here (verified with `@inferred`). This sidesteps the `Union{LinearCache, DualLinearCache}` instability that caused the value-based opt-out to be reverted (commit `25f159c`), because the subtype already encodes "A primal, b dual." Scope is narrow: only PureKLU; A-dual and mixed-A/b problems still take the split path.

Plus a `PureKLU` `[compat]` bump to `1.1`.

## Scope note (this does NOT fix #1064 by itself)

#1064 is an **A-dual** problem (the Rosenbrock `W` carries duals). This PR only covers the **b-dual** case, so it is **additive**, not a replacement for the A-dual fix (#1066). The two cover different cells:

| dual location | handled by |
|---|---|
| b only (A primal) | **this PR** + PureKLU 1.1 |
| A / A-and-b (the #1064 case) | #1066 (route A-dual PureKLU → split) |

## Why it's worth it

- **Simpler + a latent fix:** the native cache reuses the factorization across b-only `reinit!`s for free. The split path's `reinit!` unconditionally sets `linear_cache.isfresh = true`, so it **refactorizes even when only `b` changed**. For a fixed operator with a varying dual RHS (cache reused), the native path is **5–15× faster** on higher-fill matrices for that reason. (When `A` changes every step, both refactorize and it's a tie — a simplification + fixed-A win, not a claim of being fundamentally faster than split.)
- **Non-allocating:** the solve allocates nothing after warmup — the `n×(N+1)` real RHS buffer lives on the PureKLU factorization (PureKLU#67, `@allocated == 0`).

## Verification (Julia 1.12, Linux)

`test/Core/forwarddiff_overloads.jl` passes end-to-end against **registered PureKLU 1.1.0**; integration testset 12/12 (correctness through the public `solve`, routing returns `LinearCache`, `@inferred` type-stability). `Runic --check` clean.

---


ext/LinearSolveForwardDiffExt.jl
```diff
@@ -728,4 +716,8 @@ function update_partials_list!(partial_matrix::SparseMatrixCSC, list_cache)
     return list_cache
 end
 
+# NOTE: the primal-KLU-factor / Dual-RHS `ldiv!` that the duals-in-b routing above relies on
+# lives in PureKLU itself (`PureKLUForwardDiffExt`, non-allocating via a buffer on the
+# factorization), since PureKLU owns the factorization type. See the PureKLU compat bound.
+
```
ChrisRackauckas:
```suggestion
```

PR #1097, opened 2026-07-17
Drop KrylovPreconditioners from Core tests

ChrisRackauckas-Claude:
Please ignore this PR until it has been reviewed by @ChrisRackauckas.

## Summary

- Replace the external `BlockJacobiPreconditioner` used by the Core tests with a public-API `LinearAlgebra` inverse diagonal preconditioner.
- Remove the test-only `KrylovPreconditioners` compat, extra, target, and import entries.
- Remove LinearSolve's package-specific `no_promote` override from the downgrade workflow.

## Root cause and impact

`KrylovPreconditioners` was used indirectly through its exported `BlockJacobiPreconditioner` name, but its historical Arpack/LightGraphs dependency graph cannot be minimum-resolved with the rest of LinearSolve's test environment. PR #1091 worked around that resolver conflict by excluding the package from minimum promotion.

The existing tests only require an inverse-like preconditioner to exercise the Krylov wrapper. `Diagonal(inv.(diag(A)))` provides that behavior using public `LinearAlgebra` APIs. The GMRES and FGMRES preconditioned cases retain `ldiv = false`, so the inverse-preconditioner multiplication path remains covered. The `(M, I)` tuple continues to cover a nontrivial left preconditioner and identity right preconditioner, and the existing counting factory still verifies `reuse_precs = true` without rebuilding the preconditioner.

This changes only test infrastructure and downgrade coverage; it does not change LinearSolve's runtime dependencies or public API.

## Process

The requested task was to replace the downgrade promotion exception with a focused dependency fix without skipping or weakening tests. I reviewed current `main`, #1091, and the history that introduced the preconditioner tests. That review showed the dependency was used indirectly through an exported symbol, so I preserved the solver and reuse coverage with the documented stdlib preconditioner interface instead of deleting those cases. I then ran the exact downgrade resolver, locked downgrade and current Core suites, formatting, workflow, TOML, and diff checks before committing.

## Local validation

- Ran the Julia 1.10 `julia-actions/julia-downgrade-compat` v2.6.1 resolver in `deps` mode with the reusable workflow's default Mooncake exclusion. Resolution completed successfully, and the resulting manifest did not contain `KrylovPreconditioners`.
- Ran `Pkg.build()` against the downgraded manifest successfully.
- Ran `GROUP=Core` on Julia 1.10 with `force_latest_compatible_version = false` and `allow_reresolve = false`: the full Core suite passed; Basic Tests passed 590/590.
- Built the current Julia 1.12 environment and ran the same strict `GROUP=Core` command: the full Core suite passed; Basic Tests passed 590/590.
- Ran Runic over the full repository: 142/142 Julia files passed.
- Ran `actionlint .github/workflows/Downgrade.yml`, parsed and checked the `Project.toml` test target declarations, and ran `git diff --check`; all passed.

## Prior art

This replaces the package-specific workaround merged in #1091 with a dependency-level fix while preserving the affected preconditioner behavior tests.

### DIFF
--- a/.github/workflows/Downgrade.yml
+++ b/.github/workflows/Downgrade.yml
@@ -17,6 +17,4 @@ jobs:
     with:
       julia-version: "lts"
       group: "Core"
-      # Resolver.jl cannot minimum-resolve KrylovPreconditioners' historical Arpack/LightGraphs graph.
-      no_promote: "Mooncake,KrylovPreconditioners"
     secrets: "inherit"
--- a/Project.toml
+++ b/Project.toml
@@ -151,7 +151,6 @@ JacobiDavidson = "0.1"
 KernelAbstractions = "0.9.30"
 Krylov = "0.10"
 KrylovKit = "0.10"
-KrylovPreconditioners = "0.3"
 LAPACK_jll = "3"
 Libdl = "1.10"
 LinearAlgebra = "1.10"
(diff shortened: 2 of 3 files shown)


ChrisRackauckas-Claude:
CI triage: the downgrade job passed. The three failing jobs are exact clean-base failures from base SHA `0496f135` and are unrelated to this branch: both Julia 1 and Julia LTS AD jobs report the same Static Arrays inferred-return-type mismatch, and the NonlinearSolve downstream job reports the same Generalized Rosenbrock failure plus Brown-almost-linear unexpected pass. Base evidence: [AD Julia 1](https://github.com/SciML/LinearSolve.jl/actions/runs/29449653521/job/87468793497), [AD Julia LTS](https://github.com/SciML/LinearSolve.jl/actions/runs/29449653521/job/87468792913), [NonlinearSolve downstream](https://github.com/SciML/LinearSolve.jl/actions/runs/29449653038/job/87468706413). No branch change is warranted for these failures.

PR #1125, opened 2026-07-29
Import issquare from SciMLOperators in ForwardDiff extension

ChrisRackauckas-Claude:
Ignore until reviewed by @ChrisRackauckas.

## Summary

- import `issquare` directly from its public owner, `SciMLOperators`, in `LinearSolveForwardDiffExt`
- add a focused square split-dual initialization regression

## Root cause

`LinearSolveForwardDiffExt` called bare `issquare` while relying on `using SciMLBase` to make the name available indirectly. SciMLBase commit `b9c3507a` (`Make strict QA owner-contract clean (#1472)`) stopped reexporting SciMLOperators names. With SciMLBase 3.40.1, ForwardDiff initialization of a square linear problem therefore raised:

```text
UndefVarError: `issquare` not defined in `LinearSolveForwardDiffExt`
```

`SciMLOperators.issquare` is exported, has a docstring, and is included in SciMLOperators' rendered interface documentation. The ChainRules extension already uses the same owner import.

## Investigation

- searched open and recent merged LinearSolve PRs; no existing PR fixes this failure
- reproduced the failure on unmodified LinearSolve `fd8271c1` with SciMLBase 3.40.1
- ran the identical probe with registered SciMLBase 3.39.1; it passed with gradient `[0.16, 0.04]`
- applied only the owner import and confirmed the SciMLBase 3.40.1 probe passes with the same gradient

## Verification

- exact `test/Core/forwarddiff_overloads.jl` on SciMLBase 3.40.1: pass
  - new focused test: 1/1
  - all remaining testsets in the file completed successfully
- exact OrdinaryDiffEq nested-AD regression on SciMLBase 3.40.1:
  - nested ForwardDiff over implicit `NonlinearSolveAlg`: 4/4
  - DFBDF initialization with internal ForwardDiff Jacobian: 1/1
- registered SciMLBase 3.39.1 `GROUP=QA`:
  - QA: 20 pass, 2 pre-existing broken
  - JET: 34 pass, 8 pre-existing broken
  - Allocation QA: 48/48
  - SupernodalLU Allocation QA: 8/8
- whole-repository Runic check: pass
- `git diff --check`: pass

[...]

PR #1124, opened 2026-07-29
test: spell out SupernodalLU threshold variable

ChrisRackauckas-Claude:
Ignore this PR until it has been reviewed by @ChrisRackauckas.

## Summary

Rename the local `thr` loop variable in the SupernodalLU dense-kernel tests to `threshold` so Typos does not interpret it as a misspelling.

## Regression evidence

- Clean current `main` (`ed7bcb014f03d15f50e9a5cf86f8a314ab2d248d`) reports exactly two Typos 1.48.0 findings at `test/Core/supernodal_lu.jl:176-177`.
- An automated bisect identified `4c58ce272a79969d45eaa640bdf74c2a437263bf` (#1119) as the first bad commit.

## Local verification

The rebased #1124 tree is byte-identical to the separately tested current-main validation tree (`a11cb43039569967e2bd97e862cf6b32a5b1ba5c`):

- `GROUP=Core JULIA_NUM_THREADS=2 julia +1.12 --project=. -e 'using Pkg; Pkg.test()'` — passed: 2,299 assertions plus three pre-existing broken cases; `SupernodalLU internals` passed 72/72.
- Typos 1.48.0 over the whole repository — passed with zero findings.
- Runic over the changed file and whole repository — passed.
- `git diff --check` — passed.

The remote branch was audited before the lease-protected rebase push; it contained only the prior bot-authored version of this same commit.

PR #1126, opened 2026-07-29
Import cache_operator from SciMLOperators in Core tests

ChrisRackauckas-Claude:
Ignore until reviewed by @ChrisRackauckas.

## Summary

- import `cache_operator` directly from its public owner, `SciMLOperators`, in the Core test that exercises cached inverse preconditioners
- keep this compatibility fix test-only; there are no production-code changes

## Root cause

`test/Core/basictests.jl` called bare `cache_operator` while relying on the transitive chain `using LinearSolve` → reexported SciMLBase names → SciMLOperators.

SciMLBase commit `b9c3507a` (`Make strict QA owner-contract clean (#1472)`) removed the blanket SciMLOperators reexport. The dependency boundary was checked directly:

- parent `7b806c50` (SciMLBase 3.40.0): `cache_operator` is defined after `using SciMLBase`
- child `b9c3507a`: it is not
- registered SciMLBase 3.39.1: the full LinearSolve `GROUP=Core` control passes

`SciMLOperators.cache_operator` is exported/public, has a docstring, and is included in SciMLOperators' rendered interface documentation.

## Verification

- focused cached diagonal-preconditioner regression on SciMLBase 3.40.1: 6/6
- exact `Basic Tests` on SciMLBase 3.40.1, stacked only for validation with the independent `issquare` fix in #1125: 635/635
- full registered SciMLBase 3.39.1 `GROUP=Core`: passed
- whole-repository Runic 1.7.0 check: passed
- `git diff --check`: passed

The full SciMLBase 3.40.1 Core stack continues past this fixed suite and then stops at a separate bare `FunctionOperator` use in `test/Core/default_algs.jl:167`. QA on that dependency likewise exposes separate existing owner-contract failures: stale `has_concretization` in `src/LinearSolve.jl:20` and non-owner `LinearSolve.SciMLOperators.AbstractSciMLOperator` access in `ext/LinearSolveSparseArraysExt.jl:58`. Those are intentionally outside this focused PR.

### DIFF
--- a/test/Core/basictests.jl
+++ b/test/Core/basictests.jl
@@ -1,5 +1,6 @@
 using LinearSolve, LinearAlgebra, SparseArrays, MultiFloats, ForwardDiff
-using SciMLOperators: SciMLOperators, MatrixOperator, FunctionOperator, WOperator
+using SciMLOperators: SciMLOperators, MatrixOperator, FunctionOperator, WOperator,
+    cache_operator
 using RecursiveFactorization, Sparspak, FastLapackInterface
 using IterativeSolvers, KrylovKit, MKL_jll
 using Test


ChrisRackauckas-Claude:
Commit `2589daa7f49a0039d868a6941ed4422ee70c65ba` is intentionally limited to the explicit test-owner import in `test/Core/basictests.jl`.

Local verification plan/result:

1. Reproduce the current SciMLBase 3.40.1 failure on unmodified LinearSolve main.
2. Verify the dependency boundary with registered SciMLBase 3.39.1 and the SciMLBase #1472 parent/child.
3. Verify `cache_operator` is public and rendered at its owning package.
4. Run the affected 3.40.1 suite with the independent #1125 prerequisite, the full registered 3.39.1 Core control, Runic, and `git diff --check`.

All scoped checks completed as recorded in the PR body. The later `FunctionOperator`, `has_concretization`, and qualified-access failures are separate follow-ups and are not being folded into this branch.

ChrisRackauckas-Claude:
Terminal CI audit: 53 checks passed; the five red checks are independent of this one-line `cache_operator` owner import, but two earlier classifications need refinement from the completed logs.

- Spellcheck is the clean-main typo isolated in #1124.
- SciMLSensitivity is the Distributions resolver boundary isolated in SciMLSensitivity#1575.
- ModelingToolkit reports 11 errors: nine are the DelayDiffEq `DEOptions` constructor mismatch isolated in OrdinaryDiffEq#4065; the remaining clean-MTK initialization failures are `Guess Propagation` (missing `Initial` parameter for `x0`) and the constant-RHS ForwardDiff `y0 is ... not an unknown` case, which require their own clean-main investigations.
- QA’s Aqua/ExplicitImports phase actually passed (20 pass, 2 pre-existing broken). The red result is the later JET default-solver test: on the hosted MKL path, BLAS logging calls `string(typeof(A))`, and JET reports 23 possible runtime dispatches through Base type display. Registered local QA passed, so this environment-specific clean-main failure needs a separate reproduction/bisect; it is not an owner-import failure.
- The Downgrade job has no downloadable log archive. GitHub marks the job failed after about 26 minutes while its `julia-runtest` step remains recorded as in-progress, so there is not enough evidence to call this a code failure. Separately, local current-dependency Core testing proceeds past this PR and exposes a missing direct test-owner import for `FunctionOperator`; that is queued as a focused follow-up.

No unrelated changes will be added to #1126.

PR #1127, opened 2026-07-29
Import FunctionOperator from SciMLOperators in Core tests

ChrisRackauckas-Claude:
Ignore until reviewed by @ChrisRackauckas.

## Summary

- Import `FunctionOperator`, `MatrixOperator`, and `WOperator` directly from their owning package, SciMLOperators, at the start of `test/Core/default_algs.jl`.
- Remove the later duplicate owner import.
- Keep this change test-only; LinearSolve already has a direct SciMLOperators dependency.

## Why

SciMLBase's re-export cleanup removed the transitive `FunctionOperator` binding. The boundary is SciMLBase commit `b9c3507a2670fae804bf627ca32354efeee8dc15`: its parent still exposes the binding, while that commit does not. SciMLOperators v1.25.0 publicly exports and documents `FunctionOperator`, so the Core test should import it from SciMLOperators rather than rely on a transitive re-export.

## Verification

- Clean-current focused reproduction before the patch: including `test/Core/default_algs.jl` stops at line 167 with `UndefVarError: FunctionOperator not defined`.
- Rebased focused test against SciMLBase v3.40.1 in an isolated depot:
  `Default Alg Tests | 109 pass | 109 total`.
- Whole-repository Runic check exited 0.
- `git diff --check upstream/main...HEAD` exited 0.

I also ran the official `GROUP=Core` harness with the now-merged #1125 and #1126 fixes applied. It passed the default-algorithm tests (109/109) and continued through the ForwardDiff overload tests (128/128), then stopped at an independent clean-main regression in `test/Core/traits.jl`: `UndefVarError: IdentityOperator not defined in SciMLBase`. That owner-import regression needs a separate focused follow-up and is intentionally not included here.

### DIFF
--- a/test/Core/default_algs.jl
+++ b/test/Core/default_algs.jl
@@ -1,4 +1,5 @@
 using LinearSolve, RecursiveFactorization, LinearAlgebra, SparseArrays, Test
+using SciMLOperators: FunctionOperator, MatrixOperator, WOperator
 
 @test LinearSolve.defaultalg(nothing, zeros(3)).alg === LinearSolve.DefaultAlgorithmChoice.GenericLUFactorization
 prob = LinearProblem(rand(3, 3), rand(3))
@@ -357,7 +358,6 @@ sol_qr2 = solve(
 
 # Regression test for https://github.com/SciML/LinearSolve.jl/issues/890
 # WOperator with init_cacheval overload that unwraps A.J (as OrdinaryDiffEqDifferentiation does)
-using SciMLOperators: WOperator, MatrixOperator
 function LinearSolve.init_cacheval(
         alg::LinearSolve.DefaultLinearSolver, A::WOperator, b, u,
         Pl, Pr,


ChrisRackauckas-Claude:
Investigation/validation scratchpad:

- Clean LinearSolve baseline used for the original reproduction: `fd8271c19613be69afba209e5f117c29f162f8a4`, with SciMLBase v3.40.1.
- Binding probe on that environment: `(SciMLBase = v"3.40.1", FunctionOperator_visible = false)`.
- Directly including the unmodified `test/Core/default_algs.jl` reproduced `UndefVarError: FunctionOperator not defined` at line 167.
- SciMLBase boundary check:
  - parent `7b806c50a9e0419d525374d8b716aa51918876b8`: the transitive binding is visible;
  - first bad `b9c3507a2670fae804bf627ca32354efeee8dc15`: the binding is no longer visible.
- SciMLOperators v1.25.0 audit: `(owner_public = true, owner_exported = true, owner_doc = true)`.
- After rebasing this patch onto current LinearSolve `main`, the isolated focused run completed with:

  ```
  Test Summary:     | Pass  Total     Time
  Default Alg Tests |  109    109  4m15.1s
  ```

- Whole-repository Runic check and `git diff --check upstream/main...HEAD` both exited 0.
- The official current-dependency `GROUP=Core` validation, with #1125 and #1126 applied, passed `Default Alg Tests` 109/109 and `ForwardDiff Overloads` 128/128. It subsequently stopped at the independent clean-main `SciMLBase.IdentityOperator` reference in `test/Core/traits.jl` (5 trait tests passed before the error). That requires its own owner-import investigation and patch.

PR #1137, opened 2026-08-04
Handle review on #1136: one gemv over a reinterpreted ∂A, always use the cached scratch

ChrisRackauckas-Claude:
Draft — **please ignore until reviewed by @ChrisRackauckas.**

Stacks on #1136 (by @andreasnoack) and handles the two review comments there. Andreas's commit is included unchanged as the first commit; the second commit is the response to the review. **This branch still knowingly fails the one existing testset #1136 flagged** — that design question is unresolved and is restated at the bottom.

## Comment 1 — [_"Why wouldn't it match? It should just always use the cache."_](https://github.com/SciML/LinearSolve.jl/pull/1136/files#r3710709177)

Agreed, it always should. The fallback is removed. `partials_scratch` is now allocated in `__dual_init` under exactly the condition that makes the fused method dispatch, so the method reads the field unconditionally:

```julia
partials_scratch = if ∂_A isa DenseMatrix{<:Partials} && ∂_b isa DenseVector{<:Partials}
    V = eltype(eltype(∂_A))
    similar(∂_A, V, ForwardDiff.npartials(eltype(∂_A)), size(∂_A, 1))
else
    nothing
end
```

`∂_A`/`∂_b` keep their type and size for the life of the cache (`setA!`/`setb!` update them in place via `map!`), so a cached scratch cannot go stale. The `_use_direct_dual_solve(alg)` term is dropped from the condition as well: it was sound, but only via a three-hop argument about which `solve!` path a direct-Dual algorithm takes, and the invariant is worth more than the one small matrix it saved. `similar(∂_A, …)` rather than `Matrix{…}(undef, …)` so a GPU-resident `∂A` gets a GPU scratch.

## Comment 2 — [_"This should use a broadcast in order to be GPU compatible?"_](https://github.com/SciML/LinearSolve.jl/pull/1136/files#r3710712658)

The scalar indexing was the real GPU blocker, and it is gone — but not via broadcast, which measures badly. What the hand loop was doing by hand is a transpose that already exists for free: a `Partials{p,V}` stores its `p` components contiguously, so `reinterpret(V, ∂_A)` on a dense array **is** the `(p*m) × n` matrix the loop was assembling, as a strided view rather than a copy. All `p` right-hand sides are then one gemv:

```julia
rhs_flat = vec(scratch)
rhs_flat .= reinterpret(V, ∂_b)
mul!(rhs_flat, reinterpret(V, ∂_A), uu, -1, 1)

for k in eachindex(rhs_list)
    rhs_list[k] .= view(scratch, k, :)
end
```

Nothing indexes an element scalarly. `reinterpret` of a `CuArray` is a `CuArray`, so `mul!` routes to CUBLAS and the scatter is a broadcast over strided views. Still one pass over `∂A`, still nothing materialised.

I did measure a literal broadcast (one broadcast per column of `∂A`, accumulating into the scratch) and rejected it — each column pays full broadcast overhead:

| n | p | loop (#1136) | gemv (here) | per-column broadcast |
|---:|---:|---:|---:|---:|
| 29 | 6 | 1.45 µs | **1.14 µs** | 7.3 µs |
| 29 | 12 | 1.94 µs | 2.06 µs | 13.5 µs |
| 50 | 6 | 3.72 µs | **2.47 µs** | 20.1 µs |
| 50 | 12 | 5.27 µs | **4.87 µs** | 38.2 µs |
| 200 | 12 | 109.2 µs | **29.1 µs** | 600.1 µs |
| 500 | 12 | 1396 µs | **83.8 µs** | 4152 µs |

Kernel in isolation, `Dual{Float64,p}`.

## Allocations

The Dual `solve!` path is allocation-free, and there is now a QA test holding it there — `test/qa/allocations.jl`, `"ForwardDiff Dual solve! is allocation-free"`. It covers the fused dense-∂A kernel on its own plus whole `solve!` for each way partials enter the problem (both, `b` only, `A` only — the latter two take different `xp_linsolve_rhs!` methods), at two shapes so a per-element or per-partial allocation can't hide in a case small enough to unroll. Verified 0 on 1.10.11, 1.11.9 and 1.12.6.

Two things worth knowing if you touch this:

- **Measure behind a function barrier.** At global scope the returned `LinearSolution` can't be proven dead and `@allocated` reports a spurious 32 bytes on every version — larger than the regressions the test guards against, so it would mask them entirely. I initially misread that artifact as a real `LinearSolution` allocation on `main`; it isn't, `main`'s dual `solve!` is 0 too.
- **An intermediate revision of this branch did allocate**, 32 bytes per call (80 on 1.10), from `vec(scratch)` — a fresh 1-D `Array` header aliasing the scratch buffer, which can't be elided because it is handed to BLAS `gemv!` through a `ccall` that takes its address. The `mul!` itself allocates nothing, and the `reinterpret` wrappers are immutable and get elided. Fixed by holding the flat view in the cache next to the `p × m` form, both aliasing one buffer, built once in `__dual_init`. Reintroducing the `vec` fails 4 of the new test's 8 assertions, so the test does bite.

## GPU: the kernel was clean, the path around it wasn't

The fused kernel has no scalar indexing, but that was unfalsifiable until now, because nothing could *reach* it with a GPU-resident `A` — `init` threw first. Four helpers on the way in and out scalar-index:

| | what it did | now |
|---|---|---|
| `nodual_value(::AbstractArray{<:Dual})` | read `first(x)` purely to name the value type | `ForwardDiff.valtype(eltype(x))` — identical for the AbstractFloat and nested-Dual cases, touches no element |
| `partials_to_list` | scalar triple loop per partial | one `getindex.` broadcast per partial index |
| `update_partials_list!` | same, in place | same rewrite |
| `linearsolve_dual_solution!` | reassembled the Dual solution element by element | `ntuple(..., Val(N))` hoists `partials`, leaving a plain elementwise combine that broadcasts |

With those, the whole split-Dual path runs on an `AbstractGPUArray`. **The fused kernel itself needed no change** — it went straight through.

None of it costs anything on CPU; the broadcasts are slightly faster than the hand loops, and `solve!` is still 0 allocations for all three Dual configurations:

| | `init` | A-changing `solve!` |
|---|---:|---:|
| n=29, p=6 | 28.8 → **17.5 µs** | 41.5 → **38.7 µs** |
| n=50, p=12 | 135.6 → **129.9 µs** | 174.7 → **160.5 µs** |

New test is `test/Core/forwarddiff_gpu.jl`, using `JLArray` — a CPU-backed `AbstractGPUArray`, so it runs in the Core group with no GPU and adds only a lightweight pure-Julia test dep. Scalar indexing on it throws unless explicitly permitted and nothing in the file permits it, so this is a proof rather than a smoke test; the premise is pinned with an explicit `@test_throws` so it can't quietly go vacuous if a future GPUArraysCore default changes. It covers all three ways partials enter the problem (each reaching a different `xp_linsolve_rhs!` method), re-solving after `b` is mutated (which is what drives `update_partials_list!` — the first solve skips it), and `A` reassigned between solves, checking values *and* partials against the CPU path each time. 15 assertions.

Each of the four fixes is load-bearing: reverting any one individually fails the testset (4/5, 3/8, 4/5 and 1/14 assertions respectively).

## A regression in #1136 this also fixes

`SparseMatrixCSC{<:Partials} <: AbstractMatrix{<:Partials}`, so #1136's signature captured **sparse** `∂A` too and swept it densely through sparse `getindex`, bypassing the sparsity-aware `update_partials_list!`. Narrowing to `DenseMatrix`/`DenseVector` — which the reinterpret needs anyway — puts sparse back on the list-based method. Whole `solve!`, sparse Dual `A`, `A` reassigned between solves:

| case | main | #1136 | this branch |
|---|---:|---:|---:|
| n=50, p=6 | 11.3 µs | 21.5 µs | **11.3 µs** |
| n=200, p=12 | 76.7 µs | 242.7 µs | **76.6 µs** |

This went unnoticed because `test/Core/forwarddiff_overloads.jl` aborts at the known-failing validity testset (line 627) before the `"Sparse matrices"` testsets at line 659 ever run.

## Timings, remeasured

An earlier revision of this description claimed a ~2x end-to-end regression at n=200 under 64 BLAS threads, attributed to OpenBLAS threads spilling into the single-threaded `setA!` that follows. **That was a measurement artifact and is retracted.** This is a shared 128-core box; those numbers were taken while unrelated Julia suites were loading it. Re-measured round-robin across variants, three rounds, per-variant minimum:

| n | p | `main` fixed | `main` A-chg | #1136 fixed | #1136 A-chg | here (loop) fixed | here (loop) A-chg | here (gemv) fixed | here (gemv) A-chg |
|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| 10 | 4 | 3.36 | 5.52 | 3.03 | 4.48 | 3.02 | 4.45 | 3.01 | 4.45 |
| 29 | 6 | 10.34 | 41.05 | 10.08 | 21.67 | 10.03 | 21.67 | 9.88 | 21.65 |
| 29 | 12 | 19.11 | 65.13 | 18.05 | 36.55 | 18.00 | 36.34 | 17.89 | 36.34 |
| 50 | 12 | 33.78 | 173.40 | 32.51 | 84.64 | 32.04 | 84.32 | 32.13 | 86.72 |
| 100 | 12 | 85.09 | 722.18 | 84.86 | 289.06 | 84.59 | 286.81 | 84.60 | 288.77 |
| 200 | 12 | 242.02 | 3333.71 | 279.19 | 1264.85 | 274.77 | 1254.47 | 278.11 | 1289.55 |

µs, `Dual{Float64,p}`. "A-chg" reassigns `A` between solves, which is the regime the fusion targets; "fixed" re-solves an unchanged cache, where `main` can reuse its cached partials lists.

Reading it: **nothing regressed against #1136 at any size**, and both are 1.9-2.6x faster than `main` in the A-changing regime. The two kernels are within 3% of each other end-to-end everywhere, which is why the cutoff below is insurance rather than a speedup.

One inherited caveat, present in #1136 too and not introduced here: at n=200 with `A` fixed, `main` is ~13% faster (242 vs 275-279), because an unchanged `A` lets it reuse the partials lists it already built while the fused path rebuilds the rhs from ∂A each solve.

## Size cutoff

Small problems now take a hand-written loop; large ones take the gemv. The gemv's advantage is asymptotic — it pays a fixed BLAS call overhead the loop does not:

| kernel only | 5x5 | 10x10 | 20x20 | 50x50 | 200x200 |
|---|---:|---:|---:|---:|---:|
| OpenBLAS, loop/gemv | 0.33 | 0.66 | 1.18 | 1.14 | 1.42 |
| MKL, loop/gemv | 0.30 | 0.58 | 0.78 | 0.75 | 1.11 |

(p=6; >1 means the gemv is faster. Measured at 1, 8 and 64 OpenBLAS threads — the ratios are the same, so gemv is not threaded at these sizes.)

The constant is a hedge, not a sharp optimum: OpenBLAS turns over near 20x20, MKL not until nearer 100x100, and end-to-end the choice is under 3% either way. `DUAL_RHS_GEMV_CUTOFF = 30_000` multiply-adds sits between the two crossovers.

**Only a plain `Array` is eligible for the loop** — it indexes scalarly, which a GPU array forbids — so a `CuArray`/`JLArray` always takes the gemv. The cutoff cannot silently disable the GPU path, and the JLArray tests still pass.

## Still open: the failing testset

Unchanged from #1136 and still #1136's question to answer, not something this PR resolves. `test/Core/forwarddiff_overloads.jl:627` fails 4 of 12 assertions: the fused path never populates `partials_A_list`/`partials_b_list`, so `A_partials_valid`/`b_partials_valid` stay `false` where the testset requires `true` after `solve!`. Setting them `true` would mark stale lists as current for every other consumer (`∂_b::Nothing`, the sparse method, the non-square branch), so it is deliberately not done.

Worth noting the lists are **not** dead for dense after this change — the non-square branch of `linearsolve_forwarddiff_solve!` still consumes them for dense `∂A`. So the coherent resolution looks like keeping the machinery and rewriting the testset around the real invariant ("the lists are populated by the paths that read them"), rather than retiring it. Happy to implement that, or the side-effect variant, once you've picked.

## Verified

Run locally on Julia 1.11.9, `x86_64-linux-gnu`, ForwardDiff 1.3.2:

* A 79-assertion suite covering values *and* extracted partials against `A \ b` at n = 1, 2, 5, 12, 29, 50 × p = 1, 2, 4, 6, 12; scratch identity preserved across `solve!` with allocation independent of problem size; `A` and `b` reassigned between solves; `A`-only and `b`-only Dual problems (no scratch allocated); sparse `∂A` confirmed dispatching to the list-based method and matching `Matrix(A) \ b` in partials; non-square; nested `Dual{outer}` over `Dual{inner}`; `GenericLUFactorization` still on the direct-Dual path. All pass.
* `test/Core/forwarddiff_overloads.jl`: everything before line 627 passes. Running the remainder separately (the file aborts at 627), all 10 remaining testsets pass, including both `"Sparse matrices"` sets that #1136 never reached.
* Runic clean.

https://claude.ai/code/session_01KwhS5qvTPwdKUGHnnA44Zp

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -51,6 +51,7 @@ Ginkgo = "4c8bd3c9-ead9-4b5e-a625-08f1338ba0ec"
 HYPRE = "b5ffcf37-a2bd-41ab-a3da-4bd9bc8ad771"
 HSL = "34c5aeac-e683-54a6-a0e9-6e0fdc586c50"
 IterativeSolvers = "42fd0dbc-a981-5370-80f2-aaf504508153"
+JLArrays = "27aeb0d3-9eb9-45fb-866b-73c2ecf80fcb"
 JacobiDavidson = "11c68b98-9c9b-11e8-267b-bbb95576cead"
 KernelAbstractions = "63c18a36-062a-441e-b654-da1e3ab1ce7c"
 KrylovKit = "0b1a1467-8014-51b9-945f-bf0ae24f4b77"
@@ -148,6 +149,7 @@ HYPRE = "1.7"
 HSL = "0.5"
 InteractiveUtils = "1.10"
 IterativeSolvers = "0.9.4"
+JLArrays = "0.2, 0.3"
 JacobiDavidson = "0.1"
 KernelAbstractions = "0.9.30"
 Krylov = "0.10"
@@ -234,4 +236,4 @@ Zygote = "e88e6eb3-aa80-5325-afca-941959d7151f"
 blis_jll = "6136c539-28a5-5bf0-87cc-b183200dce32"
 
 [targets]
-test = ["AlgebraicMultigrid", "Arpack", "ArnoldiMethod", "BandedMatrices", "BlockDiagonals", "CliqueTrees", "ComponentArrays", "FastAlmostBandedMatrices", "FastLapackInterface", "FiniteDiff", "FixedSizeArrays", "ForwardDiff", "InteractiveUtils", "IterativeSolvers", "JacobiDavidson", "KrylovKit", "LAPACK_jll", "MultiFloats", "Pkg", "PureUMFPACK", "Random", "RecursiveFactorization", "STRUMPACK_jll", "SafeTestsets", "SciMLTesting", "SparseArrays", "Sparspak", "SpecializingFactorizations", "StableRNGs", "StaticArrays", "Test", "Zygote", "blis_jll"]
+test = ["AlgebraicMultigrid", "Arpack", "ArnoldiMethod", "BandedMatrices", "BlockDiagonals", "CliqueTrees", "ComponentArrays", "FastAlmostBandedMatrices", "FastLapackInterface", "FiniteDiff", "FixedSizeArrays", "ForwardDiff", "InteractiveUtils", "IterativeSolvers", "JLArrays", "JacobiDavidson", "KrylovKit", "LAPACK_jll", "MultiFloats", "Pkg", "PureUMFPACK", "Random", "RecursiveFactorization", "STRUMPACK_jll", "SafeTestsets", "SciMLTesting", "SparseArrays", "Sparspak", "SpecializingFactorizations", "StableRNGs", "StaticArrays", "Test", "Zygote", "blis_jll"]
--- a/ext/LinearSolveForwardDiffExt.jl
+++ b/ext/LinearSolveForwardDiffExt.jl
@@ -54,6 +54,10 @@ const DualAbstractLinearProblem = Union{
 
     # Cached intermediate values for calculations
     rhs_list
+    # p x m scratch with the partial index contiguous, used by the fused
+    # xp_linsolve_rhs!. Sized from ∂A's rows, which differ from length(b) for a
+    # non-square system. `nothing` when there is no dense Dual matrix to sweep.
+    partials_scratch
     dual_u0_cache
     primal_u_cache
     primal_b_cache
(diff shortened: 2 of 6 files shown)


ChrisRackauckas:
The QA failure was pre-existing in KLU, so I'll merge and handle that elsewhere.

PR #1139, opened 2026-08-05
Fall back to pivoted QR when the default hits a rank-deficient least-squares problem

### LINKED ISSUE #531: `QRFactorization` returning trivial result for least square solve
```julia
using LinearAlgebra, LinearSolve

A = rand(10, 4)
A[:, 1] .= 0
f = rand(10)

prob = LinearProblem(A, f)
sol = solve(prob, QRFactorization())

@show sol.u
@show A \ f
```

```julia
sol.u = [0.0, 0.0, 0.0, 0.0]
A \ f = [0.0, -0.00044412530695698443, 0.9178269036029512, 0.05768039961239375]
```

AJ0070:
## Checklist

## Additional context

Fixes #531.

- `defaultalg` picks an unpivoted `QRFactorization` for overdetermined dense systems. It is not rank-revealing, so a rank-deficient `A` returns `cache.u` untouched (all zeros) with `ReturnCode.Failure`, while `A \ b` gives the correct least-squares solution.
- Worse variant not in the issue: with a nearly dependent column, no exact zero shows up on the `R` diagonal and the result is finite, so a wrong answer is returned with a success retcode. `A[:, 1] .= 1e-14 .* A[:, 2]` gives `[4.33e28, -4.33e14, 0.020, 0.305]` where `A \ b` gives `[0.0, 0.809, 0.100, 0.286]`.
- Pivoting unconditionally would fix both but taxes every full-rank solve: `qr!(A, ColumnNorm())` measured 1.5x to 2.8x `qr!(A, NoPivot())` on typical tall-skinny shapes here (1000x100: 2.80 ms vs 7.82 ms).
- Instead, unpivoted QR stays on the fast path and now hooks into the safety-fallback machinery that already exists for LU.
- `_qr_rank_deficient(F)` scans the `R` diagonal in `O(min(m, n))` against `min(m, n) * eps`, the threshold LAPACK's `xGELSY` (and so `A \ b`) uses to truncate the rank. It dispatches on the factorization's storage, so SPQR, GPU, and `Dual` factorizations keep their current behavior.
- The check is a heuristic, since unpivoted QR does not order `diag(R)` by magnitude, and can miss a Kahan-style deficiency. The docstring says so.
- `_default_qr_solve_with_fallback` re-solves with `QRFactorization(ColumnNorm())` on failure, non-finite output, or rank deficiency. It reuses `_do_qr_fallback`, so `A` restoration and `fell_back_to_qr` cache reuse work as they do for LU. Its guards are decided by the cache's type, so the branch folds away at compile time.
- Both the exactly-singular and numerically-singular cases now reproduce `A \ b`, `@inferred solve!` still returns a concrete `LinearSolution`, and full-rank problems never enter the fallback.
- An explicit `QRFactorization()` or `LUFactorization()` still reports `ReturnCode.Failure` on rank-deficient input, matching the conclusion in the issue thread. The `QRFactorization` docstring and `solvers.md` now say so and point at `ColumnNorm()`, `SVDFactorization`, and the least-squares Krylov methods.
- Two assertions in `test/Core/default_algs.jl` encoded the old failure behavior and now assert the corrected result. The second one's purpose (the default alg choice must not bake in a rank check, since rank can change across cache reuses) is unaffected, and it now also covers the fallback firing when rank drops mid-cache.
- New coverage in `test/Core/nonsquare.jl`: tall with a zero column, tall with a duplicated column, tall numerically rank-deficient, wide rank-deficient, square singular under `VeryIllConditioned`, cache reuse across the fallback, and a full-rank case leaving `fell_back_to_qr` unset.
- Full `Core` group passes locally on Julia 1.12.6 (25 testsets, 0 failures).

AI Disclosure: Used Opus 5

### DIFF
--- a/src/LinearSolve.jl
+++ b/src/LinearSolve.jl
@@ -494,36 +494,44 @@ using the same relative threshold LAPACK's `xGELSY` (and therefore `A \\ b`) use
 to truncate the rank: a factorization is called rank-deficient when the smallest
 `|R[i, i]|` falls at or below `min(m, n) * eps * max|R[i, i]|`.
 
+This scans the `R` diagonal rather than reading a LAPACK `info` because
+unpivoted QR has none to read: `geqrf`/`geqrt` only set `info < 0` for an
+illegal argument and return `info == 0` on an exactly rank-deficient matrix, and
+`QRCompactWY` correspondingly stores only `factors` and `T` with no `issuccess`
+method. That is the same reason `_notsuccessful(::QRCompactWY)` above hand-scans
+for an exact zero on the diagonal; this is that scan with a relative threshold,
+so it also catches a merely negligible entry.
+
 Unpivoted QR does not order the diagonal of `R` by magnitude, so this is a
-heuristic rather than a rank-revealing test — it can miss a deficiency that only
+heuristic rather than a rank-revealing test: it can miss a deficiency that only
 column pivoting would expose (Kahan-style matrices). It exists so the default
 algorithm can keep unpivoted QR on the fast path and only pay for a pivoted
 refactorization when the cheap check says the answer would otherwise be garbage;
-see `_default_qr_solve_with_fallback`. Returns `false` for factorization types
-where the test does not apply (SPQR, GPU factorizations without scalar indexing),
-which leaves the existing behavior untouched.
+see `_default_qr_solve_with_fallback`. Factorization types the test does not
+apply to (SPQR, GPU) return `false` and keep their existing behavior.
 """
 @inline _qr_rank_deficient(F) = false
 
+# GPU factorizations cannot be indexed elementwise. Mirrors the GPU method on
+# `_notsuccessful` above, and is constrained on the same `T` as the scanning
+# method below so that it is strictly more specific (no dispatch ambiguity).
 @inline function _qr_rank_deficient(
-        F::Union{LinearAlgebra.QRCompactWY, LinearAlgebra.QR}
-    )
-    # Dispatch on the storage so GPU factorizations (which cannot be indexed
-    # elementwise) and exotic element types fall through to the `false` method
-    # below without an ambiguity against the type-parameter-constrained methods.
-    return _qr_rank_deficient(F, F.factors)
+        ::LinearAlgebra.QRCompactWY{
+            T, A,
+        }
+    ) where {T <: BLASELTYPES, A <: GPUArraysCore.AnyGPUArray}
+    return false
 end
 
-@inline _qr_rank_deficient(F, factors) = false
-
 @inline function _qr_rank_deficient(
-        F, factors::AbstractMatrix{T}
-    ) where {T <: BLASELTYPES}
-    ArrayInterface.fast_scalar_indexing(factors) || return false
+        F::LinearAlgebra.QRCompactWY{
+            T, A,
+        }
+    ) where {T <: BLASELTYPES, A}
     (m, n) = size(F)
     mn = min(m, n)
     mn == 0 && return false
-    R = view(factors, 1:mn, 1:n)
+    R = view(F.factors, 1:mn, 1:n)
     dmin = typemax(real(T))
     dmax = zero(real(T))
     for x in @view R[diagind(R)]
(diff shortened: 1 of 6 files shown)


src/LinearSolve.jl
```diff
@@ -486,6 +486,54 @@ end
...
+    mn == 0 && return false
+    R = view(factors, 1:mn, 1:n)
+    dmin = typemax(real(T))
+    dmax = zero(real(T))
+    for x in @view R[diagind(R)]
+        a = abs(x)
+        a < dmin && (dmin = a)
+        a > dmax && (dmax = a)
+    end
+    return dmin <= mn * eps(real(T)) * dmax
```
ChrisRackauckas, line 542: Is this GPU safe? Easier is to just check the blas return info and go based on that.

AJ0070: Yes on GPU safety, but you are right that it was resting on something too indirect. It relied on `ArrayInterface.fast_scalar_indexing`, whose `false` for GPU arrays comes from a weak-dependency extension. Pushed 69b6480 to dispatch on the factorization type instead, mirroring the GPU method on `_notsuccessful` just above it. Both methods constrain the same element type so the GPU one is strictly more specific, and `detect_ambiguities` reports none. GPU also never reaches the check anyway: the caller is guarded on `cache.A isa DenseMatrix` and `_qr_fallback_pivot(cache.A) isa ColumnNorm`, and that helper returns `NoPivot()` for GPU arrays.

On the BLAS info: there isn't one for unpivoted QR. On 1.12:

```
QRCompactWY fieldnames:               (:factors, :T)      # no info field
hasmethod(issuccess, (QRCompactWY,)): false
geqrf! on a matrix with a zero column: info == 0
R diagonal after geqrf!:              [0.0, -1.73, -1.33, -0.882]
```

`geqrf`/`geqrt` only set `info < 0` for an illegal argument, never for rank deficiency, which is what `geqp3`/`gelsy` are for. LU is the contrasting case and does carry `.info` (3 on a singular test matrix) plus `issuccess`, which is what the existing LU fallback keys off.

That is why `_notsuccessful(::QRCompactWY)` directly above already hand-scans the `R` diagonal for an exact zero instead of reading an info. This is the same scan with a relative threshold (`min(m, n) * eps * max|R[i,i]|`, the rcond `xGELSY` uses) rather than exact equality, which is what catches the near-deficient case. That case is the one that currently returns `[4.33e28, -4.33e14, ...]` with `ReturnCode.Success`, so an exact-zero test alone would not close it.

The genuinely simpler option is pivoted QR by default, since `geqp3` does give a rank estimate, but that measured 1.5x to 2.8x on typical tall-skinny shapes (1000x100: 2.80 ms vs 7.82 ms), which is why I kept the cheap check on the fast path instead.

AJ0070:
Correction to this PR's description, found after merge.

The "tall, numerically rank-deficient" test added here is flaky, at roughly a 3% failure rate on any job. It builds its matrix with an unseeded `rand` and a `1e-14` column scaling, which puts it exactly on the rank-detection cutoff. `_qr_rank_deficient` compares against `min(m, n) * eps * max|R[i, i]|`, and once the second column is orthogonalized against the first, `R[2, 2]` is eps-level noise straddling that threshold. Over 300 seeds the ratio ran from `4.2e-17` to `1.1e-15` against a cutoff of `8.9e-16`. On the draws that land above it, the check correctly declines to fire, unpivoted QR returns the overflowing solution, and the comparison against `A \ b` fails. It first surfaced on a `Downgrade` run for an unrelated PR.

Fixed in #1150 by seeding the testset and scaling by `1e-18` instead. The other six rank-deficient testsets added here were checked the same way over 300 draws and are robust.

The substantive point, which the description above overstated: this fixes the near-deficient case only when the deficiency falls below the LAPACK rcond threshold. That is the same threshold `\` uses, so matching `\` remains the right contract, but a matrix sitting *on* the boundary can legitimately disagree between unpivoted QR plus this heuristic and the pivoted QR behind `\`, which has better numerics there. Clearly rank-deficient input is handled; input engineered to sit at the cutoff is inherently ambiguous. #1150 records that on `_qr_rank_deficient`'s docstring and points callers who need a decision there at `QRFactorization(ColumnNorm())` or `SVDFactorization()`.

PR #1140, opened 2026-08-05
Expose PureKLU pivot tolerance option

hersle:
Follow-up to https://github.com/SciML/PureKLU.jl/pull/76


ChrisRackauckas (approved):

hersle:
I think the relevant tests are OK.

PR #1131, opened 2026-08-01
Release 5.4.0 so the ForwardDiff extension issquare fix ships

ChrisRackauckas-Claude:
**Please ignore this PR until it has been reviewed by @ChrisRackauckas.**

## Problem

The newest registered LinearSolve, v5.3.0, is broken against SciMLBase >= 3.40.0. Any `solve` of a `LinearProblem` with `ForwardDiff.Dual` entries throws:

```
ERROR: UndefVarError: `issquare` not defined in `LinearSolveForwardDiffExt`
Hint: a global variable of this name also exists in SciMLOperators.
Stacktrace:
 [1] solve(::LinearProblem{Nothing, true, Matrix{ForwardDiff.Dual{...}}, ...}, ::Nothing)
   @ LinearSolveForwardDiffExt ~/.julia/packages/LinearSolve/Jvqzc/ext/LinearSolveForwardDiffExt.jl:335
```

Root cause: `LinearSolveForwardDiffExt` used `issquare` unqualified. It resolved only because `using SciMLBase` inside the extension pulled it in via SciMLBase's `@reexport using SciMLOperators`. SciMLBase v3.40.0 removed that re-export (it now goes through SciMLOperators as the owner of the operator API), so the name disappeared from the extension's scope.

This is already fixed on master by f1908b92 (#1125), which adds `using SciMLOperators: issquare` to the extension. The problem is that master still carries `version = "5.3.0"` — the same version already in the General registry — so the fix is not installable by anyone.

## What this PR does

Bumps the version to 5.4.0 so the fix can be registered. Minor rather than patch because master has also gained the TriangularSolve-backed `LinearSolveRecursiveFactorizationExt` (new weakdep + extension trigger) and the SupernodalLU work since v5.3.0.

## Verification

Minimal reproduction (Julia 1.12.4):

```julia
using LinearSolve, ForwardDiff
A = [ForwardDiff.Dual(2.0, 1.0) 1.0; 1.0 3.0]
b = [ForwardDiff.Dual(1.0, 0.0), 2.0]
solve(LinearProblem(A, b))
```

- LinearSolve v5.3.0 (registered) + SciMLBase v3.41.0 -> `UndefVarError: issquare not defined in LinearSolveForwardDiffExt`
- LinearSolve v5.3.0 (registered) + SciMLBase v3.39.1 pinned -> works (`Dual{Nothing}(0.2,-0.12), Dual{Nothing}(0.6,0.04)`), confirming SciMLBase 3.40.0 is the trigger
[...]

Issue #1058, opened 2026-06-25
QA: ExplicitImports tracked @test_broken markers (qualified-accesses-public, no-implicit-imports)


ChrisRackauckas-Claude:
## ExplicitImports QA: tracked `@test_broken` markers

The QA suite (`test/qa/qa.jl` + `lib/*/test/qa/qa.jl`) now runs the full
ExplicitImports.jl check set via `SciMLTesting.run_qa(...; explicit_imports = true)`.
Most checks pass or are handled with curated `ignore`-lists for non-public names
of stdlib / other SciML packages. A few checks have too many findings to ignore
name-by-name and are tracked as `@test_broken` (via `ei_broken`) until fixed:

### Root `LinearSolve` — `all_qualified_accesses_are_public`
~90 qualified accesses (`LinearSolve.x`) of names that are not declared `public`:
- LinearSolve's own internals reached via `LinearSolve.x` from extension modules
  (e.g. `defaultalg`, `do_factorization`, `init_sparse_reduction`, `init_cacheval`,
  `DefaultLinearSolver`, `QRFactorization`, ...).
- stdlib / SciMLBase / LinearAlgebra internals (`@max_methods`, `Experimental`,
  `BlasInt`, `build_linear_solution`, `checksquare`, `chkfinite`, ...).

Fix path: declare the relevant LinearSolve internals `public`, and/or upstream
`public` declarations in SciMLBase / SciMLOperators (tracked separately). Once the
finding count is tractable, replace `ei_broken` with an `ignore`-list or drop it.

### Sublibraries `LinearSolvePyAMG`, `LinearSolveAutotune` — `no_implicit_imports`
Both sublibraries `using` whole packages (`LinearAlgebra`, `SparseArrays`,
`PythonCall`, `Plots`, ...) and rely on names brought in implicitly. Making every
such name an explicit import is a source refactor; tracked as `@test_broken` until
done.

This is a QA-tracking issue opened alongside the run_qa v1.6 conversion PR; it can
be closed once the markers above are removed.

ChrisRackauckas-Claude:
Surfaced during the run_qa v1.6 conversion (#1059).

[...]

PR #1144, opened 2026-08-07
Error clearly when PureUMFPACKFactorization is used without PureUMFPACK

### LINKED ISSUE #1143: Cannot solve ODE with PureUMFPACKFactorization
Curious to try PureKLU vs PureUMFPACK, but:
```julia
using ModelingToolkit, LinearSolve, OrdinaryDiffEqRosenbrock
@variables x(t) y(t)
eqs = [D(x) ~ y, D(y) ~ x]
@mtkbuild M = System(eqs, t)
prob = ODEProblem(M, [x => 1.0, y => 2.0], (1.0, 2.0); sparse = true)
sol = solve(prob, Rodas5P(linsolve = PureKLUFactorization())) # succeeds
sol = solve(prob, Rodas5P(linsolve = PureUMFPACKFactorization())) # fails
```

```
ERROR: MethodError: no method matching do_factorization(::PureUMFPACKFactorization, ::SparseArrays.SparseMatrixCSC{…}, ::Vector{…}, ::Vector{…})
The function `do_factorization` exists, but no method is defined for this combination of argument types.

Closest candidates are:
  do_factorization(::BunchKaufmanFactorization, ::Any, ::Any, ::Any)
   @ LinearSolve C:\Users\herma\.julia\packages\LinearSolve\Re7fI\src\factorization.jl:926
  do_factorization(::LUFactorization, ::Any, ::Any, ::Any)
   @ LinearSolve C:\Users\herma\.julia\packages\LinearSolve\Re7fI\src\factorization.jl:389
  do_factorization(::SVDFactorization, ::Any, ::Any, ::Any)
   @ LinearSolve C:\Users\herma\.julia\packages\LinearSolve\Re7fI\src\factorization.jl:864
  ...

Stacktrace:
  [1] macro expansion
    @ C:\Users\herma\.julia\packages\LinearSolve\Re7fI\src\factorization.jl:16 [inlined]
  [2] solve!(cache::LinearSolve.LinearCache{…}, alg::PureUMFPACKFactorization; kwargs::@Kwargs{…})
    @ LinearSolve C:\Users\herma\.julia\packages\LinearSolve\Re7fI\src\factorization.jl:1
  [3] solve!(::LinearSolve.LinearCache{…}; kwargs::@Kwargs{…})
    @ LinearSolve C:\Users\herma\.julia\packages\LinearSolve\Re7fI\src\common.jl:685
  [4] dolinsolve(integrator::OrdinaryDiffEqCore.ODEIntegrator{…}, linsolve::LinearSolve.LinearCache{…}; A::SparseArrays.SparseMatrixCSC{…}, linu::Nothing, b::Vector{…}, reltol::Float64)
    @ OrdinaryDiffEqDifferentiation C:\Users\herma\.julia\packages\OrdinaryDiffEqDifferentiation\mNaGk\src\linsolve_utils.jl:37
  [5] perform_step!(integrator::OrdinaryDiffEqCore.ODEIntegrator{…}, cache::OrdinaryDiffEqRosenbrock.RosenbrockCache{…}, repeat_step::Bool)
    @ OrdinaryDiffEqRosenbrock C:\Users\herma\.julia\packages\OrdinaryDiffEqRosenbrock\S9V2V\src\rosenbrock_perform_step.jl:762
  [6] perform_step!(integrator::OrdinaryDiffEqCore.ODEIntegrator{…}, cache::OrdinaryDiffEqRosenbrock.RosenbrockCache{…})
    @ OrdinaryDiffEqRosenbrock C:\Users\herma\.julia\packages\OrdinaryDiffEqRosenbrock\S9V2V\src\rosenbrock_perform_step.jl:730
  [7] solve!(integrator::OrdinaryDiffEqCore.ODEIntegrator{…})
    @ OrdinaryDiffEqCore C:\Users\herma\.julia\packages\OrdinaryDiffEqCore\NVUZl\src\solve.jl:907
  [8] #__solve#90
    @ C:\Users\herma\.julia\packages\OrdinaryDiffEqCore\NVUZl\src\solve.jl:10 [inlined]
  [9] __solve
    @ C:\Users\herma\.julia\packages\OrdinaryDiffEqCore\NVUZl\src\solve.jl:1 [inlined]
 [10] #solve_call#26
    @ C:\Users\herma\.julia\packages\DiffEqBase\kdaLg\src\solve.jl:187 [inlined]
 [11] solve_up(prob::ODEProblem{…}, sensealg::Nothing, u0::Vector{…}, p::MTKParameters{…}, args::Rodas5P{…}; originator::SciMLBase.ChainRulesOriginator, kwargs::@Kwargs{…})
    @ DiffEqBase C:\Users\herma\.julia\packages\DiffEqBase\kdaLg\src\solve.jl:661
 [12] solve_up
    @ C:\Users\herma\.julia\packages\DiffEqBase\kdaLg\src\solve.jl:634 [inlined]
 [13] solve(prob::ODEProblem{…}, args::Rodas5P{…}; sensealg::Nothing, u0::Nothing, p::Nothing, wrap::Val{…}, verbose::DiffEqBase.DEVerbosity{…}, kwargs::@Kwargs{})
    @ DiffEqBase C:\Users\herma\.julia\packages\DiffEqBase\kdaLg\src\solve.jl:618
 [14] solve(prob::ODEProblem{…}, args::Rodas5P{…})
    @ DiffEqBase C:\Users\herma\.julia\packages\DiffEqBase\kdaLg\src\solve.jl:606
 [15] top-level scope
    @ REPL[38]:1
```

ChrisRackauckas-Claude:
Fixes #1143.

> [!NOTE]

> [!IMPORTANT]
> **Blocked on PureUMFPACK 1.0.0 registering.** This sets `PureUMFPACK = "1"`, which is unresolvable until SciML/PureUMFPACK.jl#36 merges and releases. CI will fail to resolve before then.

## The bug

`PureUMFPACKFactorization` was the only extension-backed factorization without a load guard in its constructor. `PureUMFPACK` is a weakdep and `LinearSolvePureUMFPACKExt` holds the algorithm's only `SciMLBase.solve!` method, so without `using PureUMFPACK` the algorithm constructed fine and then fell through to the generic `solve!(::LinearCache, ::AbstractFactorization)` at `src/factorization.jl:1`, which calls `do_factorization` — for which no method exists:

```
ERROR: MethodError: no method matching do_factorization(::PureUMFPACKFactorization,
    ::SparseMatrixCSC{Float64, Int64}, ::Vector{Float64}, ::Vector{Float64})
```

`PureKLUFactorization` works in the reporter's script only because `PureKLU` is a *hard* dependency. Nothing was wrong with the solver itself — the issue's ODE reproducer gives correct results once the package is loaded.

## Changes

**Constructor guard.** `PureUMFPACKFactorization` gains `throwerror = true` and an extension check, matching `SparspakFactorization` (`src/factorization.jl:1872`), `CliqueTreesFactorization`, `STRUMPACKFactorization`, `PardisoJL`, and the CUDA/Metal/HSL/MUMPS algorithms. The error is now:

```
PureUMFPACKFactorization requires that PureUMFPACK is loaded, i.e. `using PureUMFPACK`
```

**Extension solves with `\`.** PureUMFPACK 1.0 unexports `solve` and only marks it `public` on Julia >= 1.11, so `PureUMFPACK.solve(F, cache.b)` would read as a non-public qualified access when QA runs on the 1.10 LTS. `\` is `solve(F, b)` with the default `refine = 0`, so behavior is identical, and no new ExplicitImports ignore entry is needed. The now-unused `PureUMFPACK` module import is dropped.

**Compat `PureUMFPACK = "1"`.** Pinned rather than `"0.1, 1"` so the error message's advice (plain `using PureUMFPACK`) is always correct — on 0.1 that form shadows `solve` and breaks `solve(prob, alg)` (the reason for SciML/PureUMFPACK.jl#36).

**Regression test.** A subprocess asserts the extension is absent, that the constructor errors, and that `throwerror = false` still builds. The test file's header also switches from `import PureUMFPACK` to plain `using PureUMFPACK`, deliberately, to lock in that `solve` is no longer shadowed.

No version bump — this repo bumps versions in separate release PRs (#1138, #1131).

## Verification

Run locally on Julia 1.11 with PureUMFPACK 1.0.0 `Pkg.develop`'d:

| Suite | Result |
|---|---|
| `GROUP=LinearSolvePureUMFPACK` | 9/9 pass |
| `GROUP=Core` | `Testing LinearSolve tests passed` |
| Issue #1143 reproducer | `PureKLU` and `PureUMFPACK` agree to ~1e-14 |

`test/Core/resolve.jl` already does `using PureUMFPACK: PureUMFPACK`, so the extension is loaded there and the new guard does not fire.

ExplicitImports run directly against `LinearSolvePureUMFPACKExt` reports only `getcolptr`, `@get_cacheval`, `LinearCache` and `init_cacheval` — all four already covered by the ignore lists in `test/qa/qa.jl`, and none of them `solve`. This PR adds no new QA violation.


ext/LinearSolvePureUMFPACKExt.jl
```diff
@@ -76,7 +76,10 @@ function SciMLBase.solve!(
     # A zero on U's diagonal means a singular (or numerically singular) factor;
     # PureUMFPACK with `check = false` produces it instead of throwing.
     return if !any(iszero, diag(F.U))
-        y = PureUMFPACK.solve(F, cache.b)
+        # `\` rather than `PureUMFPACK.solve`: the latter is unexported and only
+        # marked `public` on Julia >= 1.11, so it reads as a non-public access on the
+        # LTS. `\` is `solve(F, b)` with the default `refine = 0`.
+        y = F \ cache.b
```
ChrisRackauckas: No, this isn't PureUMFPACK... it needs to use the PureUMFPACK solve on all versions...

ChrisRackauckas-Claude: Fixed in d248eec4 — restored `PureUMFPACK.solve(F, cache.b)` on all versions.

For the record on the behavior question: `PureLU` is a plain struct (not `<: Factorization`) and `Base.:\(F::PureLU, b::AbstractVector) = solve(F, b)` at `PureUMFPACK/src/interface.jl:179` is its only `\` method, so it did land on PureUMFPACK's solve — which is why the numbers matched. But you're right that it should not go through a one-line forward in a package we do not control, and the version-dependence was bad on its own.

The concern that motivated `\` also turned out to be unfounded. I was worried `PureUMFPACK.solve` would trip ExplicitImports on the 1.10 LTS, where PureUMFPACK cannot declare it `public`. It does get flagged there — confirmed on 1.10:

```
- `solve` is not public in `PureUMFPACK` but it was imported from `PureUMFPACK`
  at ext/LinearSolvePureUMFPACKExt.jl:79:25
```

but `all_qualified_accesses_are_public` is already in `ei_broken` in `test/qa/qa.jl` (~90 pre-existing violations, tracked in #1058), so it never fails CI and needs no ignore entry. Reverting was free.

Running the QA group locally to check that also surfaced a real blocker I had missed: `test/qa/Project.toml` carried a **second** `PureUMFPACK = "0.1"` compat pin, which made the whole QA group unresolvable against PureUMFPACK 1. Bumped in the same commit.

Issue #1148, opened 2026-08-07
JET: KLUFactorization solve is no longer dispatch-clean; `getproperty` self-recursion makes `p, i, x, z = klu._L` destructure a SparseMatrixCSC


AJ0070:
## Summary

`test/qa/jet.jl:136` is failing:

```julia
JET.@test_opt solve(prob_sparse, KLUFactorization())
```

It is the only sparse test in that testset not marked `broken = true` (its `UMFPACKFactorization` and `CHOLMODFactorization` neighbours are, for genuine stdlib reasons), and the comment above it records that it used to pass:

```julia
# Passes since the 5.0 lightweight solution: with the cache no longer
# captured in the returned LinearSolution, the KLU solve is dispatch-clean.
```

So this is a regression rather than a known limitation. It is also what currently keeps the QA job red on `main`: `jet.jl` runs before `qa.jl` in the `qa` group, so it aborts the group before the Aqua/ExplicitImports/api-docs checks ever run.

This is **not** an stdlib false positive. All 40 reported errors trace back to one line in LinearSolve's vendored KLU code.

## Root cause

`src/KLU/klu.jl:395`, inside `getproperty`:

```julia
return if s ∈ (:L, :U, :F)
    if s === :L
        p, i, x, z = klu._L
    elseif s === :U
        p, i, x, z = klu._U
    elseif s === :F
        p, i, x, z = klu._F
    end
```

`klu._L` reads back through this same `getproperty`. Inference abandons the self-recursive call, so the right-hand side is inferred as a Union spanning every branch of the function -- including the `SparseMatrixCSC` that the `s ∈ (:L, :U, :F)` branch returns immediately below. The destructure therefore admits `indexed_iterate(::SparseMatrixCSC, ::Int)`, i.e. **scalar indexing into a sparse matrix**, and everything JET reports hangs off that:

```
_extract!(klu::KLU.KLUFactorization{…}; …) @ LinearSolveSparseArraysExt.KLU src/KLU/klu.jl:252
 └ indexed_iterate(I::SparseMatrixCSC, i::Int64) @ Base ./tuple.jl:165
   └ iterate(A::SparseMatrixCSC) @ Base ./abstractarray.jl:1233
     └ getindex(A::SparseMatrixCSC, I::CartesianIndex{2}) @ Base ./abstractarray.jl:1342
       ├ runtime dispatch detected: Base.Fix1(Base.to_index, A::SparseMatrixCSC)::…
[...]

PR #1149, opened 2026-08-07
Resolve the Aqua method ambiguities and drop the QA escape hatch

### LINKED ISSUE #1107: Strict Aqua QA: resolve method ambiguities and MKL_jll stale dependency
## Reproduction

On clean upstream `main` at `ebb23fe0`, install SciMLTesting v2.4 and run:

```julia
using LinearSolve, SciMLTesting
run_qa(LinearSolve)
```

Aqua reports 11 method ambiguities and `MKL_jll` as stale. Both are present before the public-doc / no-reexport migration.

## Findings

- The ambiguities are between LinearSolve factorization/default-algorithm methods and `LinearSolveSparseArraysExt`; the strict QA output lists the required intersection methods.
- `MKL_jll` is conditionally imported through `Preferences.@load_preference`, so Aqua does not recognize the runtime dependency use.

The strict migration removes the previous `aqua_broken = (:ambiguities,)` configuration rather than preserving it. ExplicitImports findings are tracked separately in #1058; the SciMLBase/CommonSolve interface-piracy classification is tracked in SciML/SciMLTesting.jl#31.

AJ0070:
## Checklist

## Additional context

Fixes #1107.

- `run_qa` carried `aqua_broken = (:ambiguities,)` as a placeholder. Removing it surfaces 13 ambiguities (the issue reported 11), in three families, all fixed by defining the intersection methods. No test additions: the Aqua ambiguity check itself is the test, and it now runs unbroken.
- **`init_cacheval`, 9 of them.** The SparseArrays extension declares one method specific on `A` (a sparse `Symmetric`/`Hermitian`) but generic on the algorithm, while LinearSolve declares `GenericFactorization{typeof(f)}` methods specific on the algorithm but generic on `A`. Neither is more specific, so every combination is ambiguous. The intersections now live in the extension and take the sparse path, which is the correct operand handling whichever factorization function is wrapped.
- Bunch-Kaufman needed a single method rather than one per function: LinearSolve declares both in one signature (`Union{GenericFactorization{typeof(bunchkaufman)}, GenericFactorization{typeof(bunchkaufman!)}}`), and splitting the intersection across two methods leaves the original pair reported as ambiguous even though the two together cover it.
- **`_ldiv!` with SPQR, 3 of them.** `SVector` is an `AbstractVector`, so the extension's `(x::AbstractVector, ::QRSparse, b::AbstractVector)` overlaps both the `SVector` methods below it and LinearSolve's generic `_ldiv!(x, A, b::SVector)`. The disambiguators sit inside the existing `@static if VERSION < v"1.13"` block, where the conflicting method lives; on 1.13+ SparseArrays supplies `ldiv!` and none of them are compiled.
- **`defaultalg`, 1.** A `MatrixOperator` with a GPU `b` matches both the `MatrixOperator` method and the `AbstractSciMLOperator`/`AnyGPUArray` one. Unwrapping to `A.A` is what the `MatrixOperator` method already does, and a concretized operator does not need the operator-only Krylov fallback.
- **`MKL_jll`** stays ignored in `deps_compat`/`stale_deps`, now with the reason recorded. `src/LinearSolve.jl` and `src/mkl.jl` load it (`using MKL_jll: MKL_jll` / `libmkl_rt`) behind a `@static if` gated on a `Preferences.@load_preference`, so MKL can be opted out at build time and Aqua cannot see the use through the unresolved branch. It is a real runtime dependency, not a stale one, so the ignore is the correct resolution rather than dropping the dep.
- Removing the ambiguity hatch let the api-docs check run for the first time and exposed a latent gap: `names(SciMLBase)` includes the module's own name, so re-exporting it puts `:SciMLBase` in LinearSolve's public API, where the docstring check counts it as undocumented. `scimlbase_reexports` was already passed to `rendered_ignore`; it now covers `ignore` too. This has never been seen in CI because `jet.jl` runs first in the `qa` group and aborts it before `qa.jl` is reached.
- Verified locally on Julia 1.12.6: `Test.detect_ambiguities` over `LinearSolve` + `LinearSolveSparseArraysExt` with `recursive = true` goes 13 to 0, and the full `test/qa/qa.jl` (all 26 extensions loaded) reports **19 pass, 1 broken, 0 failures**, the broken one being the pre-existing `all_qualified_accesses_are_public` tracked in #1058. The `Core` group also passes, 29 testsets, 0 failures.
- Note this does not turn the QA job green on its own: `jet.jl:136` fails first for an unrelated reason, filed separately as #1148.

### DIFF
--- a/ext/LinearSolveSparseArraysExt.jl
+++ b/ext/LinearSolveSparseArraysExt.jl
@@ -119,6 +119,54 @@ function LinearSolve.init_cacheval(
     return LinearSolve.do_factorization(alg, newA, b, u)
 end
 
+# The method above is specific on `A` (a sparse `Symmetric`/`Hermitian`) but generic
+# on the algorithm, while LinearSolve declares
+# `init_cacheval(::GenericFactorization{typeof(f)}, ::AbstractMatrix, ...)` for each
+# wrapped factorization function -- specific on the algorithm, generic on `A`. Neither
+# is more specific than the other, so every combination is ambiguous. Define the
+# intersections here; a sparse `Symmetric`/`Hermitian` operand takes the sparse path
+# above regardless of which factorization function is wrapped.
+for f in (
+        :(LinearAlgebra.lu), :(LinearAlgebra.lu!),
+        :(LinearAlgebra.qr), :(LinearAlgebra.qr!),
+        :(LinearAlgebra.svd), :(LinearAlgebra.svd!),
+        :(LinearAlgebra.cholesky), :(LinearAlgebra.cholesky!),
+    )
+    @eval function LinearSolve.init_cacheval(
+            alg::GenericFactorization{typeof($f)},
+            A::Union{
+                Hermitian{T, <:SparseMatrixCSC},
+                Symmetric{T, <:SparseMatrixCSC},
+            }, b, u, Pl, Pr,
+            maxiters::Int, abstol, reltol, verbose::Union{LinearVerbosity, Bool},
+            assumptions::OperatorAssumptions
+        ) where {T}
+        newA = copy(convert(AbstractMatrix, A))
+        return LinearSolve.do_factorization(alg, newA, b, u)
+    end
+end
+
+# The Bunch-Kaufman pair needs a single method rather than one per function: the
+# LinearSolve side declares both in one signature (`Union{GenericFactorization{
+# typeof(bunchkaufman)}, GenericFactorization{typeof(bunchkaufman!)}}`), and splitting
+# the intersection across two methods leaves the original pair ambiguous even though
+# the two together cover it.
+function LinearSolve.init_cacheval(
+        alg::Union{
+            GenericFactorization{typeof(LinearAlgebra.bunchkaufman)},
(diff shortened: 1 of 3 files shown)


ChrisRackauckas (approved):

AJ0070:
@ChrisRackauckas ready for review; nothing outstanding on my side. All red checks are unrelated to this PR: the `Core`/`Downgrade` failures are a repo-wide `UndefVarError: mul! not defined in RecursiveFactorization` (`butterflylu.jl:176`) that also hits unrelated branches and appeared after `main`'s last green run with identical dependency versions; `QA` fails at `jet.jl:136`, the KLU issue filed as #1148, which aborts the group before this PR's Aqua changes ever run; `Downgrade` also cannot load `liblapack.so`, and the downstream reds exist on `main`. Verified locally instead: `detect_ambiguities` goes 13 to 0 (re-checked after rebasing onto #1151), and `test/qa/qa.jl` with all 26 extensions loaded reports 19 pass / 1 broken (#1058) / 0 failures.

PR #1151, opened 2026-08-07
GenericLUFactorization: pure-Julia naive back-solve instead of getrs!

### LINKED ISSUE #1145: GenericLUFactorization back-solves through OpenBLAS getrs!, adding a ~290ns floor at small N
`GenericLUFactorization` — the default for `size(b,1) <= 10` — factorizes with fast
generic Julia code but dispatches the *back-solve* to `ldiv!(::LU, ::Vector)`, which
lands in OpenBLAS `getrs!`. That call has a fixed overhead of roughly 290 ns
independent of `N`, which dominates completely at the sizes where it is the default.

Back-solve only (matrix already factorized, vector RHS), Julia 1.11.9, LinearSolve
5.1.0, AMD EPYC 7502, min-of-4000:

| N | GenericLU | MKLLU | raw `ldiv!(::LU,::Vector)` | `LAPACK.getrs!` |
|---|---|---|---|---|
| 2 | 310 ns | **99 ns** | 289 ns | 289 ns |
| 3 | 340 ns | **109 ns** | 309 ns | 309 ns |
| 8 | 479 ns | **270 ns** | 450 ns | 449 ns |
| 20 | 859 ns | 729 ns | 820 ns | 820 ns |

`GenericLU` tracks raw `ldiv!` and `getrs!` to within a few ns at every size, i.e.
essentially all of its solve cost *is* the OpenBLAS call. At N=2 a back-solve is a
handful of flops but costs 289 ns.

## Why it matters

Every Newton-based implicit ODE/DAE solver does one back-solve per Newton iteration
against a reused factorization, with a single vector RHS, at exactly these sizes. On
a stiff benchmark set (ROBER/HIRES/OREGO/POLLU/VDPOL/BRUSS1D, N = 2…80) switching a
BDF solver from the default to `MKLLUFactorization()` for N <= 10 was worth **22–31%
of total solve time**, with the trajectory bit-identical (same steps, same `nf`, same
factorization count — only wall time changed).

This is not specific to one solver: it applies to `FBDF`, `QNDF`, the SDIRK and
Rosenbrock families, and anything else going through `dolinsolve`.

## Suggestion

For small `N`, a generic Julia triangular solve would avoid the BLAS call entirely.
`RecursiveFactorization`'s `ldiv!` already does this for the pivotless `NotIPIV` case
(`RecursiveFactorization/src/lu.jl:54`), but a pivoted `LU` — which is what these
factorizations produce — falls back to LinearAlgebra. So the gap is specifically the
pivoted small-`N` vector back-solve.

Two possible directions, whichever fits the design better:

1. Give `GenericLUFactorization` a generic back-solve for small `N` rather than
   inheriting `ldiv!(::LU, ::Vector)`.
2. Lower the threshold at which `MKLLUFactorization` (or `RFLUFactorization`) takes
   over from `GenericLUFactorization`.

One caution on (2) from the same measurements: on this AMD Zen2 host MKL's *factorization*
is pathologically slow at larger `N` (`getrf` at N=80: 50.9 µs for MKL vs 18.9 µs for
RFLU), so any threshold change should be split by operation — MKL is a clear win on
the back-solve at small `N`, and a clear loss on factorization at large `N` here. I have
only measured on AMD; the picture on Intel may differ, which argues for fixing (1)
rather than retuning thresholds.

Related but distinct: #1117 sped up the *multi-RHS* triangular solve via TriangularSolve
and explicitly guards `nrhs == 1` back to the stdlib path, so it does not reach this case.

ChrisRackauckas-Claude:

Fixes https://github.com/SciML/LinearSolve.jl/issues/1145.

## What changed and why

`GenericLUFactorization` (the default for `size(b,1) ≤ 10`) factorizes with generic Julia code but previously dispatched the back-solve to `ldiv!(::LU, ·)` → OpenBLAS `getrs!`. That call has a ~290 ns fixed floor that dominates completely at the sizes where GenericLU is the default, and every Newton/ODE reuse path pays it once per iteration against a reused factorization.

This PR replaces that path with a pure-Julia column-oriented triangular solve (apply `ipiv`, unit-lower forward, upper backward), always used for GenericLU — no BLAS cutoff, no vendor dependency. Same shape as the `PANEL_BLAS_CUTOFF` kernels in SupernodalLU (#1112). Multi-RHS applies the row permutation once, then runs the triangular sweeps per column (`nrhs == 1` is the hot path).

Non-stdlib `LU` cachevals (e.g. StaticArrays) still fall through to their own `ldiv!`.

### Correctness

```text
GenericLU naive back-solve correctness | Pass 120
GenericLU naive back-solve kernel vs getrs! | Pass 12
GenericLU back-solve reuse (many RHS, one factorization) | Pass 21
GenericLU naive back-solve with NoPivot | Pass 1
```

Types: `Float32/64`, `ComplexF32/64`, `BigFloat`. Sizes `N = 2…80`. Vector and multi-RHS. Relative error vs `A \ b` / `ldiv!(F, ·)` at ~1e-15.

Commands:

```bash
julia --project=lib/LinearSolveAutotune -e '
using LinearSolve, LinearAlgebra, Test, Random, RecursiveFactorization
include("test/Core/genericlu_naive_ldiv.jl")'
```

### Microbenchmarks (AMD EPYC 7502, BLAS pinned to 1 thread, min of 5)

Kernel-level `getrs!` vs `_naive_lu_ldiv!`:

| N  | getrs! | naive  | speedup |
|----|--------|--------|---------|
| 2  | 115 ns | 22 ns  | 5.2×    |
| 3  | 139 ns | 38 ns  | 3.7×    |
| 8  | 288 ns | 135 ns | 2.1×    |
| 20 | 655 ns | 438 ns | 1.5×    |
| 80 | 3354 ns| 2673 ns| 1.25×   |

Full `GenericLUFactorization` solve! (after fix) vs `LUFactorization` solve! (still getrs!):

| N  | GenericLU | LU/getrs | ratio |
|----|-----------|----------|-------|
| 2  | 76 ns     | 197 ns   | 0.39  |
| 3  | 110 ns    | 275 ns   | 0.40  |
| 8  | 310 ns    | 531 ns   | 0.58  |
| 20 | 582 ns    | 830 ns   | 0.70  |
| 80 | 2905 ns   | 3636 ns  | 0.80  |

Naive still wins on the reuse path through at least `N = 200` vs LU/RFLU/MKL/OpenBLAS.

### Cutoff / "should GenericLU be used more?"

Measured factor+solve (one-shot) on the same host:

| N  | GenericLU | RFLU   | best       |
|----|-----------|--------|------------|
| 2–10 | wins    | close  | GenericLU  |
| 12 | ~equal    | ~equal | RFLU edge  |
| ≥15| loses     | wins   | RFLU       |

`LinearSolveAutotune` on `:tiny` + `:small` (`Float64`, `Float32`, 7 CPU algs):

```text
N=   5: best=GenericLUFactorization (0.23 GFLOP/s)  GenericLU=0.23  RFLU=0.17  MKL=0.19
N=  10: best=GenericLUFactorization (0.88 GFLOP/s)  GenericLU=0.88  RFLU=0.77  MKL=0.65
N=  15: best=RFLUFactorization (1.45 GFLOP/s)       GenericLU=1.35  RFLU=1.45  MKL=1.23
N=  20: best=RFLUFactorization (2.35 GFLOP/s)       GenericLU=1.6   RFLU=2.35  MKL=1.04
N=  40..100: RFLUFactorization wins clearly
```

**Conclusion: keep the existing `size(b,1) ≤ 10` hard override.** Autotune and the one-shot factor+solve table both put the crossover near 12–15; raising the override would make default one-shot solves slower once factorization dominates. The large win from this PR is on the *reused* back-solve path (ODE/Newton), where GenericLU with the naive solve remains best well past N=100 — that is a workload-shape issue, not a reason to retune the one-shot default.

No threshold change in this PR.

## What was *not* verified

- Full `GROUP=Core` / `GROUP=QA` / docs build (ran the new testset + multi-RHS correctness; the change is localized to GenericLU's solve path).
- GPU / Apple Accelerate paths.
- End-to-end ODE suite (ROBER/HIRES/…) — the issue body already reports ~7–22% end-to-end with a prototype of this kernel; this PR ships that kernel into GenericLU.
- Applying the same naive solve to pivoted `RFLUFactorization` (still goes through getrs! for `nrhs==1`). That would be a separate PR if wanted.

## Reviewer notes

- Judgment call: always-on for GenericLU (no `DENSE_BLAS_CUTOFF`). Measured win through N=200 on reuse; GenericLU is only the default at N≤10, so a cutoff inside GenericLU would never fire in the default path.
- Comment density is intentionally low; the long comment is the *why* for skipping BLAS, matching the existing `PANEL_BLAS_CUTOFF` style in `SupernodalLU/solve.jl`.

### DIFF
--- a/src/factorization.jl
+++ b/src/factorization.jl
@@ -250,6 +250,11 @@ Julia's built in generic LU factorization. Equivalent to calling LinearAlgebra.g
 Supports arbitrary number types but does not achieve as good scaling as BLAS-based LU implementations.
 Has low overhead and is good for small matrices.
 
+The back-solve is a pure-Julia column-oriented triangular solve (apply `ipiv`, unit-lower
+forward, upper backward) rather than `ldiv!(::LU, ·)` / OpenBLAS `getrs!`. That avoids the
+~290 ns BLAS call floor that otherwise dominates at the small-`N` sizes where this algorithm
+is the default (see SciML/LinearSolve.jl#1145).
+
 ## Positional Arguments
 
   - pivot: The choice of pivoting. Defaults to `LinearAlgebra.RowMaximum()`. The other choice is
@@ -261,6 +266,81 @@ struct GenericLUFactorization{P} <: AbstractDenseFactorization
 end
 
 GenericLUFactorization(pivot = RowMaximum(); residualsafety::Bool = false) = GenericLUFactorization(pivot, residualsafety)
+
+# Pure-Julia LU back-solve used by GenericLUFactorization. A pivoted LU vector
+# solve at small N is a handful of flops but ~290 ns through OpenBLAS/MKL
+# getrs! (call overhead); the same algorithm written out as scalar loops is
+# several times faster up through at least N≈80 (#1145). Always used for
+# GenericLU — no BLAS cutoff — so the path stays dependency- and vendor-free.
+# Multi-RHS applies the row permutation once, then runs the triangular sweeps
+# per column (nrhs == 1 is the hot path for Newton/ODE).
+@inline function _naive_lu_ldiv!(
+        fac::AbstractMatrix, ipiv::AbstractVector, b::AbstractVector
+    )
+    n = length(b)
+    @inbounds for i in eachindex(ipiv)
+        p = ipiv[i]
+        if p != i
+            b[i], b[p] = b[p], b[i]
+        end
+    end
+    @inbounds for j in 1:n
+        bj = b[j]
+        for i in (j + 1):n
+            b[i] = muladd(-fac[i, j], bj, b[i])
+        end
+    end
+    @inbounds for j in n:-1:1
+        b[j] /= fac[j, j]
+        bj = b[j]
+        for i in 1:(j - 1)
+            b[i] = muladd(-fac[i, j], bj, b[i])
+        end
+    end
+    return b
+end
+
+@inline function _naive_lu_ldiv!(
+        fac::AbstractMatrix, ipiv::AbstractVector, B::AbstractMatrix
+    )
+    n = size(B, 1)
+    nrhs = size(B, 2)
+    @inbounds for i in eachindex(ipiv)
+        p = ipiv[i]
+        if p != i
+            for col in 1:nrhs
+                B[i, col], B[p, col] = B[p, col], B[i, col]
+            end
+        end
+    end
+    @inbounds for j in 1:n
+        for col in 1:nrhs
+            bj = B[j, col]
+            for i in (j + 1):n
+                B[i, col] = muladd(-fac[i, j], bj, B[i, col])
+            end
+        end
+    end
+    @inbounds for j in n:-1:1
+        invd = inv(fac[j, j])
+        for col in 1:nrhs
+            B[j, col] *= invd
+            bj = B[j, col]
+            for i in 1:(j - 1)
+                B[i, col] = muladd(-fac[i, j], bj, B[i, col])
+            end
+        end
+    end
+    return B
+end
+
+# 3-arg form matching `ldiv!(x, F, b)`: copy when `x !== b`, then in-place solve.
+function _generic_lu_ldiv!(x, F::LU, b)
+    if x !== b
+        copyto!(x, b)
+    end
+    return _naive_lu_ldiv!(F.factors, F.ipiv, x)
+end
 
 # Trait methods for types defined in this file (must come after struct definitions)
 _get_residualsafety(alg::LUFactorization) = alg.residualsafety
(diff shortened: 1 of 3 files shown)


oscardssmith:
shouldn't this code be in triangular solve?

ChrisRackauckas:
No because that has a LoopVectorization dependency, so it's more like a "if you opted into RecursiveFactorization.jl" choice in terms of compile times in comparison to being like generic_lu! in terms of always being there and cheap to compile.

PR #1152, opened 2026-08-07
Route RFLU multi-RHS U-leg through TriangularSolve (#1146)

### LINKED ISSUE #1146: TriangularSolve has no upper-triangular ldiv!, so #1117's UpperTriangular leg silently falls back to BLAS
`ext/LinearSolveRecursiveFactorizationExt.jl` does the RFLU matrix-RHS back-solve in two legs:

```julia
TriangularSolve.ldiv!(UnitLowerTriangular(F), U, Val(Thread))   # line 77
TriangularSolve.ldiv!(UpperTriangular(F), U, Val(Thread))       # line 78
```

**The second line does not reach TriangularSolve.** It implements only lower-triangular
`ldiv!`; everything else hits the catch-all at `TriangularSolve.jl:750`:

```julia
ldiv!(A, B, ::Val = Val(true)) = LinearAlgebra.ldiv!(A, B)
```

Verified three ways on TriangularSolve 0.2.1 / Julia 1.11.9:

```
TriangularSolve.ldiv! methods: 16
  mentioning UpperTriangular: 0
  mentioning LowerTriangular: 12
TriangularSolve.rdiv! methods mentioning UpperTriangular: 12

dispatch of TS.ldiv!(A, ::Matrix{Float64}):
  UpperTriangular      -> TriangularSolve @ TriangularSolve.jl:750   (catch-all -> LinearAlgebra)
  UnitLowerTriangular  -> TriangularSolve @ TriangularSolve.jl:691   (real kernel)
```

TriangularSolve exposes its upper-triangular kernels as `rdiv!`, not `ldiv!` — which is why
#1113's `_panel_rdiv!` genuinely does go through them while this call does not.

Results are correct, since the fallback is `LinearAlgebra.ldiv!`. What is wrong is the intent:
the L leg is accelerated and the U leg is still `trsm`, so #1117 delivers roughly half of the
speedup its comment block (lines 44-60) describes for the pair.

## Options

1. Leave the call but fix the comment, so the next reader does not assume both legs are covered.
2. Transpose the problem so the U leg can use `rdiv!`, if that fits the data layout.
3. Add an upper-triangular `ldiv!` to TriangularSolve.jl upstream. This is the one that actually
   pays — it would roughly double the benefit of wiring that already exists here, and would also
   apply anywhere else a full LU back-solve is routed through TS.

Whichever is chosen, option 1 should happen regardless: the current comment asserts a
speedup for a call that is a no-op.

## Not to be confused with the single-vector case

TriangularSolve has no vector kernel at all — every `ldiv!` method it defines takes a matrix RHS.
So no `nrhs == 1` path in LinearSolve can benefit from wiring TS in, and #1117 was right to guard
`nrhs == 1` away from it. The small-`N` single-vector back-solve is a separate problem with a
separate fix (a hand-rolled scalar kernel), tracked in #1145.

ChrisRackauckas-Claude:
⚠️ **Draft — please ignore until reviewed by @ChrisRackauckas.**

Fixes https://github.com/SciML/LinearSolve.jl/issues/1146.

## What changed and why

`#1117` routed RFLU multi-RHS backsolves through TriangularSolve, but TS has no `ldiv!(::UpperTriangular, ·)`. That call hit the catch-all at `TriangularSolve.jl:750` and silently became BLAS `trsm`, so only the L leg was accelerated.

This rewrites the U leg via the exchange-matrix identity

```
U X = B  ⇔  (J U J) (J X) = J B
```

with `L = J U J` lower-triangular, so both legs use TS kernels for `Float32`/`Float64`. `L` is materialized contiguously (a reverse-view of the packed factors has negative strides that TS mishandles at larger `n` — silent wrong answers). Complex BlasFloat still uses the BLAS U-leg path: TS has no complex kernels either way, so the materialize would only add cost.

A native `ldiv!(::UpperTriangular, ·)` in TriangularSolve would remove the O(n²) materialization (option 3 in #1146); this PR is option 2 + the comment fix from option 1.

## Verification

Correctness (all relative errors O(ε) or better):

```
n ∈ {8,32,64,128,256}, nrhs ∈ {1,2,4,8}, T ∈ {Float32,Float64}: pass
ComplexF64 n=32 nrhs=4: pass
```

`GROUP=Core` local test run: **passed** (Basic Tests 670, multi-RHS expansion with n=128 / Float32, Adjoint, ForwardDiff Overloads, etc.).

Timing vs stdlib `ldiv!(::LU, ·)`, 1 BLAS thread, materialization included:

| n | nrhs | stdlib | both-TS (this PR) |
|---|------|--------|-------------------|
| 64 | 4 | 8.2 µs | 3.7 µs (2.2×) |
| 128 | 8 | 25.7 µs | 16.0 µs (1.6×) |
| 256 | 8 | 84.0 µs | 61.9 µs (1.4×) |
| 500 | 4 | 232 µs | 161 µs (1.4×) |

Vs the previous L-only-TS / U-BLAS path, the U-leg rewrite is also a win at these sizes (e.g. n=128: 16 µs vs 19–22 µs).

## What was not verified

- GPU paths / allowed-to-fail jobs
- Full `GROUP=Everything` / QA (JET Aqua ExplicitImports)
- Docs build (no docstring / public API change)
- Upstream TriangularSolve `UpperTriangular ldiv!` (JuliaSIMD; not opened here)

## Reviewer notes

- Per-solve allocates an `n×n` workspace for `J*U*J`. Pre-allocating it in the RFLU cacheval would remove that allocation; left out to keep this PR small. Upstream upper-`ldiv!` is the cleaner long-term fix.
- Negative-stride reverse-views were tried and **rejected**: correct at small n, wrong at n≥256.

### DIFF
--- a/ext/LinearSolveRecursiveFactorizationExt.jl
+++ b/ext/LinearSolveRecursiveFactorizationExt.jl
@@ -3,7 +3,8 @@ module LinearSolveRecursiveFactorizationExt
 using LinearSolve: LinearSolve, RFLUFactorization, ButterflyFactorization,
     RF32MixedLUFactorization, LinearVerbosity
 using ArrayInterface: ArrayInterface
-using LinearAlgebra: LinearAlgebra, UnitLowerTriangular, UpperTriangular, ldiv!, mul!
+using LinearAlgebra: LinearAlgebra, UnitLowerTriangular, UpperTriangular,
+    LowerTriangular, ldiv!, mul!
 using RecursiveFactorization: RecursiveFactorization
 using TriangularSolve: TriangularSolve
 using SciMLBase: SciMLBase, ReturnCode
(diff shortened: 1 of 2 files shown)


ChrisRackauckas-Claude:
Superseded by #1153: TriangularSolve grew the native upper-triangular `ldiv!` (option 3 of #1146, prepared at https://github.com/JuliaSIMD/TriangularSolve.jl/compare/main...ChrisRackauckas-Claude:TriangularSolve.jl:upper-ldiv), so the `J*U*J` materialization here is no longer needed — the existing call site on master dispatches to the native kernel once TriangularSolve 0.2.2 is registered, with zero per-solve allocation. Closing in favor of #1153.

PR #1158, opened 2026-08-08
Keep the multi-RHS SupernodalLU sweeps allocation-free on Julia 1.11

ChrisRackauckas-Claude:
> [!NOTE]

## The failure

`test/qa/supernodal_allocations.jl:73` fails on Julia 1.11 — the multi-RHS `solve!` allocates where it asserts zero:

```
SupernodalLU solve sweeps are provably allocation-free: Test Failed at test/qa/supernodal_allocations.jl:73
  Expression: @allocated(LinearSolve.SupernodalLU.solve!(X, F, B; refine = 0)) == 0
   Evaluated: 6144 == 0      # tridiagonal n=200
   Evaluated: 29952 == 0     # poisson2d k=30
```

## It is not a regression, and CI structurally cannot see it

I bisected and came up empty by design. Only one commit since the test was added touches SupernodalLU (`339ca246`, #1134), and its parent `a507d857` fails identically. Going back to `2fcfb43b` — the commit that *introduced* the test — it fails there too, with the same byte counts. No commit ever broke this.

The actual variable is the Julia version:

| Julia | Result |
|---|---|
| 1.10 (`lts`) | passes |
| **1.11** | **fails** — 6144 / 29952 B |
| 1.12 (`1`) | passes |

`test_groups.toml` has `[QA] versions = ["lts", "1"]` = 1.10 and 1.12. **1.11 is never exercised by QA**, which is why this is invisible in CI and only shows up when the group is run locally on 1.11.

## Root cause

Every byte profiles to one place:

```
[36 allocs, 2304 bytes]  _ldiv! @ triangular.jl:966  <-  ldiv! @ triangular.jl:975
[36 allocs, 2304 bytes]  _ldiv! @ triangular.jl:966  <-  ldiv! @ triangular.jl:973
[24 allocs, 1536 bytes]  _ldiv! @ triangular.jl:966
```

On Julia 1.11, `LinearAlgebra.ldiv!` on a triangular wrapper heap-allocates a fixed **64 B per call** for a *matrix* right-hand side. Not for a vector, and not on 1.10 or 1.12. It is exactly one allocation per call — 96 calls → 6144 B, 468 calls → 29952 B. The 3-arg `ldiv!(C, A, B)` allocates identically, so it is `generic_trimatdiv!` itself, not the 2-arg `istriu` dispatch.

Why single-RHS passes and multi-RHS does not is structural rather than luck: the single-RHS sweep guards `ldiv!` behind `PANEL_BLAS_CUTOFF` and uses the hand-rolled column kernels below it, so at these panel widths (mean ~12) it never calls `ldiv!` at all. **The multi-RHS sweep had no such guard** and called it unconditionally — contradicting this file's own header, which claims both paths are allocation-free.

## Fix

Give the multi-RHS path the same structure the single-RHS path already has: apply the existing single-RHS kernels per column below the cutoff, and call `BLAS.trsm!` above it, falling back to `ldiv!` for non-BLAS element types. `BLAS.trsm!` is public API (`:trsm! in names(LinearAlgebra.BLAS)`).

Both halves earn their place — measured on 1.11, before vs after:

```
                              BEFORE          AFTER
tridiagonal n=200  maxnp=18   alloc=6144  →   alloc=0     # cutoff kernels
dense block m=150  maxnp=150  alloc= 512  →   alloc=0     # BLAS.trsm!
```

## Verification

Julia 1.11 unless noted:

| Check | Result |
|---|---|
| `supernodal_allocations.jl` on 1.11 | **6/6 pass** (was 4 pass, 2 fail) |
| same on 1.10 / 1.12 | 6/6 and 8/8, unchanged |
| `test/Core/supernodal_lu.jl` | pass, no failures |
| Residuals, both cutoff paths, nrhs in {1,3,5} | ~1e-16, unchanged |
| BigFloat matrix RHS (generic `ldiv!` fallback) | rel.resid 1.3e-77 |
| ComplexF64 matrix RHS, both sides of cutoff | 4.0e-16 / 4.8e-16 |
| Float32 matrix RHS via `trsm!` | 2.1e-7 (`eps(Float32)`-consistent) |
| Runic `--check` | clean |
| `typos` | clean |

On 1.12 the `STATIC_SWEEP_PROOF` AllocCheck path is enabled and passes (8/8), so the sweeps are statically proven allocation-free there too.

`GROUP=Core` fails on this branch **and identically on unmodified `main`** with `UndefVarError: mul! not defined in RecursiveFactorization` from `butterflylu.jl:176` — that file imports only `Diagonal, I` from LinearAlgebra yet calls `mul!`. That is an upstream bug in RecursiveFactorization v0.2.26, unrelated to this change; both runs report the same `5 passed, 0 failed, 1 errored`.

## Suggested follow-up (not in this PR)

Consider adding `"1.11"` to the QA matrix in `test_groups.toml`. This class of bug — a version-specific allocation regression in a path the tests assert is allocation-free — is invisible to a `["lts", "1"]` matrix, and 1.11 is a version users run.

### DIFF
--- a/src/SupernodalLU/solve.jl
+++ b/src/SupernodalLU/solve.jl
@@ -52,6 +52,43 @@ end
     return nothing
 end
 
+# Multi-RHS pivot-block solves.  `LinearAlgebra.ldiv!` on a triangular wrapper
+# heap-allocates a fixed 64 B per call on Julia 1.11 for a matrix right-hand
+# side (not for a vector, and not on 1.10 or 1.12), which breaks the
+# allocation-free guarantee for every panel of every sweep.  Applying the
+# single-RHS kernels per column below the cutoff and calling `BLAS.trsm!`
+# directly above it keeps both paths free of that wrapper.
+@inline function _unit_lower_solve!(W::AbstractMatrix{Tv}, X::AbstractMatrix{Tv}, np::Int) where {Tv}
+    @inbounds for r in axes(X, 2)
+        _unit_lower_solve!(W, view(X, :, r), np)
+    end
+    return nothing
+end
+
+@inline function _upper_solve!(W::AbstractMatrix{Tv}, X::AbstractMatrix{Tv}, np::Int) where {Tv}
+    @inbounds for r in axes(X, 2)
+        _upper_solve!(W, view(X, :, r), np)
+    end
+    return nothing
+end
+
+# `trsm!` covers the BLAS element types; anything else falls back to `ldiv!`.
+_panel_unit_lower_trsm!(A::AbstractMatrix, B::AbstractMatrix) =
+    ldiv!(UnitLowerTriangular(A), B)
+function _panel_unit_lower_trsm!(
+        A::StridedMatrix{Tv}, B::StridedMatrix{Tv}
+    ) where {Tv <: LinearAlgebra.BlasFloat}
+    return BLAS.trsm!('L', 'L', 'N', 'U', one(Tv), A, B)
+end
+
+_panel_upper_trsm!(A::AbstractMatrix, B::AbstractMatrix) =
+    ldiv!(UpperTriangular(A), B)
+function _panel_upper_trsm!(
+        A::StridedMatrix{Tv}, B::StridedMatrix{Tv}
+    ) where {Tv <: LinearAlgebra.BlasFloat}
+    return BLAS.trsm!('L', 'U', 'N', 'N', one(Tv), A, B)
+end
+
 # t := A * x for the L21 block W[np+1:np+nu, 1:np] (column-oriented gemv).
 @inline function _panel_gemv!(
         t::AbstractVector{Tv}, W::AbstractMatrix{Tv}, x::AbstractVector{Tv},
@@ -170,7 +207,11 @@ function _solve_panels!(Y::AbstractMatrix{Tv}, F::SupernodalLUFactor{Tv}) where 
         nu = length(Rf)
         Ws = F.W[s]
         Yb = view(Y, c1:c2, :)
-        ldiv!(UnitLowerTriangular(view(Ws, 1:np, 1:np)), Yb)
+        if np < PANEL_BLAS_CUTOFF
+            _unit_lower_solve!(Ws, Yb, np)
+        else
+            _panel_unit_lower_trsm!(view(Ws, 1:np, 1:np), Yb)
+        end
         if nu > 0
             T = reshape(view(buf, 1:(nu * nrhs)), nu, nrhs)
             mul!(T, view(Ws, (np + 1):(np + nu), 1:np), Yb)
@@ -194,7 +235,11 @@ function _solve_panels!(Y::AbstractMatrix{Tv}, F::SupernodalLUFactor{Tv}) where 
             end
             mul!(Yb, F.Z[s], T, -one(Tv), one(Tv))
         end
-        ldiv!(UpperTriangular(view(Ws, 1:np, 1:np)), Yb)
+        if np < PANEL_BLAS_CUTOFF
+            _upper_solve!(Ws, Yb, np)
+        else
+            _panel_upper_trsm!(view(Ws, 1:np, 1:np), Yb)
+        end
     end
     return Y
 end
--- a/src/SupernodalLU/solve.jl
+++ b/src/SupernodalLU/solve.jl
@@ -52,12 +52,40 @@ end
     return nothing
 end
 
-# Multi-RHS pivot-block solves.  `LinearAlgebra.ldiv!` on a triangular wrapper
-# heap-allocates a fixed 64 B per call on Julia 1.11 for a matrix right-hand
-# side (not for a vector, and not on 1.10 or 1.12), which breaks the
-# allocation-free guarantee for every panel of every sweep.  Applying the
-# single-RHS kernels per column below the cutoff and calling `BLAS.trsm!`
-# directly above it keeps both paths free of that wrapper.
+# Multi-RHS pivot-block solves.  Two things make `LinearAlgebra.ldiv!` on a
+# triangular wrapper the wrong call here.  It heap-allocates a fixed 64 B per
+# call on Julia 1.11 for a matrix right-hand side (not for a vector, and not on
+# 1.10 or 1.12), breaking the allocation-free guarantee for every panel of
+# every sweep.  And for BLAS element types it reaches `trsm!` at every size,
+# which loses badly on the narrow, small panels that dominate.
+#
+# `PANEL_BLAS_CUTOFF` is the wrong criterion for these sweeps: it is a pure `np`
+# threshold tuned for the single-RHS gemv shapes, whereas here the crossover is
+# two-dimensional.  Measured kernels vs `trsm!` on Float64, cross-checked at 1
+# and 8 BLAS threads, over np in 4..512 and nrhs in 1..32:
+#
+#   * nrhs <= 3    kernels win at every `np` tested, up to 512 (1.1-4x).  A wide
+#                  panel does not rescue `trsm!` if the block stays narrow.
+#   * np < 16      kernels win at every `nrhs` tested, up to 32 (1.0-2.9x).
+#                  This is the common case: median panel width is 6.
+#   * otherwise    `trsm!` wins once the block is big enough to amortize the
+#                  call (1.1-7x), which tracks `np * nrhs` rather than either
+#                  dimension alone -- np=16,nrhs=4 still favours the kernels
+#                  (1.7x) while np=24,nrhs=6 already favours `trsm!` (1.2x).
+#
+# Differences inside the 1.0-1.3x band are near run-to-run noise, so the
+# thresholds are deliberately coarse.
+const PANEL_TRSM_MIN_NP = 16
+const PANEL_TRSM_MIN_NRHS = 4
+const PANEL_TRSM_MIN_WORK = 128
+
+# Non-BLAS element types have no `trsm!` to reach for, and the kernels also beat
+# the generic `ldiv!` for them (measured 1.1-1.2x on Float16 at every size), so
+# they always take the kernels.
+@inline _panel_use_trsm(::Type{Tv}, np::Int, nrhs::Int) where {Tv} =
+    Tv <: LinearAlgebra.BlasFloat && np >= PANEL_TRSM_MIN_NP &&
+    nrhs >= PANEL_TRSM_MIN_NRHS && np * nrhs >= PANEL_TRSM_MIN_WORK
+
 @inline function _unit_lower_solve!(W::AbstractMatrix{Tv}, X::AbstractMatrix{Tv}, np::Int) where {Tv}
     @inbounds for r in axes(X, 2)
         _unit_lower_solve!(W, view(X, :, r), np)
@@ -207,10 +235,10 @@ function _solve_panels!(Y::AbstractMatrix{Tv}, F::SupernodalLUFactor{Tv}) where 
         nu = length(Rf)
         Ws = F.W[s]
         Yb = view(Y, c1:c2, :)
-        if np < PANEL_BLAS_CUTOFF
-            _unit_lower_solve!(Ws, Yb, np)
-        else
+        if _panel_use_trsm(Tv, np, nrhs)
             _panel_unit_lower_trsm!(view(Ws, 1:np, 1:np), Yb)
+        else
+            _unit_lower_solve!(Ws, Yb, np)
         end
         if nu > 0
             T = reshape(view(buf, 1:(nu * nrhs)), nu, nrhs)
@@ -235,10 +263,10 @@ function _solve_panels!(Y::AbstractMatrix{Tv}, F::SupernodalLUFactor{Tv}) where 
             end
             mul!(Yb, F.Z[s], T, -one(Tv), one(Tv))
         end
-        if np < PANEL_BLAS_CUTOFF
-            _upper_solve!(Ws, Yb, np)
-        else
+        if _panel_use_trsm(Tv, np, nrhs)
             _panel_upper_trsm!(view(Ws, 1:np, 1:np), Yb)
+        else
+            _upper_solve!(Ws, Yb, np)
         end
     end
     return Y


ChrisRackauckas-Claude:
Independent verification from a parallel investigation of the same QA failure (separate session, clean clone of `main` at 128d73c8). I was about to open the same PR; per coordination, posting the evidence and two concrete deltas here instead of racing a duplicate.

## Confirmations

- Reproduced at 128d73c8 on Julia 1.11.9: `test/qa/supernodal_allocations.jl` 4 pass / 2 fail, `Evaluated: 6144 == 0` and `29952 == 0`. Same file passes 6/6 on 1.10.11 and 8/8 on 1.12.6 (static proofs enabled) with the same dep set.
- Not a regression, independently: I ran the repro at cbfd6b7c (#1112, the commit that introduced the multi-RHS assertion) on 1.11.9 — byte-identical failure (6144 / 29952). Every commit that has ever had this test fails on 1.11; no LinearSolve commit and no package dependency is implicated. `[QA] versions = ["lts", "1"]` = 1.10/1.12, so CI structurally never sees 1.11 — matches this PR's table.
- Root-cause attribution agrees, with one refinement: `Profile.Allocs` (sample_rate=1) puts every byte at `_ldiv! @ triangular.jl:966` reached from the sweep's `ldiv!` calls, and the allocated *objects* are boxed `SubArray` arguments — i.e. the inner `generic_trimatdiv!` call is executed as a runtime dispatch on 1.11 (devirtualization failure), 4 boxes x 64 B = **256 B per `ldiv!` call**: tridiag200 = 12 supernodes x 2 sweeps x 256 B = 6144; the "96/468 allocs" in the profiles are boxes, not calls (24/140 calls). The vector form devirtualizes fine on 1.11 — that is the whole single-RHS/multi-RHS asymmetry, as this PR says.
- Extra CI context for whoever reconciles this with the dashboards: `QA (julia 1)` on `main` has been red since 339ca246 (2026-08-01) anyway — for an **unrelated** JET failure on 1.12 (`33 passed, 1 failed, 8 broken`; runtime-dispatch reports in Base `show`/Static/SparseArrays frames): https://github.com/SciML/LinearSolve.jl/actions/runs/30716868800/job/91413886406. Last green was a507d857. That failure is separate from this PR's subject and still needs an owner.

## Delta 1: the per-column kernel loop regresses wide-RHS solves on 1.11

I benchmarked three variants of the multi-RHS sweep (min over 2x2000 warm `solve!(X, F, B; refine = 0)` runs, `OPENBLAS_NUM_THREADS=1`, `BLAS.set_num_threads(1)`, poisson2d(k), AMD EPYC 7502, times in us):

**Julia 1.11.9**

| case | `main` (stdlib trtrs) | per-column kernels (this PR's shape) | fused kernels |
|---|---|---|---|
| k=30 n=900 nrhs=3 | 172.4 | 138.2 | **133.2** |
| k=30 n=900 nrhs=32 | **632.8** | 874.1 (+38%) | 725.9 (+15%) |
| k=60 n=3600 nrhs=3 | 783.3 | 652.0 | **625.2** |
| k=60 n=3600 nrhs=32 | **2827.0** | 3968.8 (+40%) | 3308.9 (+17%) |

**Julia 1.12.6** (fused vs main): 164.2/162.6, 622.0/625.9, 749.6/749.8, 2791.8/2828.0 — neutral everywhere. **Julia 1.10.11** (fused vs main): 165.2/165.7, 642.2/638.7, 758.2/758.3, 2870.1/2860.4 — neutral.

"Per-column kernels" = applying the vector kernel once per RHS column below the cutoff, which is what this PR's `_unit_lower_solve!`/`_upper_solve!` matrix methods do (measured on an equivalent local variant). It re-sweeps the whole pivot block once per RHS, and on 1.11's codegen that costs +38-40% vs `main` at 32 RHS. Fusing the kernels over the RHS columns (j outer, r middle, `@simd` over i — the pivot-block column `W[:,j]` is loaded once per j instead of once per (j, r)) keeps the small-nrhs win and cuts the wide-RHS penalty to +15-17%; crossover on 1.11 is ~8 RHS (fused 216.6 vs main 224.1), and on 1.12 fused matches trtrs at every point. The residual +15% at 32 RHS on 1.11 only is 1.11-codegen (same source is neutral on 1.10/1.12); I did not chase it further — an nrhs-dependent switch to `trsm!` below the cutoff would close it at the price of a tuned constant.

Fused-kernel bodies, as a diff against this PR's branch (allocation result identical — 6/6 on 1.11 — and Runic-clean):

```diff
 @inline function _unit_lower_solve!(W::AbstractMatrix{Tv}, X::AbstractMatrix{Tv}, np::Int) where {Tv}
-    @inbounds for r in axes(X, 2)
-        _unit_lower_solve!(W, view(X, :, r), np)
+    @inbounds for j in 1:np
+        for r in axes(X, 2)
+            xj = X[j, r]
+            iszero(xj) && continue
+            @simd for i in (j + 1):np
+                X[i, r] = muladd(-W[i, j], xj, X[i, r])
+            end
+        end
     end
     return nothing
 end

 @inline function _upper_solve!(W::AbstractMatrix{Tv}, X::AbstractMatrix{Tv}, np::Int) where {Tv}
-    @inbounds for r in axes(X, 2)
-        _upper_solve!(W, view(X, :, r), np)
+    @inbounds for j in np:-1:1
+        Wjj = W[j, j]
+        for r in axes(X, 2)
+            xj = X[j, r] / Wjj
+            X[j, r] = xj
+            iszero(xj) && continue
+            @simd for i in 1:(j - 1)
+                X[i, r] = muladd(-W[i, j], xj, X[i, r])
+            end
+        end
     end
     return nothing
 end
```

(Division stays `X[j, r] / Wjj` rather than a hoisted reciprocal so columns stay bit-identical to the single-RHS kernel.)

## Delta 2: the non-BLAS fallback above the cutoff re-enters the allocating path

```julia
_panel_unit_lower_trsm!(A::AbstractMatrix, B::AbstractMatrix) =
    ldiv!(UnitLowerTriangular(A), B)
```

is the exact stdlib matrix-RHS chain this PR exists to avoid, so on 1.11 a non-`BlasFloat` eltype (or non-strided scratch) with a panel at/above `PANEL_BLAS_CUTOFF` still allocates 256 B per call. Nothing asserts it — the `== 0` QA tests are Float64-only — so I can't show a red test for it; it just leaves the header's "allocation-free after warmup" claim false in that corner. The per-column **vector** `ldiv!` form is statically clean on 1.11 (it is the path the passing single-RHS poisson case already exercises at np=71):

```julia
else
    for r in axes(Yb, 2)
        ldiv!(UnitLowerTriangular(view(Ws, 1:np, 1:np)), view(Yb, :, r))
    end
end
```

(and `UpperTriangular` respectively).

## Tests: making the trsm leg's coverage explicit

The committed suite covers the `trsm!` leg only incidentally: the QA poisson30 case (widest panel 71) asserts *allocations* but never checks the multi-RHS residual, and the Core column-match matrices reach it only via `sprand(800, 800, 0.01) + 12I` (widest panel 512 on my run — seed-dependent). Since the wide-panel block A3 = poisson3d(16) already asserts `widest >= PANEL_BLAS_CUTOFF`, hanging deliberate multi-RHS coverage on it (plus a BigFloat matrix RHS for the generic kernel leg) makes the coverage self-documenting:

```diff
     x3 = similar(b3)
     SNLU.solve!(x3, F3, b3; refine = 0)
     @test norm(A3 * x3 - b3) <= 1.0e-11 * norm(b3)
+    # multi-RHS on the same factor covers the trsm-backed block solves
+    B3 = randn(size(A3, 1), 5)
+    X3 = similar(B3)
+    SNLU.solve!(X3, F3, B3; refine = 0)
+    @test norm(A3 * X3 - B3) <= 1.0e-11 * norm(B3)
+    for r in 1:5
+        xr = similar(b3)
+        SNLU.solve!(xr, F3, view(B3, :, r); refine = 0)
+        @test X3[:, r] ≈ xr rtol = 1.0e-12
+    end
 end
```

```diff
     x = similar(b)
     SNLU.solve!(x, F, b)
     @test norm(T * x - b) <= big"1e-60" * norm(b)
+    BB = BigFloat.(reshape(1:80, 40, 2))
+    XB = similar(BB)
+    SNLU.solve!(XB, F, BB)
+    @test norm(T * XB - BB) <= big"1e-60" * norm(BB)
```

## Verification of the combined variant (fused kernels + vector fallback + tests)

| Check | Result |
|---|---|
| `test/qa/supernodal_allocations.jl`, Julia 1.11.9 | **6/6** (4/2 before the fix) |
| same, 1.10.11 / 1.12.6 | 6/6 and 8/8 (AllocCheck static proofs still pass with the new kernels in the module) |
| `test/Core/supernodal_lu.jl` incl. the new tests, with RecursiveFactorization loaded, all three versions | green, 0 fail/error ("hand-rolled panel kernels" 18/18, "generic eltypes" 3/3) |
| Runic `--check` on both files | clean |
| `typos` | clean |

Not verified: GPU groups, downstream, and the pre-existing `QA (julia 1)` JET red above.

I am not opening a second PR; if the numbers persuade, fold the two deltas and the tests into this branch (diffs above apply on top of it). If the fused kernels feel like scope creep for a QA fix, delta 2 + the tests are the part I'd argue belongs here regardless.

https://claude.ai/code/session_01UGVN6qeNL2jGCtaYg1386X

ChrisRackauckas-Claude:
Both points addressed in 36bd2b41.

## "should this also use the non-BLAS fallback for smaller matrices?"

Small matrices already took the kernels for every element type, so there was nothing to change there — but checking it turned up that the `ldiv!` fallback was pointless. The Julia 1.11 allocation hits **exactly** the BlasFloat types:

| eltype | `BlasFloat` | alloc, np=8 and np=120 |
|---|---|---|
| Float64 / Float32 / ComplexF64 | true | **256 B** |
| Float16 | false | 0 |
| BigFloat | false | 20072 / 4492904 (inherent to BigFloat arithmetic) |

So it is LinearAlgebra's BLAS-backed triangular path that boxes, not the generic one. And for non-BLAS types the kernels beat `ldiv!` anyway (1.1–1.2× on Float16 at every size). Non-BLAS element types now always take the kernels; the `ldiv!` methods stay only as a safety net for a BlasFloat eltype in a non-strided array.

## "but also, blas is slow on small matrices"

Correct, and it made the first cut wrong in *both* directions. `PANEL_BLAS_CUTOFF` is a pure `np` threshold tuned for the single-RHS gemv shapes; for a matrix RHS the crossover is two-dimensional. Measured kernels vs `trsm!` on Float64, cross-checked at 1 and 8 BLAS threads, np in 4..512 and nrhs in 1..32:

| region | winner | margin | what the old rule did |
|---|---|---|---|
| nrhs ≤ 3, any np up to 512 | kernels | 1.1–4× | sent np≥64 to BLAS, losing up to 1.9× |
| np < 16, any nrhs up to 32 | kernels | 1.0–2.9× | correct |
| otherwise, `np*nrhs` large | `trsm!` | 1.1–7× | kept kernels below np=64, losing up to 2.3× |

A wide panel does not rescue `trsm!` if the block stays narrow — at np=512, nrhs=1 the kernels are 3.6× faster. Conversely np=24, nrhs=6 already favours `trsm!`. Hence three clauses rather than one:

```julia
const PANEL_TRSM_MIN_NP   = 16
const PANEL_TRSM_MIN_NRHS = 4
const PANEL_TRSM_MIN_WORK = 128
```

The rule agrees with the measurement at **all 64 sampled (np, nrhs) points, with no disagreement worse than 1.3×** — about the run-to-run noise floor, which is why the thresholds are deliberately coarse.

## Net effect

It is now faster as well as allocation-free. End-to-end `solve!` on poisson2d, Julia 1.11, against `main`'s unconditional `ldiv!`:

| case | main | this branch | change |
|---|---|---|---|
| k=30, nrhs=1 | 96.6 µs | 91.2 µs | −5.6 % |
| k=30, nrhs=3 | 161.2 µs | 155.5 µs | −3.5 % |
| k=30, nrhs=8 | 278.9 µs | 237.5 µs | **−14.8 %** |
| k=30, nrhs=32 | 898.1 µs | 767.4 µs | **−14.6 %** |
| k=60, nrhs=8 | 1251.4 µs | 1051.2 µs | **−16.0 %** |
| k=60, nrhs=32 | 4066.8 µs | 3374.8 µs | **−17.0 %** |

Re-verified after the change: allocation test 6/6 on 1.10 and 1.11, 8/8 on 1.12 (AllocCheck static proof); `test/Core/supernodal_lu.jl` clean; residuals ~1e-16 across nrhs ∈ {1,3,4,8,32} on both dispatch paths and for Float16 / ComplexF64 / Float32 / BigFloat; Runic and typos clean.

## One thing I did not change

The **single-RHS** path has the same latent mis-tuning: it switches to `ldiv!` at `np >= PANEL_BLAS_CUTOFF = 64`, but at nrhs=1 the kernels measured 2.2–3.6× *faster* than BLAS at np=64..512. That is a pre-existing perf question on a path this PR is not otherwise touching, and it deserves its own before/after on the real `_solve_panels!` timings that the header comment documents, so I left it alone. Happy to do it as a follow-up if you want.

ChrisRackauckas-Claude:
Reworked in 3ff6acf6 to the tiered arrangement, plus #1172 to measure the cutoffs.

## Backends

| band | backend |
|---|---|
| `np <= PANEL_KERNEL_MAX_NP` (256) | in-tree column kernels |
| up to `PANEL_BLAS_MIN_NP` (1024) | `TriangularSolve.ldiv!` when RecursiveFactorization is loaded, else `BLAS.trsm!` |
| `np > PANEL_BLAS_MIN_NP` (1024) | `BLAS.trsm!` |

TriangularSolve gets the whole middle band rather than a hand-derived sub-range. It already carries its own small-size cutoffs (`VECTOR_RHS_CUTOFF = 128` plus the blocked-kernel thresholds) and falls back by itself where a naive sweep wins, so there is nothing gained by re-deriving that boundary in LinearSolve. The large end stays on BLAS on the expectation that its blocked sweep still wins once the triangle leaves cache.

## Why hooks are required at all

`TriangularSolve.ldiv!` is a **distinct function** from `LinearAlgebra.ldiv!`, not an extension of it:

```
TriangularSolve.ldiv! === LinearAlgebra.ldiv!  : false
parentmodule(TriangularSolve.ldiv!)           : TriangularSolve
parentmodule(LinearAlgebra.ldiv!)             : LinearAlgebra
```

So a bare `ldiv!(UnitLowerTriangular(view(Ws,1:np,1:np)), Yb)` in these sweeps dispatches to `LinearAlgebra` (`triangular.jl:971`) → BLAS, no matter how the environment is configured. Loading TriangularSolve changes nothing at that call site. The same holds a level up: `RecursiveFactorization.lu` returns a plain `LinearAlgebra.LU` and a bare `ldiv!(F, b)` goes to `getrs!` — which is exactly why the extension carries `_rf_ldiv!` and why #1169 exists.

`defaultalg` cannot reach them either: it selects an algorithm for a `LinearProblem`, and these are dense triangular sub-solves inside a sweep. `numeric.jl:290-296` already says this for the factorization phase. That phase had the hook arrangement (`_panel_rdiv!`/`_panel_ldiv!`, overridden in the RF extension); **the solve phase had none**, so nothing could override it. That was the gap.

## `[compat]` floor moved to TriangularSolve 0.2.2

The matrix `ldiv!(::UpperTriangular{T,<:StridedMatrix{T}}, ::StridedMatrix{T}, ::Val)` family landed in 0.2.2. On **0.2.1** there is no `UpperTriangular` method at all and the call silently forwards through

```julia
ldiv!(A, B, ::Val = Val(true)) = LinearAlgebra.ldiv!(A, B)   # TriangularSolve.jl:750
```

an untyped catch-all — a runtime dispatch onto the very path that allocates on Julia 1.11. Worth noting the same applies to the pre-existing `_rf_ldiv!`, whose `UpperTriangular` call is that stub on 0.2.1.

## Cutoffs are inherited, not measured here

`256` comes from #1169; #1164 confirms 256 for a `Matrix` orientation (512 wrapped) and finds multi-RHS sweeps still ahead through N=1000. Both were measured on whole LU back-solves, which is a different shape from a supernode panel trsm, so I have marked them provisional in the source and opened **#1172** to measure them here — including the `nrhs` dependence and the forward/back sweeps separately.

## Corrections to my earlier comments on this PR

Three things I posted before this are wrong and I want them on the record rather than buried:

1. **"TriangularSolve's `UpperTriangular` path is catastrophic"** — no. My environment had resolved a stale **0.2.1** (compat allowed `0.2`, the Manifest was pinned). I was measuring the forwarding stub above, not a slow kernel. 0.2.2+ has a real matrix upper kernel.
2. **The (np, nrhs) dispatch rule in 36bd2b41** was fitted to unsound benchmarks: `@belapsed` re-ran on the same buffer, so the RHS degenerated to `Inf`, and the kernels' `iszero(xj) && continue` fast path made them look faster than they are.
3. **A claimed 640× pathology on main** did not reproduce. That box is shared and was at load ~90 on 128 cores; 64-thread BLAS under contention stalls. Machine artifact, not a property of main.

Both methodology traps are written up in #1172 so the next person measuring these does not repeat them.

## Verification

Julia 1.11 unless noted, run with and without RecursiveFactorization loaded:

| check | result |
|---|---|
| `supernodal_allocations.jl` | 6/6 on 1.10 and 1.11, 8/8 on 1.12 (AllocCheck static proof) |
| all three tiers, alloc + residual | `np` = 18 / 150 / 400 / 700 / 1200, `nrhs` ∈ {4, 8}: 0 alloc, residuals ~1e-15 |
| element types | Float32, ComplexF64, BigFloat (generic kernels) — correct; BigFloat allocates, which is inherent to BigFloat arithmetic |
| tier routing | confirmed the hook resolves to `LinearSolveRecursiveFactorizationExt` and that `np`=400 takes TriangularSolve, `np`=1200 takes `trsm!` |
| `test/Core/supernodal_lu.jl` | pass |
| Runic `--check`, `typos` | clean |

I have **not** re-run an end-to-end timing comparison for this tiering — the box is still heavily loaded and, per the corrections above, timings taken on it are not trustworthy. The claim here is correctness and structure; the performance question is #1172.

PR #1163, opened 2026-08-08
KLU: extract factor parts through concrete helpers, not getproperty recursion

### LINKED ISSUE #1148: JET: KLUFactorization solve is no longer dispatch-clean; `getproperty` self-recursion makes `p, i, x, z = klu._L` destructure a SparseMatrixCSC

AJ0070:
## Checklist

## Additional context

Fixes #1148. This is what has been keeping the QA job red: the failure at `test/qa/jet.jl:136` aborts the QA group before anything behind it runs.

- `getproperty(::AbstractKLUFactorization)` read `klu._L` and `klu.n` back through itself; inference abandons the self-recursive call, so the destructure was inferred against a Union spanning every branch, admitting `indexed_iterate(::SparseMatrixCSC, ::Int)` and failing `JET.@test_opt solve(prob_sparse, KLUFactorization())` with a 40-error cascade.
- The `:_L`/`:_U`/`:_F` bodies move verbatim into concrete helpers (`_extract_L`/`_extract_U`/`_extract_F`) that read only through `getfield`/`_numeric_struct`, and the `:L`/`:U`/`:F` branch calls them directly via `_assemble_factor`. Same fix the file already applied once for the factor sizes, finished for the factor parts. Each access still extracts fresh buffers, so `increment!` cannot be applied twice.
- Getting past `jet.jl:136` unmasked one failure CI has never reached: `JET.@test_opt solve(prob)` at `jet.jl:175` flags the MKL branch of the generated `solve!` (JET analyzes every branch regardless of which runs), whose failure-logging path does `string(::Type)` inside Base's `show` machinery. Scoped with `target_modules = (LinearSolve, SciMLBase)`, the treatment the file already uses for the direct MKL test; real dispatch in package code is still reported.
- With both, the entire QA group passes locally for the first time: `jet.jl` (34 pass, 8 expected-broken), `allocations.jl`, `supernodal_allocations.jl`, `qa.jl` (19 pass, 1 broken tracked in #1058).
- KLU factor reconstruction `(Rs .\ A)[p, q] == L*U + F` verified for real, block-triangular, and complex matrices; the complex path exercises the non-Float64 helper branches. No test additions needed: `jet.jl:136` is the regression test, and the existing `KLU factor parts` testset covers reconstruction.
- The local `Core` run aborts in `test/Core/forwarddiff_gpu.jl` with JLArrays 0.3.2, reproduced identically on pristine `main`; that is the pre-existing breakage `fix/forwarddiff-gpu-jlarrays-032` addresses, unrelated to this change.

### DIFF
--- a/src/KLU/klu.jl
+++ b/src/KLU/klu.jl
@@ -313,63 +313,21 @@ function getproperty(
     if s ∉ (:L, :U, :F, :p, :q, :R, :Rs, :(_L), :(_U), :(_F))
         return getfield(klu, s)
     end
-    # Factor parts: the sizes come from `_numeric_struct` rather than from
-    # `klu.lnz` etc., because reading them back through this same `getproperty`
-    # is a self-recursive call that inference abandons — the resulting `Any`
-    # sizes make the `_extract!` keyword calls below uninferable.
-    n = getfield(klu, :n)
+    # Factor parts live in concrete helper functions (`_extract_L` and friends)
+    # rather than inline: everything they need comes from `getfield`/
+    # `_numeric_struct` directly, because reading `klu.lnz` or `klu._L` back
+    # through this same `getproperty` is a self-recursive call that inference
+    # abandons — the resulting `Any` values made the `_extract!` keyword calls
+    # (and, in the `:L`/`:U`/`:F` branch below, the tuple destructure) uninferable,
+    # to the point of admitting scalar iteration of a `SparseMatrixCSC` (#1148).
     if s === :(_L)
-        lnz = Int(_numeric_struct(klu).lnz)
-        Lp = Vector{Ti}(undef, n + 1)
-        Li = Vector{Ti}(undef, lnz)
-        Lx = Vector{Float64}(undef, lnz)
-        Lz = Tv == Float64 ? C_NULL : Vector{Float64}(undef, lnz)
-        _extract!(klu; Lp, Li, Lx, Lz)
-        return Lp, Li, Lx, Lz
+        return _extract_L(klu)
     elseif s === :(_U)
-        unz = Int(_numeric_struct(klu).unz)
-        Up = Vector{Ti}(undef, n + 1)
-        Ui = Vector{Ti}(undef, unz)
-        Ux = Vector{Float64}(undef, unz)
-        Uz = Tv == Float64 ? C_NULL : Vector{Float64}(undef, unz)
-        _extract!(klu; Up, Ui, Ux, Uz)
-        return Up, Ui, Ux, Uz
+        return _extract_U(klu)
     elseif s === :(_F)
-        fnz = Int(_numeric_struct(klu).nzoff)
-        # We often don't have an F, so create the right vectors for an empty SparseMatrixCSC
-        if fnz == 0
-            Fp = zeros(Ti, n + 1)
-            Fi = Vector{Ti}()
-            Fx = Vector{Float64}()
-            Fz = Tv == Float64 ? C_NULL : Vector{Float64}()
-        else
-            Fp = Vector{Ti}(undef, n + 1)
-            Fi = Vector{Ti}(undef, fnz)
-            Fx = Vector{Float64}(undef, fnz)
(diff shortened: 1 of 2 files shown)


AJ0070:
@ChrisRackauckas this is ready for review.

PR #1165, opened 2026-08-08
Fix CUDAOffload32MixedLUFactorization: complete the modular-stack migration

AJ0070:
## Checklist

## Additional context (should be very concise)

`CUDAOffload32MixedLUFactorization` fails on every use with `UndefVarError: CUDA not defined in LinearSolveCUDAExt`: its two functions were skipped in the extension's migration to the modular cuSOLVER/CUDACore stack, leaving nine references to the old names (`CUDA.functional` x1, bare `CuMatrix` x4, bare `CuVector` x4). This qualifies them with `CUDACore.`, matching the rest of the file.

Validated on a Tesla T4 with exactly this patch: the mixed path constructs, solves at expected Float32 accuracy (relerr 7e-7 on a dense n=2000 system), and runs about 2x faster than full-precision `CudaOffloadLUFactorization` on the same problem; all sibling GPU paths still pass. No tests added because the repo has no GPU CI job to hook into; the break was unconditional at `init`, so a construct-and-init smoke on any GPU runner would prevent regression.

AI Disclosure: Used Opus 5

### DIFF
--- a/ext/LinearSolveCUDAExt.jl
+++ b/ext/LinearSolveCUDAExt.jl
@@ -184,9 +184,9 @@ function SciMLBase.solve!(
         fact, A_gpu_f32, b_gpu_f32, u_gpu_f32 = LinearSolve.@get_cacheval(cache, :CUDAOffload32MixedLUFactorization)
         if isempty(A_gpu_f32)
             m, n = size(cache.A)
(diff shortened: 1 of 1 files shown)


AJ0070:
@ChrisRackauckas

PR #1167, opened 2026-08-08
Thread backend convergence metadata through the a-posteriori safety check

### LINKED ISSUE #1166: solve! discards iteration count and residual when the a-posteriori safety check fails
## Description

When an iterative solve converges by the backend's own stopping test but fails LinearSolve's a-posteriori residual safety check, the returned solution reports `iters = 0` and `resid = nothing`. The real convergence metadata is computed and then thrown away, so users debugging an `APosterioriSafetyFailure` lose exactly the diagnostics they need (how many iterations ran, what residual was reached).

## Where

- `src/factorization.jl` (`_check_residual_safety`, around L138-145): on failure it returns

  ```julia
  SciMLBase.build_linear_solution(
      alg, y, nothing, nothing; retcode = ReturnCode.APosterioriSafetyFailure
  )
  ```

  passing `nothing` for the residual and defaulting `iters`, even though the check itself computed `res_norm` two lines earlier.

- `ext/LinearSolvePETScExt.jl` (`solve!`, around L822-836): `iters` and `resid` are read from the KSP right before `postsolve_solution_check` is called; the early return on failure drops both.

## Reproduction

2-D FD Laplacian (N = 40k), PETSc GAMG-CG, `reltol = 1e-8`:

```julia
using LinearSolve, PETSc, SparseArrays, MPI
A = <2-D 5-point Laplacian, 200x200 grid>; b = ones(size(A, 1))
alg = PETScAlgorithm(:cg; pc_type = :gamg)
sol = solve(LinearProblem(A, b), alg; reltol = 1e-8)
sol.retcode  # APosterioriSafetyFailure
sol.iters    # 0   <- PETSc's own log says CONVERGED_RTOL iterations 10
```

Root cause of the disagreement itself is well understood: preconditioned CG converges in the preconditioned norm while the safety check uses the true residual (`ksp_norm_type = "unpreconditioned"` is a user-side workaround). This issue is only about the metadata being dropped on that path.

## Suggested fix

Let callers pass the metadata through, defaulting to today's behavior:

```julia
function _check_residual_safety(cache, alg, A_original, y; iters = 0, resid = nothing)
    ...
    return SciMLBase.build_linear_solution(
        alg, y, something(resid, res_norm), nothing;
        retcode = ReturnCode.APosterioriSafetyFailure, iters
    )
end
```

Even with no caller changes, `res_norm` (already computed) is a strictly better residual to report than `nothing`. The PETSc extension would then forward its `iters`/`resid`. Happy to submit this as a PR.

## Context

Found while standing up the distributed LinearSolve benchmarks for SciMLBenchmarks; the benchmark surfaced it because iteration counts are printed on every run as a scaling-validity check. Related finds from the same effort: #1142 and the CUDAOffload32Mixed fix PR.

AJ0070:
Fixes #1166.

## Checklist

## Additional context

- _check_residual_safety now accepts iters and resid kwargs and puts them on the APosterioriSafetyFailure solution instead of dropping them.
- The PETSc extension forwards the KSP iteration count and residual norm through postsolve_solution_check, on both the serial and the MPI PSparseMatrix paths.
- The MPI path previously hardcoded nothing for resid even though it had just computed the distributed residual norm; it now reports the forwarded KSP norm with the computed norm as fallback.
- Defaults keep the failure solution type identical to the success path. Reporting the check's own res_norm by default would put a Float64/Nothing union in solve!'s return type for every LU algorithm and break the concrete-return QA tests in test/qa/jet.jl; this was measured, not guessed.
- Tests: metadata pass-through and unchanged defaults in test/Core/default_algs.jl, plus PETSc serial and MPI safety-failure tests asserting iters > 0 and a positive Float64 resid.
- Verified locally: solve! return types stay concrete for the default solver, LUFactorization and GenericLUFactorization; QA jet.jl and allocations suites green; Runic clean.

AI Disclosure: Used Opus 5

### DIFF
--- a/ext/LinearSolvePETScExt.jl
+++ b/ext/LinearSolvePETScExt.jl
@@ -617,14 +617,17 @@ function gather_petsc_vec_to_all!(u::AbstractVector, petsclib, petsc_x)
     return nothing
 end
 
-function postsolve_solution_check(cache, alg, pcache, A::AbstractMatrix, u::AbstractVector)
+function postsolve_solution_check(
+        cache, alg, pcache, A::AbstractMatrix, u::AbstractVector;
+        iters::Int = 0, resid = nothing
+    )
     if pcache.comm == MPI.COMM_SELF || A isa SparseMatrixCSC
-        return LinearSolve._check_residual_safety(cache, alg, A, u)
+        return LinearSolve._check_residual_safety(cache, alg, A, u; iters, resid)
     end
     return nothing
 end
 
-postsolve_solution_check(cache, alg, pcache, A, u) = nothing
+postsolve_solution_check(cache, alg, pcache, A, u; iters::Int = 0, resid = nothing) = nothing
 
 function run_ksp_mpi!(pcache, petsclib, alg, b::AbstractVector, u::AbstractVector)
     petsc_x = pcache.petsc_x
@@ -830,7 +833,7 @@ function SciMLBase.solve!(cache::LinearCache, alg::PETScAlgorithm; kwargs...)
         retcode = ReturnCode.Failure
     end
     if retcode === ReturnCode.Success
-        failed = postsolve_solution_check(cache, alg, pcache, cache.A, cache.u)
+        failed = postsolve_solution_check(cache, alg, pcache, cache.A, cache.u; iters, resid)
         failed !== nothing && return failed
     end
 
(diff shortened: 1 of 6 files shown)


AJ0070:
@ChrisRackauckas

PR #1178, opened 2026-08-09
Tune SupernodalLU panel solve cutoffs

### LINKED ISSUE #1172: Benchmark the SupernodalLU panel-solve backend cutoffs directly
The multi-RHS pivot-block triangular solves in `src/SupernodalLU/solve.jl` pick between three backends by panel width `np`:

| band | backend |
|---|---|
| `np <= PANEL_KERNEL_MAX_NP` (256) | in-tree column kernels |
| up to `PANEL_BLAS_MIN_NP` (1024) | `TriangularSolve.ldiv!` when RecursiveFactorization is loaded, else `BLAS.trsm!` |
| `np > PANEL_BLAS_MIN_NP` (1024) | `BLAS.trsm!` |

**Both boundaries are inherited, not measured on this kernel.** They come from the dense back-solve work: #1169 puts the in-tree-kernel/BLAS crossover at 256, and #1164 confirms 256 for a `Matrix` orientation (512 wrapped) while finding multi-RHS sweeps still ahead through N=1000. Those were measured on whole LU back-solves, which is a different shape from a supernode panel trsm, so the numbers should be treated as placeholders until measured here.

## What to measure

1. **Kernel → TriangularSolve crossover.** TriangularSolve carries its own small-size cutoffs (`VECTOR_RHS_CUTOFF = 128` plus blocked-kernel thresholds) and falls back on its own, so the extension currently hands it the entire band below `PANEL_BLAS_MIN_NP` rather than re-deriving where it stops winning. Worth confirming that its internal fallbacks really are the right boundary at panel shapes, or whether LinearSolve should cut over earlier.
2. **TriangularSolve → BLAS crossover.** `PANEL_BLAS_MIN_NP = 1024` is a guess that BLAS's blocked sweep eventually wins. Unverified.
3. **`nrhs` dependence.** Both constants are pure `np` thresholds, but the crossover for a trsm plausibly depends on the width of the right-hand-side block too. #1164 already saw multi-RHS behave differently from vectors.
4. **Forward vs back sweep separately.** The unit-lower and upper solves are different kernels and need not cross over at the same place.

## Methodology warnings

Two traps cost me real time on #1158, both worth repeating here:

- **Refresh the right-hand side per sample.** Re-running a triangular solve on the same buffer degenerates it — for a unit-lower solve with `rand` off-diagonals the values blow up to `Inf` — and the in-tree kernels have an `iszero(xj) && continue` fast path, so degenerate data makes them look faster than they are. Use `@belapsed ... setup=(Y = copy(Y0)) evals=1` and a well-conditioned panel.
- **Pin BLAS threads and check machine load.** `BLAS.trsm!` on small panels is thread-count sensitive. On a shared box at load ~90 I measured small-panel `trsm!` calls degrading by orders of magnitude, which did not reproduce on a quiet machine. Any cutoff measured under contention is meaningless. Record `BLAS.get_num_threads()` and the load average alongside the numbers.
- Check the installed `TriangularSolve` version. `0.2.1` has no matrix `UpperTriangular` `ldiv!` at all and silently forwards to `LinearAlgebra.ldiv!` through an untyped catch-all; the matrix family landed in `0.2.2`. The `[compat]` floor is now `0.2.2` for this reason.

Realistic panel widths matter for picking the sweep range: a poisson-2D k=30 factorization has 70 panels with `np` from 1 to 71, median 8. The interesting bands here are well above that, so the benchmark should include problems that actually produce wide supernodes (3D meshes, dense blocks) rather than only 2D Poisson.

Context: #1158, #1164, #1169.

ChrisRackauckas-Claude:
> [!NOTE]

Closes https://github.com/SciML/LinearSolve.jl/issues/1172.

## What changed

Adds a reproducible benchmark for forward and backward SupernodalLU multi-RHS panel solves. It refreshes the RHS for every sample, pins BLAS to one thread, and records machine load.

The solve routing now retains the in-tree kernels for one-column RHS at all panel widths and keeps TriangularSolve for wider panels through `np = 1792`; larger panels use BLAS. The no-extension fallback remains conservative at 256.

On the pinned benchmark host (Julia 1.12.6, TriangularSolve 0.2.4, one BLAS thread), at `np = 1280`, `nrhs = 4`, TriangularSolve took 260349 ns versus BLAS at 1008164 ns for the lower sweep and 269298 ns versus 996645 ns for the upper sweep. At `np = 2048`, `nrhs = 8`, BLAS becomes faster (3247412 ns versus 4311808 ns lower; 3039994 ns versus 4408396 ns upper).

## Verification

- `GROUP=Core /home/crackauc/.juliaup/bin/julia +1.12 --project -e 'using Pkg; Pkg.test()'`\n  - Passed: `Testing LinearSolve tests passed`.\n- `GROUP=QA /home/crackauc/.juliaup/bin/julia +1.12 --project -e 'using Pkg; Pkg.test()'`\n  - Passed: Quality Assurance 48 pass, 1 expected broken.\n- `Runic --check --diff` on the four changed files, `typos` on those files, and `git diff --check` all passed.\n- Ran the new benchmark smoke test with `np = 8`, `nrhs = 1`; all three routes completed.\n\n## Not verified\n\nThe docs build is not applicable: this does not change public API or documentation. I did not run the Everything, downstream, or GPU groups.

### DIFF
--- a/benchmarks/supernodal_panel_solve.jl
+++ b/benchmarks/supernodal_panel_solve.jl
@@ -1,0 +1,60 @@
+using BenchmarkTools, LinearAlgebra, LinearSolve, Random, RecursiveFactorization
+using TriangularSolve
+
+const SNLU = LinearSolve.SupernodalLU
+
+machine_load() = Sys.islinux() ? first(split(read("/proc/loadavg", String))) : "unavailable"
+
+function panel_matrix(np)
+    W = Matrix{Float64}(I, np, np)
+    scale = inv(sqrt(np))
+    for j in 1:np
+        for i in 1:(j - 1)
+            W[i, j] = scale * randn()
+        end
+        for i in (j + 1):np
+            W[i, j] = scale * randn()
+        end
+    end
+    return W
+end
+
+function lower_times(W, Y0, np)
+    kernel = @belapsed SNLU._unit_lower_solve!($W, Y, $np) setup = (Y = copy($Y0)) evals = 1
+    triangularsolve = @belapsed TriangularSolve.ldiv!(
+        UnitLowerTriangular(view($W, 1:$np, 1:$np)), Y, Val(false)
+    ) setup = (Y = copy($Y0)) evals = 1
+    blas = @belapsed SNLU._panel_unit_lower_trsm!($W, Y, $np) setup = (Y = copy($Y0)) evals = 1
+    return kernel, triangularsolve, blas
+end
+
+function upper_times(W, Y0, np)
+    kernel = @belapsed SNLU._upper_solve!($W, Y, $np) setup = (Y = copy($Y0)) evals = 1
+    triangularsolve = @belapsed TriangularSolve.ldiv!(
+        UpperTriangular(view($W, 1:$np, 1:$np)), Y, Val(false)
+    ) setup = (Y = copy($Y0)) evals = 1
+    blas = @belapsed SNLU._panel_upper_trsm!($W, Y, $np) setup = (Y = copy($Y0)) evals = 1
+    return kernel, triangularsolve, blas
+end
+
+function benchmark_panel_cutoffs(; nps = (8, 16, 32, 64, 128, 256, 512, 768, 1024, 1280, 1536, 1792, 2048), nrhss = (1, 2, 4, 8, 16, 32))
+    BLAS.set_num_threads(1)
+    Random.seed!(1172)
+    BenchmarkTools.DEFAULT_PARAMETERS.seconds = 0.25
+    BenchmarkTools.DEFAULT_PARAMETERS.samples = 100
+    println("Julia $(VERSION), TriangularSolve $(Base.pkgversion(TriangularSolve)), BLAS threads $(BLAS.get_num_threads()), load $(machine_load())")
+    println("np nrhs sweep kernel_ns triangularsolve_ns blas_ns")
+    for np in nps, nrhs in nrhss
+        W = panel_matrix(np)
+        Y0 = randn(np, nrhs)
+        for (sweep, times) in (("lower", lower_times(W, Y0, np)), ("upper", upper_times(W, Y0, np)))
+            kernel, triangularsolve, blas = times
+            println("$np $nrhs $sweep $(round(Int, kernel * 1.0e9)) $(round(Int, triangularsolve * 1.0e9)) $(round(Int, blas * 1.0e9))")
+        end
+    end
+    return nothing
+end
+
+if abspath(PROGRAM_FILE) == abspath(@__FILE__)
+    benchmark_panel_cutoffs()
+end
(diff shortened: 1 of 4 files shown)


ChrisRackauckas-Claude:
## Benchmark results

Benchmarked on AMD EPYC 7502 with Julia 1.12.6, TriangularSolve 0.2.4, one BLAS thread, and machine load 32.09. Each BenchmarkTools sample copied a fresh RHS. Times are nanoseconds; lower and upper sweeps are measured independently.

| np | nrhs | sweep | in-tree kernel | TriangularSolve | BLAS |
|---:|---:|:---|---:|---:|---:|
| 1280 | 1 | lower | 219799 | 245909 | 620817 |
| 1280 | 1 | upper | 174990 | 240099 | 654947 |
| 1280 | 4 | lower | 855035 | 260349 | 1008164 |
| 1280 | 4 | upper | 763116 | 269298 | 996645 |
| 1280 | 8 | lower | 1967440 | 546747 | 1132624 |
| 1280 | 8 | upper | 1668821 | 571017 | 1028014 |
| 1280 | 16 | lower | 3653921 | 857445 | 1471632 |
| 1280 | 16 | upper | 2796765 | 860145 | 1451352 |
| 1280 | 32 | lower | 6774943 | 1724051 | 2187458 |
| 1280 | 32 | upper | 5956578 | 1744611 | 2239518 |
| 1792 | 1 | lower | 690196 | 771376 | 1380853 |
| 1792 | 1 | upper | 630586 | 779396 | 1407802 |
| 1792 | 4 | lower | 3081644 | 1056215 | 2180698 |
| 1792 | 4 | upper | 2943465 | 1830470 | 2908675 |
| 1792 | 8 | lower | 5733969 | 1757120 | 2244478 |
| 1792 | 8 | upper | 5856628 | 1727710 | 2172378 |
| 1792 | 16 | lower | 11631368 | 2518126 | 3031004 |
| 1792 | 16 | upper | 9598529 | 2483437 | 2978034 |
| 1792 | 32 | lower | 22317381 | 4485496 | 4389107 |
| 1792 | 32 | upper | 20281543 | 4572366 | 4342577 |
| 2048 | 1 | lower | 1063324 | 2138168 | 1973230 |
| 2048 | 1 | upper | 1018335 | 2133299 | 1880520 |
| 2048 | 4 | lower | 4584455 | 2345758 | 2953454 |
| 2048 | 4 | upper | 6622535 | 3317953 | 3821920 |
| 2048 | 8 | lower | 13559468 | 4311808 | 3247412 |
| 2048 | 8 | upper | 10005827 | 4408396 | 3039994 |
| 2048 | 16 | lower | 18193873 | 5846099 | 4146548 |
| 2048 | 16 | upper | 17374617 | 6188088 | 4094988 |
| 2048 | 32 | lower | 36661206 | 11547389 | 6144687 |
| 2048 | 32 | upper | 37340132 | 11740607 | 6047838 |

This supports retaining the in-tree kernels for a single RHS, extending TriangularSolve through np = 1792 for wider panels, and using BLAS past that cutoff. The added benchmark script captures the complete sweep, including intermediate widths and nrhs = 2.

PR #1180, opened 2026-08-09
BlockDiagonal: blockwise LU and QR, make it the default, fix the QRFactorization TypeError, add tutorial

### LINKED ISSUE #203: Tutorial BlockDiagonals
Does LinearSolve.jl specialize automatically with BlockDiagonals.jl?
If so, it might be worth a tutorial to showcase this functionality.

AJ0070:
Fixes #203.

## Checklist

## Additional context

The ask was a tutorial "and making sure it works". Making sure it works turned up a bug, and review turned up that a documented type should actually be specialized and then defaulted to, so this is a docs PR with three fixes attached.

### 1. `QRFactorization` threw on every `BlockDiagonal` solve

`init_cacheval` predicted the factorization type through `ArrayInterface.qr_instance`, which densifies and reports `QRCompactWY`. `do_factorization` called `qr!(A, pivot)`, which returned a `QR` wrapping the `BlockDiagonal`. Assigning that into the differently typed cacheval field raised a `TypeError`.

### 2. LU and QR factorize block by block

- The blocks of a block diagonal system are independent, so the previous behaviour of handing the whole `BlockDiagonal` to `lu!` ran the generic scalar LinearAlgebra path over every structural zero, indexing through the block lookup on each access. It was slower than densifying first.
- Each block is now factorized on its own and the per-block factorizations are stored, which turns `O(N^3)` into `O(sum m_i^3)` and puts every per-block call on BLAS. On 20x20 blocks through `LUFactorization` this is 147x at N = 200 and 1238x at N = 800 against the old path, and 26x to 110x against densifying and calling `lu`.
- Blockwise also makes `QRFactorization(ColumnNorm())` work. An earlier revision documented pivoting as impossible, which is only true when the result has to be stored as one `BlockDiagonal` wrapped factorization. Per block the permutation stays inside a block and is valid, so that limitation is gone rather than documented.

### 3. The default algorithm now uses it

- `BlockDiagonal` previously reached the "not a factorizable operator" arm of `defaultalg` and defaulted to `KrylovJL_GMRES`, which was the worst option available: matrix free, so a full multiply per iteration over structure it never exploits, stopping at the Krylov tolerance instead of machine precision. Measured at 20x20 blocks: 33 ms at N = 200, 220 ms at N = 400, 1.39 s at N = 800, against 0.05, 0.10 and 0.21 ms for the direct blockwise solve.
- `defaultalg` now returns the `LUFactorization` choice for square blocks. Non-square blocks are not independent subsystems, so those keep the generic operator handling and still get `KrylovJL_GMRES`.
- Wiring that up needed the default solver path fixed too. Its LU branch calls `solve!(cache, LUFactorization(...))` with the same cache, whose cacheval is a `DefaultLinearSolverInit`, so a `solve!` keyed on the cacheval type never fired and the stock path then stored a plain `LU` into the slot `init_cacheval` had typed as blockwise. The method now dispatches on `A` and reads the slot through `@get_cacheval`, which covers the direct and default paths alike.
- That also settles the `residualsafety` mismatch: `algchoice_to_alg` builds the slot from a bare `LUFactorization()` while the solve passes `residualsafety = alg.residualsafety`. The blockwise predicate therefore depends only on the matrix, never on algorithm options, so the slot and the solve cannot disagree. `residualsafety` stays supported by factorizing from a copy so the a-posteriori check still sees an unfactored operator.

### Notes for review

- `LUFactorization` has its own `solve!` that calls `lu!`/`lu` directly instead of going through `do_factorization`, so specializing `do_factorization` alone was silently ineffective for it.
- `issuccess` is not defined for QR factorizations, so success detection is per factorization type rather than a blanket call.
- Two traps are recorded as comments so they are not reintroduced: the `init_cacheval` method needs `verbose` and `assumptions` annotated or it is ambiguous against the generic method, and `similar(A, T, 0, 0)` quietly densifies a `BlockDiagonal` to a `Matrix`.
- `test/qa/qa.jl` gains one entry, `_check_residual_safety`, because the blockwise solve now calls it. That is the inventory #1058 made strict.
- Not addressed here, and pre-existing: a one shot `solve(prob)` on a `BlockDiagonal` is still dominated by `init`, roughly 80 ms at N = 400, because `_init_default_cacheval` eagerly builds a cacheval for every algorithm choice. `GenericLUFactorization` is 30 ms of that, `RFLUFactorization` 17 ms and `CholeskyFactorization` 12 ms. Cached workflows get the full win today; one shot calls are still paying that construction. Happy to look at it separately if you want lazy slots.

### Verification

- Full `GROUP=Core` suite passes with zero failures, and `test/qa/qa.jl` is 20 of 20.
- Tests cover both pivots of LU and QR on uniform and non uniform blocks, that the cacheval really is the blockwise type, batched matrix right hand sides, cache reuse with a new `b`, Float32 and ComplexF64, a singular block returning `ReturnCode.Failure` rather than crashing, the default choice for square and non square blocks, and `residualsafety` leaving `A` unfactored.
- Aqua ambiguities pass with BlockDiagonals loaded, including the new `defaultalg` method, and Runic is clean.
- Every tutorial code block was executed verbatim in the docs environment before being written down.
- Benchmarks are single threaded on Windows, Julia 1.12.6.


ChrisRackauckas:
We probably want to make the algorithms specialize on this if we say we support it?

AJ0070:
@ChrisRackauckas

ChrisRackauckas:
Incorporate it into the defaults?

AJ0070:
Agreed on the direction. `defaultalg` currently falls through to the "not factorizable, use A*x" branch for `BlockDiagonal` and picks `KrylovJL_GMRES`, which turns out to be the worst available option by a wide margin. Blocks of 20x20, single threaded:

| blocks | N | default (GMRES) | blockwise LU |
|---|---|---|---|
| 10 | 200 | 33 ms | 0.05 ms |
| 20 | 400 | 220 ms | 0.10 ms |
| 40 | 800 | 1.39 s | 0.20 ms |

GMRES also stops around 1e-7 residual where the direct solve is at machine precision.

One thing to settle before I wire it up. Forcing the default onto its LU choice errors today with `TypeError: expected BlockDiagonalFactorization, got LU`. The default LU branch calls `solve!(cache, LUFactorization(...))` with the same cache, whose cacheval is the `DefaultLinearSolverInit`, so the blockwise `solve!` here does not fire and the stock path then stores a plain `LU` into the slot that `init_cacheval` typed as blockwise. Re-keying the override on the `A` type fixes that, but it also surfaces that `algchoice_to_alg` builds the slot from a bare `LUFactorization()` while the solve passes `residualsafety = alg.residualsafety`, so a `residualsafety = true` default would disagree with its own slot.

Happy to do it either way: fold the `defaultalg` change plus that cacheval fix into this PR, or land this one for the specialization and tutorial and send the default change separately so it gets its own review. Which do you prefer? @ChrisRackauckas

docs/src/tutorials/block_diagonal.md
```diff
@@ -0,0 +1,156 @@
...
+
+Measured on 20x20 blocks, going through `LUFactorization`, this is worth two to
+three orders of magnitude against the previous behaviour of handing the whole
+`BlockDiagonal` to the generic `lu!`:
+
+| blocks | block size | N | before | now |
+|:---|:---|:---|:---|:---|
+| 10 | 20 | 200 | 7.8 ms | 0.05 ms |
+| 20 | 20 | 400 | 67 ms | 0.10 ms |
+| 40 | 20 | 800 | 255 ms | 0.21 ms |
```
ChrisRackauckas: Why is this in docs?

ChrisRackauckas:
fold the defaultalg change plus that cacheval fix into this PR

AJ0070:
@ChrisRackauckas Updated the PR as suggested

PR #1197, opened 2026-08-10
Docs: document the LoadMKL_JLL preference in the FAQ

### LINKED ISSUE #524: Why does LinearSolve depend on MKL_jll
**Question❓**

I would prefer not to use MKL. I only use static arrays, and for static arrays MKL does not have any advantage. Can the dependency on MKL.jll be removed or be made optional?

Furthermore, MKL.jll spits out this warning:
```
Precompiling project...
  227 dependencies successfully precompiled in 99 seconds
  1 dependency had output during precompilation:
┌ MKL_jll
│   Downloading artifact: MKL
│  
│  [pid 100819] waiting for IO to finish:
│   Handle type        uv_handle_t->data
│   timer              0x1906150->0x7358aae8abc0
│  This means that a package has started a background task or event source that has not finished running. For precompilation to complete successfully, the event source needs to be closed explicitly. See the developer documentation on fixing precompilation hangs for more help.
│  
│  [pid 100819] waiting for IO to finish:
│   Handle type        uv_handle_t->data
│   timer              0x1906150->0x7358aae8abc0
│  This means that a package has started a background task or event source that has not finished running. For precompilation to complete successfully, the event source needs to be closed explicitly. See the developer documentation on fixing precompilation hangs for more help.
└  
```

AJ0070:
Fixes #524.

## Checklist

## Additional context

- Documentation only, one new FAQ entry, no source change.
- The question in #524 was whether the MKL dependency can be made optional. It can, through the `LoadMKL_JLL` preference, and that was answered in the issue thread, but the preference is not documented anywhere: it appears only in `src/LinearSolve.jl` and had no mention in `docs/`.
- The entry says why MKL is there, gives the `Preferences.set_preferences!` call with the LinearSolve UUID, and notes that the preference is read at load time so Julia has to be restarted.
- Verified rather than transcribed: setting the preference in a scratch environment and reloading in a fresh process flips `LinearSolve.usemkl` to `false`, and the default algorithm falls back from MKL to `LUFactorization` with solves still correct.
- Two behaviours worth documenting turned up while reading the source, and both are in the entry. The default is already architecture aware, x86_64 and i686 only, and off by default on AMD EPYC. And the preference controls whether MKL_jll is loaded and used, not whether it is installed: it stays a declared dependency. That second point matters because the reporter's actual complaint was the artifact download during precompilation, which this does not remove.
- Left as a plain fenced block rather than `@example`, so the docs build does not execute `set_preferences!` during CI.
- Tests box is unchecked because there is no test surface for a docs page. The recipe itself was verified as described above.

AI Disclosure: Used Opus 5

### DIFF
--- a/docs/src/basics/FAQ.md
+++ b/docs/src/basics/FAQ.md
@@ -87,3 +87,36 @@ Pr = LA.Diagonal(weights)
 prob = LS.LinearProblem(A, b)
 sol = LS.solve(prob, LS.KrylovJL_GMRES(precs = Returns((Pl, Pr))))
 ```
+
+## Why does LinearSolve.jl depend on MKL_jll, and how do I stop it from loading?
+
+MKL is the fastest BLAS on most x86 hardware, often by a wide margin, so LinearSolve.jl
+ships it by default and lets the default algorithm choice pick `MKLLUFactorization`
+where it wins. Without it, most installations end up substantially slower.
+
+If you would rather not load it, for example because you only solve small static-array
+systems where it brings nothing, set the `LoadMKL_JLL` preference to `false`:
+
+```julia
+using Preferences, UUIDs
+
+Preferences.set_preferences!(
+    UUID("7ed4a6bd-45f5-4d41-b270-4a48e9bafcae"),  # LinearSolve
+    "LoadMKL_JLL" => false; force = true
+)
+```
+
+The preference is read when LinearSolve.jl loads, so restart Julia afterwards and let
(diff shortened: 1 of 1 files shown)


AJ0070:
Not sure this one is necessary. If you read #524 as an individual question rather than a documentation gap worth covering, feel free to close this.

ChrisRackauckas:
This is a very good FAQ to add though. The hooks for how to disable it weren't documented but really should've been.

Issue #487, opened 2024-03-29
Direct and Iterative eigensolvers


dynamic-queries:
**Is your feature request related to a problem? Please describe.**
As @ChrisRackauckas  pointed out, algorithms for the new problems class 
(EigenProblems,EigenSolutions) can live in this package.

**Additional context**
Eventually, a ModelingToolkit.jl interface that services different preconditioners with iterative eigensolvers (that would eventually live here) should be implemented.

PR #1198, opened 2026-08-10
Eigenvalue: add the ArnoldiMethodJL keyword constructor

AJ0070:
## Checklist

## Additional context

- `ArnoldiMethodJL` is exported but had no keyword constructor, only the positional `ArnoldiMethodJL(::NamedTuple)`. Its siblings all have one, so this adds the matching `ArnoldiMethodJL(; kwargs...)`.
- Using this backend requires `using ArnoldiMethod`, which binds `ArnoldiMethod` to the module, so a bare `ArnoldiMethod(; kwargs...)` then fails with `objects of type Module are not callable`. With no keyword constructor on the JL name either, a `using LinearSolve` user had no working unqualified spelling.
- Nothing renamed or removed. `LinearSolve.ArnoldiMethod(...)` still works and the two produce identical objects, which the new test asserts.
- Every existing assertion in `test/Core/eigenvalue.jl` uses the qualified form, which is why this was never caught. The new test uses the unqualified one.
- Verified on a 50x50 SPD matrix: the new constructor forwards kwargs and its eigenvalues match `eigvals` to 8 significant digits. `test/Core/eigenvalue.jl` passes, Runic clean.
- Found while checking #487, but this does not close it: the eigensolvers are already implemented, and the ModelingToolkit preconditioner interface that issue also asks for is not built.

AI Disclosure: Used Opus 5

### DIFF
--- a/src/eigenvalue.jl
+++ b/src/eigenvalue.jl
@@ -41,13 +41,22 @@ end
 ArpackJL(; kwargs...) = ArpackJL((; kwargs...))
 
 """
-    ArnoldiMethodJL
+    ArnoldiMethodJL(; kwargs...)
 
-Algorithm type constructed by [`ArnoldiMethod`](@ref); see its docstring for details.
+Same solver as [`ArnoldiMethod`](@ref); see its docstring for details.
+
+Prefer this spelling when the ArnoldiMethod.jl package is in scope. Loading that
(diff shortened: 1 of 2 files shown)


ChrisRackauckas (approved):

AJ0070:
@ChrisRackauckas

PR #1199, opened 2026-08-10
Initialize u0 with a typed zero rather than false

### LINKED ISSUE #173: Support for Unitful quantities?
Is this expected to work? (related to https://github.com/PainterQubits/Unitful.jl/issues/46):

```
julia> using LinearSolve, Unitful

julia> A = rand(3,3)u"nm"; x = rand(3)u"nm"
3-element Vector{Quantity{Float64, 𝐋, Unitful.FreeUnits{(nm,), 𝐋, nothing}}}:
 0.8842061361482864 nm
 0.0439747397826753 nm
 0.6721916403320359 nm

julia> solve(init(LinearProblem(A, x)))
ERROR: DimensionError: nm and false are not dimensionally compatible.
Stacktrace:

```

AJ0070:
## Checklist

## Additional context

- `__init_u0_from_Ab` filled the default `u0` with `fill!(u0, false)`. `false` is a strong zero for the usual number types but is not a valid element of every array LinearSolve accepts, so `init` threw before reaching any solver. These three sites now use `zero(eltype(u0))`.
- Reported against Unitful in #173, where `init(LinearProblem(A, x))` on a `u"nm"` matrix failed with `DimensionError: nm and false are not dimensionally compatible`. Confirmed in isolation: `fill!(x, false)` throws on a `Quantity` vector while `fill!(x, zero(eltype(x)))` succeeds. `init` works after this change.
- No behaviour change for standard element types: `zero(T)` and `false` fill identically. Checked that `u0` is still all zeros, that a `Float32` problem keeps a `Float32` `u0`, and that solutions still match backslash.
- This does not make Unitful solves work, and does not close #173. Every solver path is blocked upstream: `\\`, `lu`, `qr` and `svd` all fail on a Unitful matrix in LinearAlgebra itself, Krylov.jl cannot build a workspace for `Quantity` element types, and SimpleGMRES hits dimensionally inconsistent tolerance arithmetic. Only `inv(A) * x` works. What changes here is that LinearSolve is no longer the first thing to fail.
- Full `GROUP=Core` suite passes with zero failures, Runic clean. The existing suite covers the changed path on every solve, since this is the default `u0` for any problem constructed without one.

AI Disclosure: Used Opus 5

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -195,6 +195,7 @@ StaticArrays = "1.9"
 StaticArraysCore = "1.4.3"
 SuperLUDIST = "1"
 Test = "1.10"
+Unitful = "1"
 Zygote = "0.7"
 blis_jll = "0.9.0"
 julia = "1.10"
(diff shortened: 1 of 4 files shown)


ChrisRackauckas:
Needs a test of what this actually helps. I wouldn't think this does much.

AJ0070:
This does add Unitful as a test dependency. It is stdlib only, but happy to drop the test if you would rather not carry it.

AJ0070:
Question before I go further on #173, since the eager default cacheval is the remaining blocker.

`_init_default_cacheval` builds a cacheval for every `DefaultAlgorithmChoice`, so `qr_instance` runs a real QR before the chosen algorithm is ever used. That throws for a dimensionful eltype, so a Unitful `Diagonal` never reaches `DiagonalFactorization` even though that solver handles it fine.

Making the slots lazy looks invasive: `_do_qr_fallback` needs the prebuilt `QRFactorizationPivoted` slot, and the slot types are fixed at construction. The narrower option is to emit `nothing` for factorization slots the eltype cannot support, the way the Krylov slots already are for non dense/sparse `A`. The QR fallback is impossible for those eltypes anyway.

Worth doing, or would you rather leave it?

ChrisRackauckas:
It is very invasive, and unitful doesn't even support linear algebra anyways, so I'm not sure this makes sense.

AJ0070:
Closing this. Dense LU and QR for dimensionful eltypes are an upstream LinearAlgebra limitation, so this is not worth carrying here.

PR #1200, opened 2026-08-11
UMFPACK: expose the control settings as a keyword

### LINKED ISSUE #383: How to set UMFPACK control vector ?

AJ0070:
## Checklist

## Additional context

- Closes #383. Adds a `control` keyword to `UMFPACKFactorization` taking any of the 16 control-vector entries SparseArrays exposes, named as SuiteSparse names them (`UMFPACK_IRSTEP` becomes `irstep`).
- `UMFPACKFactorization(control = (; irstep = 2, pivot_tolerance = 0.5))`.
- A NamedTuple rather than the raw vector: the caller never needs the layout, entries that are not named keep whatever SparseArrays defaults them to, it does not depend on the matrix eltype, and an unknown name errors at construction instead of being a silently ignored index. Happy to also accept a plain vector if that is wanted.
- Defaults to `(;)`, so existing behaviour is unchanged. Iterative refinement is the motivating case: SuiteSparse defaults it to 2 steps and SparseArrays disables it (JuliaLang/julia#122), so it has to be asked for.
- The vector is built from `get_umfpack_control` rather than edited in place. A hand-built vector drops the other defaults, and the `Float64`/`Int` cacheval is the shared `PREALLOCATED_UMFPACK`, so mutating its `control` would leak settings into every other cache. Tested that no leak occurs.
- `lu!` takes no `control`, so settings ride in on a fresh `lu` and the cached-reuse branch inherits them from that factorization. Tested by updating the matrix and re-solving.
- Tests cover each of the 16 names landing on its own entry and disturbing no other, and assert the name list and the index mapping cover each other so they cannot drift.
- Measured on an ill-conditioned sparse system: relative error 3.3e-14 with refinement off, 6.9e-15 with two steps.
- The 16 `SparseArrays.UMFPACK` control names are non-public, so they are added to the ExplicitImports inventory next to the existing `UMFPACK_OK` entry.
- Core and QA groups pass locally.

AI Disclosure: Used Opus 5


src/sparsearrays.jl
```diff
@@ -329,6 +329,39 @@ end
...
+    # the shared `PREALLOCATED_UMFPACK` const, and mutating it would leak the
+    # setting into every other cache. Starting from `get_umfpack_control` also
+    # keeps SparseArrays' remaining defaults, which a hand-built vector would drop.
+    function _umfpack_control(
+            alg::UMFPACKFactorization, ::Type{Tv}, ::Type{Ti}
+        ) where {Tv, Ti}
+        steps = alg.max_iterative_refinement_steps
+        steps === nothing && return nothing
+        control = get_umfpack_control(Tv, Ti)
+        control[JL_UMFPACK_IRSTEP] = steps
```
ChrisRackauckas: Why not allow for all settings in control?

AJ0070: Done, `control` now takes any of the 16 settings by name: `UMFPACKFactorization(control = (; irstep = 2, pivot_tolerance = 0.5))`.

Went with a NamedTuple rather than the raw vector so callers do not need the layout, unnamed entries keep SparseArrays' defaults, and an unknown name errors at construction instead of being silently ignored. Happy to also accept a plain `Vector{Float64}` if you want the escape hatch.

AJ0070:
Currently working on ci fix

PR #1206, opened 2026-08-11
Krylov: support an ArrayPartition right-hand side

### LINKED ISSUE #384: RecursiveArrayTools.jl Support
Is there any way to use `LinearSolve.jl` algorithms with `RecursiveArrayTools.ArrayPartition`? Concrete example:

```julia
using DifferentialEquations
f!(du, u, p, t) = du .= 1.01 .* u
u0 = [0.25, 0.5, 0.5, 0.75]
tspan = (0.0, 1.0)
prob = ODEProblem(f, u0, tspan)
sol = solve(prob, KenCarp47(linsolve = KrylovJL_GMRES()))
```

Fails with:
```julia
ERROR: LoadError: MethodError: no method matching ArrayPartition{Float64, Tuple{Vector{Float64}, Vector{Float64}}}(::UndefInitializer, ::Int64)

Closest candidates are:
  (::Type{ArrayPartition{T, S}} where {T, S<:Tuple})(::Any)
   @ RecursiveArrayTools ~/.julia/packages/RecursiveArrayTools/X30HP/src/array_partition.jl:27

<Stacktrace truncated>
```

Which - at a first glance - does not seem too horrible to fix. But would this be performant?

AJ0070:
## Checklist

## Additional context

- Closes #384. `KrylovJL_*` on an `ArrayPartition` right-hand side failed before any iteration ran, which is what the issue's `KenCarp47(linsolve = KrylovJL_GMRES())` example hits. The factorization algorithms already worked.
- Krylov.jl allocates its workspace as `S(undef, n)` with `S = typeof(b)`, which an `ArrayPartition` cannot provide: it is stored as several separate arrays, so there is no way to know how to split `n` across them.
- `Krylov.KrylovConstructor` builds the workspace with `similar` instead, so the solve runs on the partitioned vectors directly: no flattening, no copying, and `u` comes back in the type it went in as. Pointed out by @amontoison. It is present in every Krylov 0.10.x, so the existing compat bound already covers it and this PR adds no dependency and changes no `Project.toml` entry.
- `solve!` needs no specialization: the workspace holds the caller's array type throughout, so the generic path writes the answer straight into it.
- There was already a method meant to dodge this, but it declared no `zeroinit`, so the `solve!` path (which always passes `zeroinit = false`) never reached it. Adding the keyword alone would not have been enough: `solve!` hands the cacheval straight to Krylov as the workspace, so returning `nothing` just fails later.
- Tests cover five Krylov algorithms, three uneven partitions, cached re-solves picking up a new `b` and a new `A`, and that the partitioned shape survives. They were confirmed to fail without the fix, with the original error from the issue.
- The FAQ says what the requirement actually is: `similar` support, rather than any particular interface.
- Core and QA groups pass locally, on Julia 1.10 and 1.12.


src/iterative_wrappers.jl
```diff
@@ -352,12 +352,33 @@ function init_cacheval(
...
+#
+# So the Krylov solve runs on flat copies and the result is written back into the
+# partitioned `u`. `A` is passed through untouched: an `ArrayPartition` and the flat
+# vector it flattens to have the same length, so a matrix or an operator sized for
+# one multiplies the other unchanged. See SciML/LinearSolve.jl#384.
+#
+# The previous method here returned `nothing` to dodge the workspace allocation, but
+# it declared no `zeroinit`, so the `solve!` path below (which always passes
+# `zeroinit = false`) never reached it and hit the error anyway.
+_krylov_flat(x::RecursiveArrayTools.ArrayPartition) = copyto!(
```
ChrisRackauckas: Instead of a special new function here, it should use the ArrayInterface.jl primitive for vector flattening.

src/iterative_wrappers.jl
```diff
@@ -438,6 +459,32 @@ function _krylov_warm_start!(workspace, cache, mode::WarmStart.T, M, atol, rtol)
...
+    copyto!(flat.u, cache.u)
+    if cache.isfresh
+        flat.A = cache.A
+        cache.isfresh = false
+    end
+    flat.Pl, flat.Pr = cache.Pl, cache.Pr
+    flat.abstol, flat.reltol, flat.maxiters = cache.abstol, cache.reltol, cache.maxiters
+
+    sol = SciMLBase.solve!(flat, alg; kwargs...)
+    copyto!(cache.u, sol.u)
```
ChrisRackauckas: This is making too strong of an assumption. It should use an ArrayInterface.restructure

ChrisRackauckas (changes requested):

src/iterative_wrappers.jl
AJ0070: @ChrisRackauckas second one applied, `restructure` is better than the `copyto!` here.

On the first: `safevec` returns an `AbstractVector` unchanged (`safevec(v::AbstractVector) = v`), and `ArrayPartition` is one, so it does not flatten and cannot build the workspace Krylov needs. There is no RecursiveArrayTools extension in ArrayInterface adding a method either.

Did you have a different primitive in mind? Otherwise the options are `Vector(b)`, which forces a CPU array and would break GPU-backed partitions, or the current `similar(first(b.x), length(b))` form which keeps the backing type. Happy to inline it rather than name it if the objection is to the helper itself.

Blocked on this one, the rest of the PR is green.

ChrisRackauckas: I guess this is a better fit to SciMLStructures.jl and its canonicalize framework.

AJ0070: Checked this. `canonicalize(::Tunable, p::AbstractArray) = vec(p), ArrayRepack(p), true`, and `vec` of an `ArrayPartition` returns the same `ArrayPartition`, so the buffer comes back unflattened. The repack half then errors:

```
The original type ArrayPartition{...} does not support the SciMLStructures interface via
the AbstractArray `repack` rules. No method exists to take in a regular array and construct
the parent type back. Please define the SciMLStructures interface for this type.
```

RecursiveArrayTools does not define `canonicalize` for `ArrayPartition`, so it falls back to the generic array method that assumes `vec` flattens. `ArrayRepack` itself calls `ArrayInterface.restructure`, which is what this PR already uses for the write-back.

Happy to add the SciMLStructures interface for `ArrayPartition` in RecursiveArrayTools and then simplify this to a canonicalize/repack pair. Should that land first, or go with the flat cache here and switch over once it does?

ChrisRackauckas: > Happy to add the SciMLStructures interface for ArrayPartition in RecursiveArrayTools and then simplify this to a canonicalize/repack pair. Should that land first, or go with the flat cache here and switch over once it does?

Yes let's do that. And then we should document that any type that wants to be compatible with this should define the SciMLStructures.jl interface. That makes it a general solution.

amontoison:
A few months ago, I added this constructor in Krylov.jl such that we can use `similar` instead of `S(undef, n)` during the creation of the workspace:
https://jso.dev/Krylov.jl/dev/inplace/#Krylov.KrylovConstructor

AJ0070:
@amontoison thanks, that is much better than what this had. Rewritten to build the workspace with `KrylovConstructor`, so the solve runs on the partitioned vectors directly: no flattening, no copying, and `u` comes back in the type it went in as. It is present in every 0.10.x, so the existing `Krylov = "0.10"` compat already covers it.

@ChrisRackauckas this replaces the SciMLStructures approach. That version canonicalized to a flat buffer and repacked, which worked but copied twice per solve and needed a SciMLStructures dependency plus a RecursiveArrayTools 4.4 bound. The Krylov constructor needs neither, so the diff is now four files with no `Project.toml` change at all.

SciML/RecursiveArrayTools.jl#643 still stands on its own: `ArrayPartition` genuinely lacked the SciMLStructures interface and the generic `repack` errored for it. It just turned out not to be what this needed.

The FAQ entry changed accordingly: the requirement is `similar` support rather than a particular interface, which is both simpler and broader.

amontoison:
It is better to use `S(undef, n)` when it is supported because we avoid (in Krylov.jl) to allocate "optional" vectors that are needed for regularization, warm-start or preconditioners with `S(undef, 0)`. We allocate these additional vectors on-the-fly in the workspace if we need them.

Another advantage is that I have an internal `ktypeof(b)` and if `b` is a sparse vector, I detect it and allocate the `KrylovWorkspace` with the most relevant dense vector type.
The most common case is `b isa SparseMatrixVector`.

I will not suggest to use `KrylovConstructor` by default for all array types. I added for advanced users that are sure that they don't have alternatives with their array types and require `similar`.

One constructor is for performance while the other one is for portability.

AJ0070:
That matches what this does. `KrylovConstructor` is used only for `b::ArrayPartition`, where `S(undef, n)` has no valid answer, since the partitions are separate arrays and there is no way to know how to split `n` across them. The generic `init_cacheval` is untouched, so `Vector`, a sparse `b`, and everything else keep the `KS(A, b)` path with `ktypeof` picking the storage and the optional vectors staying zero length.

Three methods now: the generic one, the batched matrix one, and the `ArrayPartition` one. So it is the portability constructor only where there is no alternative, as you describe, and the `ArrayPartition` case does give up the on-the-fly optional vectors in exchange for working at all.

amontoison:
Great! I think it is the current best solution.

ChrisRackauckas:
> @ChrisRackauckas this replaces the SciMLStructures approach. That version canonicalized to a flat buffer and repacked, which worked but copied twice per solve and needed a SciMLStructures dependency plus a RecursiveArrayTools 4.4 bound. The Krylov constructor needs neither, so the diff is now four files with no Project.toml change at all.

It shouldn't allocate at all. You only do it once in the `init` phase, and then in the solve you only use `replace!`. Then it's non-allocating. 

> Another advantage is that I have an internal ktypeof(b) and if b is a sparse vector, I detect it and allocate the KrylovWorkspace with the most relevant dense vector type.
> The most common case is b isa SparseMatrixVector.

This does not seem equivalent at all. What does it do on an MTKParameters object? @AJ0070 test it with this object: https://sciml.github.io/SciMLStructures.jl/stable/example/ . The `SubproblemParameters` is the kind of thing we'd want to support for adjoints, and I am not sure this is going to work because Krylov.jl does not do the destructuring.

ChrisRackauckas:
> That matches what this does. KrylovConstructor is used only for b::ArrayPartition, where S(undef, n) has no valid answer, since the partitions are separate arrays and there is no way to know how to split n across them. The generic init_cacheval is untouched, so Vector, a sparse b, and everything else keep the KS(A, b) path with ktypeof picking the storage and the optional vectors staying zero length.

You mean an `Array` path, `AbstractArray` path, and a `isscimlstructures` path? That's what I'd expect this to land on.

src/iterative_wrappers.jl
```diff
@@ -352,12 +352,61 @@ function init_cacheval(
     return solver
 end
 
-# Krylov.jl tries to init with `ArrayPartition(undef, ...)`. Avoid hitting that!
+# Krylov.jl allocates its workspace as `S(undef, n)` with `S = typeof(b)`, which an
+# `ArrayPartition` cannot provide: it is stored as several separate arrays, so there
+# is no way to know how to split `n` across them, and its constructor says so rather
+# than guessing. `Krylov.KrylovConstructor` builds the workspace with `similar`
+# instead, which such a type does support, so the solve runs on the partitioned
+# vectors directly with no flattening and no copying.
```
ChrisRackauckas: The general type does not support `similar`

AJ0070:
Tested the `Parameters` object from the docs example. You were right, and my "replaces" framing was wrong.

```
similar(p)                               MethodError
length(p)                                MethodError
GmresWorkspace(KrylovConstructor(p, p))  MethodError: no method matching zero(::Type{Any})
canonicalize -> solve flat -> repack     works, repacks with subparams and coeffs intact
```

`KrylovConstructor` needs `similar` and the array interface, so it covers `ArrayPartition` but not a non-array structure. The two are complementary rather than alternatives. Fair on the allocation too: canonicalizing once in `init` and using `replace!` in the solve is non-allocating, and my earlier version copying per solve was my mistake rather than the approach's. I have corrected the comment that claimed `similar` support in general.

One thing worth deciding before I build the third path. It cannot be an `init_cacheval` method next to the other two, because `init` fails before any algorithm dispatch:

```
zero(::Type{Any})  <-  real(T::Type)  <-  __init(...)
```

`eltype(p)` is `Any`, so `real(eltype(b))` throws while computing tolerances, well before `init_cacheval` is reached. So the structure path has to canonicalize at the top of `__init`, carry the repack on the cache, and repack when the solution is built. That is core `common.jl` work affecting every algorithm, not just `KrylovJL`, which is arguably the right place since a structured `b` is broken for all of them today, not only Krylov.

So the three paths are not the same shape: `Array` and `AbstractArray` are `init_cacheval` methods, the structure one is an init-level change.

Do you want that in this PR, or land the two array paths here for #384 and do the structure path separately? Happy either way, just checking since it widens the blast radius from the Krylov wrapper to the shared init path.

ChrisRackauckas:
I also don't think this is something Krylov specific at all: it should be done at a higher level. At least the SciMLStructures interfacing. So it's probably fine to finish this PR as a limited Krylov-specific abstractarray thing, but we should then bubble through more general structure support.

AJ0070:
@ChrisRackauckas this is at the scope you described, a Krylov-specific `AbstractArray` fix, with the general structure support tracked separately in #1208.

Everything raised is in:

- `KrylovConstructor` only for `b::ArrayPartition`. The generic `init_cacheval` is untouched, so `Vector`, a sparse `b` and batched matrices keep the `KS(A, b)` path with `ktypeof` picking the storage.
- The comment no longer claims `similar` support in general.
- The FAQ names the SciMLStructures interface as the general answer for a right-hand side that is not an array, pointing at #1208, rather than only mentioning `similar`.

Could you clear the changes-requested when you get a chance?

ChrisRackauckas (approved):

AJ0070:
Thanks @ChrisRackauckas and @amontoison for the reviews on this one.

The result ended up a good deal smaller than where it started. @amontoison pointing at `KrylovConstructor` removed the flattening and the copying entirely, and the earlier push toward SciMLStructures turned up two things worth having on their own: SciML/RecursiveArrayTools.jl#643, since `ArrayPartition` genuinely lacked the interface, and SciML/SciMLStructures.jl#79, where `isscimlstructure` opts in numeric arrays whose `repack` then errors.

Also useful was the request to test an `MTKParameters`-shaped object, which is what established that the array and structure cases need different routes rather than one replacing the other. That is now #1208, and I have a working implementation of it in progress.

PR #1207, opened 2026-08-12
Add a ConjugateGradients.jl extension

### LINKED ISSUE #567: Wrap ConjugateGradients.jl
https://github.com/mcovalt/ConjugateGradients.jl provides CG and BiCGStab. In its README.md, it says

> There are a few great iterative solver packages available for Julia: [IterativeSolvers.jl](https://github.com/JuliaMath/IterativeSolvers.jl), [KrylovMethods.jl](https://github.com/lruthotto/KrylovMethods.jl), and [Krylov.jl](https://github.com/JuliaSmoothOptimizers/Krylov.jl). These are all very well rounded and complete packages.
>
> This package, ConjugateGradients.jl, is built around reducing allocations as much as possible for a particular type of problem. As far as I know, if your program will be using an iterative solver within another iterative process, this module will result in less allocations compared to the previously mentioned packages.

I haven't tested it, but this sounds at least interesting enough to provide wrappers in LinearSolve.jl.

AJ0070:
## Checklist

## Additional context

- Closes #567. Adds `ConjugateGradientsJL_CG` and `ConjugateGradientsJL_BICGSTAB`, with `ConjugateGradientsJL(; solver)` behind them, as a package extension.
- Three details of the ConjugateGradients.jl API are handled and commented, since none are apparent from its signatures: the operator is applied as `A(output, input)` rather than by multiplication; `precon(z, r)` is called unconditionally and defaults to `copy!`, so the no-preconditioner case passes `copy!` rather than `nothing`; and the solvers are defined for `Vector{<:Real}` only.
- The element and array type limits are checked at `init` and point at `KrylovJL_*` instead, rather than surfacing as a `MethodError` from inside the solver.
- A right preconditioner is refused rather than ignored. ConjugateGradients takes a single left `precon`, so accepting a right one would silently solve a different problem.
- `CGData`/`BiCGStabData` hold the solver's working vectors, so they serve as the cacheval and are reused across solves.
- Tests cover both solvers, the generic constructor, BiCGStab on a nonsymmetric system, cached re-solves with a new `b` and a new `A`, a left preconditioner being applied, a right one being refused, and the rejection messages naming the alternative.
- Core and QA groups pass locally.

AI Disclosure: Used Opus 5


ext/LinearSolveConjugateGradientsExt.jl
```diff
@@ -0,0 +1,119 @@
...
+            ArgumentError(
+                "ConjugateGradientsJL supports a left preconditioner only; got a right " *
+                    "preconditioner of $(typeof(cache.Pr))."
+            )
+        )
+    end
+    # `precon(z, r)` is called unconditionally, so the no-preconditioner case is
+    # ConjugateGradients' own default of `copy!` rather than `nothing`.
+    precon = if LinearSolve._isidentity_struct(cache.Pl) || cache.Pl === nothing
+        copy!
```
ChrisRackauckas, line 91: Weird, but if that's what it wants 🤷

PR #1212, opened 2026-08-14
Carry the forward preconditioners into the adjoint solve

### LINKED ISSUE #476: Using preconditioners in the adjoints of a linear solve
From @avik-pal https://github.com/SciML/LinearSolve.jl/pull/449

>  Preconditioning. These need to be added to the Adjoint Sensitivity Struct. Can we use a left preconditioner for the forward problem as the transpose right preconditioner? Need someone with a better grasp of linear algebra to chime in here

@amontoison I think some Krylov methods only support pre-conditioners on one side though?

Maybe we need some way for someone to specify adjoint preconditioners?

AJ0070:
## Checklist

## Additional context

Fixes #476, and removes the `# TODO: Preconditioners?` at the top of `src/adjoint.jl`.

- The adjoint solve was built as a bare `LinearProblem(adjoint(A), b)`, so a preconditioner passed as `Pl`/`Pr` was dropped and the adjoint ran unpreconditioned.
- The rule takes `.u` regardless of the return code, so on a system that needs preconditioning the adjoint stopped at `maxiters` and the gradient came back wrong rather than slow:

  | n | adjoint before | adjoint after |
  | --- | --- | --- |
  | 64 | MaxIters, 64 iters, 4.0e-4 relerr | Success, 5 iters, 5.7e-9 |
  | 256 | MaxIters, 256 iters, 1.2e-3 relerr | Success, 5 iters, 3.4e-9 |
  | 512 | MaxIters, 512 iters, 1.2e-3 relerr | Success, 5 iters, 1.0e-9 |

- Left preconditioning the forward system solves with `Pl \ A`, whose adjoint is `A' / Pl'`, that is the adjoint system right preconditioned by `Pl'`. So the two sides swap and each is conjugated, which is what the TODO guessed.
- Methods restricted to centered preconditioning have no right slot to swap into and would silently discard it, so for those the forward left preconditioner stays on the left. Per @amontoison in the issue, those apply to symmetric systems where `A' = A`, so the same preconditioner is the right one to reuse.
- An algorithm carrying its own `precs` is left alone, since `precs` is evaluated against the matrix the adjoint problem is initialized with and already rebuilds the pair for `A'`. Deriving a second pair applied both at once and measured worse than the original bug.
- `LinearSolveAdjoint` gains `Pl` and `Pr` to override either side, which answers the "some way to specify adjoint preconditioners" question in the issue and covers a matrix-free preconditioner with no usable `adjoint`.
- Applied in the ChainRulesCore, Mooncake, and Enzyme rules.
- Separately noticed while reading this code: the Krylov branch of `defaultalg_adjoint_eval` refers to `cache.val.abstol` and returns the solution rather than `.u`. Both callers pass a `LinearCache`, which has no `val`, so that branch looks like it would throw if reached. I could not construct a default selection that reaches it, so it is left alone here.

AI Disclosure: Used Opus 5

### DIFF
--- a/ext/LinearSolveChainRulesCoreExt.jl
+++ b/ext/LinearSolveChainRulesCoreExt.jl
@@ -55,6 +55,8 @@ function CRC.rrule(
         ∂∅ = NoTangent()
 
         ∂u = hasproperty(∂sol, :u) ? ∂sol.u : ∂sol
+        adj_alg = sensealg.linsolve === missing ? alg : sensealg.linsolve
+        adj_Pl, adj_Pr = LinearSolve._adjoint_precs(adj_alg, sensealg, cache.Pl, cache.Pr)
         if sensealg.linsolve === missing
             cached_adjoint_solution = LinearSolve._adjoint_factorization_solve(
                 alg, cache.cacheval, cache.A, ∂u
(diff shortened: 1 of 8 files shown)


ChrisRackauckas (approved):

AJ0070:
@ChrisRackauckas

PR #1213, opened 2026-08-14
Copy the cache where it changes rather than on every solve

### LINKED ISSUE #380: Improve Enzyme rules with less copying
Right now in order to make the first pass of the Enzyme rule there's a lot of defensive copying. I think we can get away with less by putting a good rule over `setproperty!(linsolve, ...)`, since if it knows just when to copy there, it can ensure that the `linsolve` in reverse is "fixed" as it goes backwards. That would mean that the other rules would not need to copy each time in case of a change, since we would instead be public API point where it can change.

AJ0070:
## Checklist

## Additional context

Fixes #380, following the approach in the issue.

- The `solve!` augmented primal took a `deepcopy` of the whole `LinearCache` on every solve, to guard the reverse pass against a later mutation. That was about 29% of the allocations of a reverse-mode solve, independent of size.
- `setproperty!` is the public way to change a cache between solves, so it is now the one place that copies. Its rules record what an assignment overwrote and restore it in reverse, which walks the cache backwards into the state each earlier `solve!` ran against, so `solve!` can keep the live cache.
- Only an assignment to `A` copies anything, since that is what invalidates the factorization the next `solve!` builds over. A cache solved repeatedly against a fixed `A` now copies nothing.

Reverse-mode allocations, `init` plus `solve!` with `LUFactorization`, Enzyme 0.13.199 on Julia 1.12:

  | n | before | after | change |
  | --- | --- | --- | --- |
  | 50 | 149 KB | 106 KB | -28.7% |
  | 200 | 2.26 MB | 1.62 MB | -28.6% |
  | 500 | 14.05 MB | 10.04 MB | -28.6% |

- Gradients match ForwardDiff to machine precision for a plain solve and for caches whose `b`, `A`, or both are replaced between solves. Those four patterns are the added tests.
- The existing batched (`BatchDuplicated`) rules and the "Adjoint case currently not handled" check both still pass, which are the paths the new shadow handling touches.

AI Disclosure: Used Opus 5

### DIFF
--- a/ext/LinearSolveEnzymeExt.jl
+++ b/ext/LinearSolveEnzymeExt.jl
@@ -621,6 +621,16 @@ function EnzymeRules.reverse(
     return (nothing, nothing)
 end
 
+"""
+    _shadow_of_constant(x)
+
+Shadow value to store alongside a constant assigned into a cache. A constant carries no
+derivative, so an array shadow is zeroed; anything else is a field the shadow cache holds
+verbatim (the staleness flags, the algorithm, a cacheval), and is passed through.
+"""
+_shadow_of_constant(x::AbstractArray) = zero(x)
+_shadow_of_constant(x) = x
+
 # `setproperty!` is the public way to change a `LinearCache` between solves, so it is the
 # one place that has to copy. Reverse restores what the assignment overwrote, which walks
 # the cache backwards into the state each earlier `solve!` ran against, and that is what
@@ -644,7 +654,7 @@ function EnzymeRules.augmented_primal(
     if !(cache isa Const)
         dcaches = EnzymeRules.width(config) == 1 ? (cache.dval,) : cache.dval
         dxs = if x isa Const
-            ntuple(_ -> EnzymeCore.make_zero(x.val), Val(EnzymeRules.width(config)))
+            ntuple(_ -> _shadow_of_constant(x.val), Val(EnzymeRules.width(config)))
         elseif EnzymeRules.width(config) == 1
             (x.dval,)
         else
--- a/ext/LinearSolveEnzymeExt.jl
+++ b/ext/LinearSolveEnzymeExt.jl
@@ -621,6 +621,60 @@ function EnzymeRules.reverse(
     return (nothing, nothing)
 end
 
+# `setproperty!` is the public way to change a `LinearCache` between solves, so it is the
+# one place that has to copy. Reverse restores what the assignment overwrote, which walks
+# the cache backwards into the state each earlier `solve!` ran against, and that is what
+# lets the `solve!` rule hold the live cache instead of copying it defensively every time.
+# See https://github.com/SciML/LinearSolve.jl/issues/380.
+function EnzymeRules.augmented_primal(
+        config, func::Const{typeof(Base.setproperty!)}, ::Type{RT},
+        cache::EnzymeCore.Annotation{<:LinearSolve.LinearCache},
+        name::Const{Symbol}, x::EnzymeCore.Annotation
+    ) where {RT}
+    field = name.val
+    old = getfield(cache.val, field)
+    # Assigning any of these also flips the staleness flags, so they rewind together.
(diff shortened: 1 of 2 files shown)


AJ0070:
@ChrisRackauckas this one is ready for review when you have a moment.

The four red checks are all pre-existing on main:

- BoundaryValueDiffEq and ModelingToolkit downstream fail on main as well.
- Core on julia 1 and lts fails at `test/Core/genericlu_naive_ldiv.jl:122`, same line and same 503 passed / 1 failed on a plain main run, so it is not from this PR. Worth a separate look: that check asserts at least one of 32 random draws differs bitwise from `ldiv!`, and the comment puts bit agreement at 0.13% per draw, so all 32 agreeing suggests the naive path stopped being taken rather than an unlucky run. Happy to open an issue for it.

Enzyme and Mooncake rule tests and QA are green, and reverse-mode allocations drop 28.6%.

ext/LinearSolveEnzymeExt.jl
```diff
@@ -675,9 +739,9 @@ function EnzymeRules.augmented_primal(
         end
     end
 
-    cachesolve = deepcopy(linsolve.val)
-
-    cache = (copy(res.u), resvals, cachesolve, dAs, dbs)
+    # No copy: the `setproperty!` rules above rewind the cache as the tape unwinds, so by
+    # the time this solve's reverse runs the live cache is back to what this solve saw.
+    cache = (copy(res.u), resvals, linsolve.val, dAs, dbs)
```
ChrisRackauckas, line 744: there is still a copy?

ext/LinearSolveEnzymeExt.jl
```diff
@@ -621,6 +621,70 @@ function EnzymeRules.reverse(
...
+        name::Const{Symbol}, x::EnzymeCore.Annotation
+    ) where {RT}
+    field = name.val
+    old = getfield(cache.val, field)
+    # Assigning any of these also flips the staleness flags, so they rewind together.
+    old_isfresh = getfield(cache.val, :isfresh)
+    old_precsisfresh = getfield(cache.val, :precsisfresh)
+    # A new `A` is what invalidates the factorization, and the next `solve!` builds the
+    # replacement over the same cacheval, so this is the assignment that has to copy.
+    old_cacheval = field === :A ? deepcopy(getfield(cache.val, :cacheval)) : nothing
```
ChrisRackauckas, line 651: does it need to deepcopy?

ext/LinearSolveEnzymeExt.jl
AJ0070: Yes, one is left, but it is the solution vector rather than the cache, so O(n) instead of O(n^2). At n = 500 it is 4.1 KB against the 4.0 MB whole cache copy it replaces.

It cannot go through the same mechanism: `solve!` writes the solution into `cache.u` directly rather than through `setproperty!`, so the rewind never sees it, and the reverse pass needs that solution for `dA -= z y'`. Dropping it makes `dA` wrong on any second solve (0.2 relative error on the two solve case here), while `db` stays correct since it only uses `z`.

Removing it would mean rewinding `u` in the `solve!` rule, which means storing the old `u`, which is the same copy.

ext/LinearSolveEnzymeExt.jl
AJ0070: It does. I tried a plain reference: gradients go to 0.4 relative error on the A replacement case and 0.61 when both `b` and `A` are replaced.

The factorization holds the cache-owned pivot vector, so the next factorization writes into storage the earlier one still points at, and the earlier factorization is destroyed even when the new `A` is a fresh array.

It only runs on an assignment to `A`, not per solve. At n = 500 it is 2.0 MB against the 4.0 MB whole cache copy that used to run on every solve, so a cache resolved against a fixed `A` now copies nothing beyond the solution vector.

AJ0070:
@ChrisRackauckas

PR #1215, opened 2026-08-15
Add LHLFactorization: O(n²) re-shift for repeated solves of I - γJ

ChrisRackauckas-Claude:
> **Draft — please ignore until reviewed by @ChrisRackauckas.**

## What and why

An implicit ODE/DAE solver refactorizes `W = I - γJ` (in OrdinaryDiffEq's W-transform, `J - M/(dt·γ)`) on nearly every step, though only the scalar `γ` has moved — adaptive step-size control changes `γ` constantly while `J` is held fixed for tens of steps.

`LHLFactorization` reduces `J` once to upper Hessenberg form by Gaussian similarity with partial pivoting (`J = Z H Z⁻¹`, `Z = D·P·L`). The shift never reaches `Z`:

```
W = I - γJ = Z (I - γH) Z⁻¹
```

and `I - γH` is Hessenberg, so **a new `γ` costs `O(n²)`** instead of `O(n³)`. The factorization itself lives in the new dependency-free [LHL.jl](https://github.com/SciML/LHL.jl); this PR is the LinearSolve algorithm, the split-matrix plumbing and the default-algorithm hook.

The system matrix is handed over unassembled as a `ShiftedJacobian` — a lazy `AbstractMatrix` equal to `I - γJ` — and the shift moves with `update_gamma!`. That falls back to invalidating the factorization for any other algorithm, so callers need no special case.

Companion PR: SciML/OrdinaryDiffEq.jl (linked below). **This must be registered before that one can pass CI**, since it bumps `OrdinaryDiffEqDifferentiation`'s `LinearSolve` compat to 5.11.

## Verification

`GROUP=Core julia --project -e 'using Pkg; Pkg.test()'`, green:

```
LHL Factorization   | 171  171   8.0s
Adjoint Sensitivity | 123  123   2m11.3s
Basic Tests         | 843  843   8m22.4s
Default Alg Tests   | 127  127   5m47.7s
```

The first run of that group **failed** one assertion, which is worth flagging because it is exactly the kind of thing a new factorization gets wrong: `test/Core/adjoint.jl` enumerates every `AbstractFactorization` leaf type and requires each to declare an adjoint-reuse policy, and `LHLFactorization` had not. It now declares `_NoAdjointFactorizationReuse` — the workspace is a Hessenberg reduction plus a shifted-Hessenberg LU, not a `Factorization` of the system matrix, so the adjoint reduces `Aᴴ` from scratch. `Wᴴ = Z⁻ᴴ(I-γH)ᴴZᴴ` is solvable from the same pieces; that is left for later.

Runic and typos clean on every touched file.

### Numbers

Per-phase cost, single-threaded BLAS, dense random `J` (`proto/bench_final.jl` in the session):

| n | LU refact | LHL shift | × | LHL reduce (once) | LU solve | LHL solve r=1 | r=0 |
|---|---|---|---|---|---|---|---|
| 100 | 8.66e-5 | 5.86e-6 | 15× | 2.30e-4 | 4.47e-6 | 1.88e-5 | 8.70e-6 |
| 400 | 2.17e-3 | 8.03e-5 | 27× | 1.21e-2 | 3.86e-5 | 1.82e-4 | 7.53e-5 |
| 800 | 1.37e-2 | 3.43e-4 | 40× | 9.26e-2 | 1.26e-4 | 7.27e-4 | 2.72e-4 |
| 1600 | 9.53e-2 | 2.11e-3 | 45× | 1.00e0 | 8.65e-4 | 5.10e-3 | 1.80e-3 |

End to end on a dense Brusselator through `NordsieckBDF` (see the companion PR), 2.6–3.7× at matched accuracy for `n = 800`.

## Stability — the part to push back on

`Z` is not orthogonal, so unlike an LU the backward error carries a factor `κ(Z)`. Two findings:

**Partial pivoting is mandatory**, not a tuning knob. Without it the method loses every digit on ordinary matrices — a random Gaussian matrix gives `1e-1` backward error and the Wilkinson growth matrix diverges (element growth `8e31`). It is always on; there is no switch.

**`κ(Z)` can be large.** An adversarial search over six matrix families (400 draws each) found `κ∞(Z)` reaching `8.4e10` on near-nilpotent matrices, costing eight digits. One step of fixed-precision iterative refinement removes it:

| matrix | κ(Z) | refine=0 | refine=1 | LU |
|---|---|---|---|---|
| randn | 2.3e3 | 1.3e-15 | 1.9e-16 | 4.5e-16 |
| Wilkinson growth | 9.9e3 | 1.4e-14 | 6.5e-17 | 1.9e-16 |
| **near-nilpotent** (worst of 200) | **7.5e10** | **5.3e-09** | **1.3e-16** | 4.0e-16 |
| clustered spectrum (worst of 200) | 3.7e3 | 1.0e-14 | 2.0e-16 | 5.4e-16 |

Hence `refine = 1` by default. **But `refine = 0` is the right setting inside a Newton loop**, and that is a judgment call worth challenging: refinement roughly triples a solve, and an implicit solver runs ~14 solves per `γ`, so it is charged against exactly the quantity being saved. Newton is itself a correction loop. Measured on the Brusselator, `refine = 0` roughly halves the run time at the same accuracy. The default stays at 1 because a bare linear solve has no outer loop to lean on — I can flip it if you'd rather optimize for the ODE caller.

Matrices notoriously bad for LU cause no trouble here: a Hessenberg LU's growth factor is bounded by `n` rather than `2ⁿ⁻¹`, because each elimination step chooses between two candidates. The whole exposure is in `Z`.

## Judgment calls

- **`defaultalg(::ShiftedJacobian)` returns `LHLFactorization` at `n ≥ 100`.** Constructing a `ShiftedJacobian` is itself the statement that the shift will move while `J` stays put — nothing else builds one — so the type is a reliable signal of intent. The cutoff is measured, not derived. Below it, and for a non-dense `J`, the fallback is `LUFactorization`, which is deliberate: a `ShiftedJacobian` is lazy and cannot be factorized in place, so the fallback has to be an algorithm with an allocating path. Handing one to `GenericLUFactorization` or `QRFactorization` raises a pointed `ArgumentError` from LHL.jl rather than a bare `CanonicalIndexError`.
- **`LHL` as a hard dependency** rather than an extension: it is `LinearAlgebra`-only and ~500 lines, and `defaultalg` cannot return an algorithm that might not be loaded.
- **Minor version bump** (5.10.0 → 5.11.0) for the new public API.

## Not verified

Other LinearSolve test groups, GPU paths, the docs build.

https://claude.ai/code/session_014CrTAi2YLkosEXXbudXyKf


ChrisRackauckas-Claude:
Companion PR: SciML/OrdinaryDiffEq.jl#4266 (blocked on this one being registered — it needs `LinearSolve = "5.11"`). New dependency: https://github.com/SciML/LHL.jl, which needs registering first of all.

ChrisRackauckas-Claude:
**Restructured** after review discussion — the split-W representation is now `SciMLOperators.WOperator` rather than a bespoke `ShiftedJacobian`, since `WOperator` already holds `J` and `gamma` apart and OrdinaryDiffEq already builds one.

What changed since the first push:

- **`ShiftedJacobian` is gone.** `LHLFactorization` takes a `WOperator` (or a plain matrix) as its system matrix. One concept instead of two overlapping ones, and no lazy-`AbstractMatrix` sharp edges.
- **`update_shift!` is gone**; `update_gamma!(cache, γ)` sets the `WOperator`'s `gamma`, which is the ODE convention (`W = J - M/γ`) rather than the `I - γJ` one.
- **`defaultalg(::WOperator)`** selects `LHLFactorization` when the Jacobian is dense, the mass matrix is a multiple of `I`, and `n ≥ 100`; otherwise it falls through with `@invoke` to the pre-existing `AbstractSciMLOperator` method, so matrix-free `WOperator`s keep their Krylov default. That fall-through is deliberate — materializing the operator there would break the matrix-free case.
- **New dependency name**: SciML/LHLFactorization.jl (was `LHL.jl`), and it now depends on SciML/SciMLOperators.jl#419 for the Jacobian generation counter.

Merge order is now: SciMLOperators#419 → LHLFactorization.jl registration → this → SciML/OrdinaryDiffEq.jl#4266.

Re-verified after the restructure — `GROUP=Core` green, and end-to-end on the Brusselator through `NordsieckBDF` at n=800, single-threaded BLAS, best of 3:

| tol | LU | LHL refine=1 | LHL refine=0 |
|---|---|---|---|
| 1e-6 | 1.296 | 0.690 (1.88×) | 0.389 (**3.33×**) |
| 1e-8 | 1.748 | 1.020 (1.71×) | 0.540 (**3.24×**) |
| 1e-10 | 2.027 | 1.798 (1.13×) | 0.784 (**2.58×**) |

Same final state to 1e-6 in every case.

ChrisRackauckas-Claude:
Follow-up: SciML/SciMLOperators.jl#419 now exposes a `jac_stale` flag rather than a generation counter, so `_lhl_needs_reduce` reads `jacobian_stale(W)` and the solve calls `mark_jacobian_current!` after reducing. The workspace correspondingly tracks only *whether* it holds a reduction (`ws.reduced::Bool`), not whose — that bookkeeping belongs to the consumer. Note the flag is shared, so one LHL cache per `WOperator`. Tests updated and green.

ChrisRackauckas-Claude:
Re-measured on the current HEAD, since the earlier tables were taken before the counter → `jac_stale` flag change. Brusselator, `NordsieckBDF`, single-threaded BLAS, best of 3:

| n | tol | LU | LHL r=1 | LHL r=0 | speedup (r=0) |
|---|---|---|---|---|---|
| 128 | 1e-6 | 0.0159 | 0.0194 | 0.0110 | 1.45× |
| 128 | 1e-8 | 0.0228 | 0.0305 | 0.0158 | 1.44× |
| 128 | 1e-10 | 0.0314 | 0.0545 | 0.0311 | 1.01× |
| 800 | 1e-6 | 1.2730 | 0.6985 | 0.3916 | **3.25×** |
| 800 | 1e-8 | 1.6325 | 1.0018 | 0.5338 | **3.06×** |
| 800 | 1e-10 | 2.0435 | 1.7732 | 0.7870 | **2.60×** |

Within run-to-run noise of the previous figures (3.33 / 3.24 / 2.58), as expected — the flag change is control flow only. Final state matches LU in every row.

src/lhl.jl
```diff
@@ -0,0 +1,257 @@
...
+
+`LHLWorkspace` deliberately records only *that* it holds a reduction, not whose — so the
+consumer has to track that itself. Two independent things can invalidate a reduction and
+each needs its own signal: swapping in a *different* matrix (caught by `===` on the
+Jacobian) and writing into the *same* one (caught by `jacobian_stale`, since the object is
+unchanged). Keying on only one of them silently reuses a stale reduction.
+"""
+mutable struct LHLCache{WS}
+    ws::WS
+    jac::Any
```
ChrisRackauckas: type unstable? Make this concrete?

src/default.jl
```diff
@@ -223,6 +223,40 @@ function defaultalg(A::Hermitian, b, ::OperatorAssumptions{Bool})
...
+    J = A.J isa MatrixOperator ? convert(AbstractMatrix, A.J) : A.J
+    if assump.issq && J isa DenseMatrix && size(A, 1) >= LHL_DEFAULT_MIN_SIZE &&
+            _lhl_scalar_massmatrix(A.mass_matrix)
+        # Refinement is the right default for a bare linear solve; a caller running an
+        # outer correction loop should ask for `LHLFactorization(refine = 0)`.
+        return LHLFactorization()
+    end
+    # Everything else — a matrix-free Jacobian, a general mass matrix, or a problem too
+    # small to pay for the reduction — keeps whatever the operator path already chose.
+    return @invoke defaultalg(A::SciMLOperators.AbstractSciMLOperator, b, assump)
```
ChrisRackauckas, line 252: Type unstable?

ChrisRackauckas-Claude:
CI triage on the three red checks:

**`tests / QA` — one sub-failure was mine, fixed in e811f3a.**
SciMLTesting flagged `No unapproved public reexports: isempty([:mark_jacobian_updated!])`. LinearSolve was re-exporting a name SciMLOperators declares public, making it a second undeclared owner. Removed the `export`; the import stays, and the docs (`SO.mark_jacobian_updated!`) and tests (`using SciMLOperators`) already reach it through the owner.

`test/Core/lhl.jl` after the removal, exit 0:
```
Test Summary:                                          | Pass  Total  Time
a different Jacobian is never served a stale reduction |    6      6  0.1s
Test Summary:                                                     | Pass  Total  Time
concrete factorizations refuse an externally-maintained WOperator |    3      3  0.4s
Test Summary:                                            | Pass  Total  Time
refinement recovers backward error on a hostile Jacobian |    2      2  0.3s
Test Summary: | Pass  Total  Time
adjoint solve |    2      2  0.4s
```

The remaining QA content in that job — an Aqua method-ambiguity report and `NonPublicQualifiedAccessException` for `_check_matrix_support` / `csrlsvqr!` in `LinearSolveCUDAExt` — reproduces on `main` and is not from this PR.

**`BoundaryValueDiffEq.jl/All` downstream — pre-existing.** Red on `main` as well.

**`ModelingToolkit.jl/All` downstream — not this PR.** Resolution succeeded (`SciMLOperators v1.28.0`, `LHLFactorization v2.0.0` both install fine). The failure is internal to MTK:
```
Got exception outside of a @test
UndefVarError: `Operator` not defined in `ModelingToolkit`
  [1] isolate_subsystem(sys::ModelingToolkitBase.System, ...)
    @ ModelingToolkit .../src/systems/analysis_points.jl:301
  [2] top-level scope
    @ .../lib/ModelingToolkitBase/test/analysis_points.jl:885
```
A missing symbol in MTK's own `analysis_points.jl`, on a code path this PR does not touch.

Not verified locally: the GPU/CUDA extension paths, and the downstream suites themselves (run against package `main`s, not pinned).

ChrisRackauckas-Claude:
Good catch on the fallback coverage — the answer was **no**, and chasing it turned up a real defect of mine.

## The gap
The three fallback assertions only checked *which algorithm `defaultalg` returns*. None of them called `solve`, none constructed a matrix-free `J` (the branch the code comment explicitly calls out), and the word `Pl` appeared nowhere in the file.

## What that hid
`_init_default_cacheval` gated the LHL slot on `A isa WOperator`, while `defaultalg` also required a dense `J`, a scalar mass matrix, and `n >= LHL_DEFAULT_MIN_SIZE`. Any `WOperator` failing one of the extra conditions allocated a full LHL workspace it could never use. Both now ask a single `_lhl_defaultable` predicate.

Discriminating check — reverting only the init gate to `A isa WOperator`:
```
branches that decline LHL still solve: Test Failed at test/Core/lhl.jl:306
  Expression: (init(LinearProblem(W, rhs))).cacheval.LHLFactorization === nothing
   Evaluated: LinearSolve.LHLCache{LHLWorkspace{Float64, Float64, Float64}, Matrix{Float64}}(... n=40 ...) === nothing
```
(fires for the n=40 general-mass-matrix case and again for the n=31 below-cutoff case). With the fix:
```
Test Summary:                                             | Pass  Total  Time
branches that decline LHL still solve                     |   11     11  3.8s
a split W keeps its preconditioners on an iterative solve |    4      4  1.2s
```

## On preconditioners specifically
`Pl`/`Pr` are **not** regressed by this PR, and I checked rather than assumed. The default polyalgorithm drops preconditioners for all of these inputs already — on a fallback branch this PR does not touch (general mass matrix -> Krylov):

| path | alg selected | Pl applications |
|---|---|---|
| `Wm` via default | `KrylovJL_GMRES` | 0 |
| `Wm` via bare `KrylovJL_GMRES()` | `KrylovJL_GMRES` | 21 |
| `Wd` via default | `LHLFactorization` | 0 |
| `Wd` via bare `KrylovJL_GMRES()` | `KrylovJL_GMRES` | 28 |

So swapping the default from Krylov to LHL for a dense-J split `W` costs no preconditioner that was being honoured. The new testset pins the part that does work: a split `W` solved with an explicit `KrylovJL_GMRES()` applies both `Pl` and `Pr` and returns the right answer.

## Pre-existing bug found on the way
A matrix-free `WOperator` cannot be solved through the default at all — `_init_default_cacheval` builds the `PureKLUFactorization` slot, which calls `convert(AbstractMatrix, W)`. Reproduced on unmodified `main` (a449aab), filed as https://github.com/SciML/LinearSolve.jl/issues/1236. The matrix-free test here names Krylov explicitly rather than working around it with `@test_broken`.

Not verified: sparse `J` in a `WOperator`, and GPU paths.

ChrisRackauckas-Claude:
You were right that this needed resolving before merge, though the mechanism turned out to be a different one — and worse than a crash.

## Matrix-free specifically: already dead upstream of LinearSolve

I set up a released-stack environment (OrdinaryDiffEq v7.6.0, LinearSolve v5.11.0, none of this PR) with a matrix-free `jac_prototype`:

```
jac_prototype only              TRBDF2 default linsolve   THREW MethodError @ convert
jac_prototype only              TRBDF2 + GMRES            THREW MethodError @ convert
jac_prototype + mass_matrix=I   TRBDF2 default linsolve   THREW MethodError @ convert
jac_prototype + mass_matrix=I   TRBDF2 + GMRES            THREW MethodError @ convert
```

Explicit GMRES fails identically, so this never reaches a linear solve. `ODEFunction` defaults `sparsity = jac_prototype`, and `prepare_user_sparsity` (OrdinaryDiffEqDifferentiation `alg_utils.jl:141`) calls `concrete_mass_matrix(sparsity)` -> `convert(AbstractMatrix, ...)` unconditionally. That is an OrdinaryDiffEq bug independent of this PR, and it means the matrix-free route cannot be regressed by merging — it is already broken.

## What the question did expose: silent wrong answers on the `islin` path

`build_J_W`s `islin` branch builds a `WOperator` whose `J` is a **`MatrixOperator`**, and `defaultalg` was unwrapping exactly that, so it selected LHL. But `calc_W!` guards its `mark_jacobian_updated!` behind `!(W.J isa AbstractSciMLOperator)` — so for an operator `J` the flag is never set, while `update_coefficients!` moves the numbers in place. Same object identity, same `jac_stale`. The reduction cannot tell.

Before, second solve after J moves:
```
step 1 relres = 2.46e-16
step 2 relres = 1.34            <- wrong answer, no error
   (vs OLD J) = 1.07
```
After:
```
step 1 relres = 1.56e-8
step 2 relres = 9.11e-9         <- tracks the new J
   (vs OLD J) = 0.909
```

`_lhl_defaultable` now requires a plain dense `J` — which is what OrdinaryDiffEq `_uses_split_W` already restricts itself to. The LinearSolve default had simply been more permissive than the OrdinaryDiffEq gate; they now agree. An operator `J` keeps Krylov.

Regression test added; reverting only the predicate fails all three assertions:
```
Expression: !(LinearSolve._lhl_defaultable(W, assump))
Expression: LinearSolve.defaultalg(W, b, assump) != DefaultLinearSolver(DefaultAlgorithmChoice.LHLFactorization)
Expression: ≈((solve!(cache)).u, (Anew - I / γ) \ b, rtol = 1.0e-6)
```

Full `test/Core/lhl.jl` exits 0. Runic and typos clean.

Not verified: GPU paths, and the OrdinaryDiffEq-side `prepare_user_sparsity` bug is filed only in this comment so far.

Issue #211, opened 2022-10-24
BibTex (citation) for LinearSolve.jl


ytdHuang:
Hi,
I am currently writing an article and also going to publish a new package.
This package (and also `DifferentialEquations.jl`) really helps me a lot on my research topic, and I would like to cite these packages.

I did see `DifferentialEquations.jl` provides the bibtex in github, but not for `LinearSolve.jl`.

Does the package owner agree with the following citations?
```bib
@software{LinearSolve.jl,
  author = {SciML},
  title = {LinearSolve.jl},
  url = {https://github.com/SciML/LinearSolve.jl}
}
```

I just want to make sure and hope to cite in the way the package owner and the developers want.

ChrisRackauckas:
@Wimmerer and @vpuri3 deserve to be in there. Let's say it's:

```bib
@software{LinearSolve.jl,
  author = {Will Kimmerer, Vedant Puri, Chris Rackauckas},
  title = {LinearSolve.jl},
  url = {https://github.com/SciML/LinearSolve.jl}
}
```

We should do a JOSS paper soon.

Wimmerer:
What's the policy on citing individual solvers. I guess once we have a paper out we can cite those there and that's sufficient.

ChrisRackauckas:
We should put the references for each solver in the solver's docstring, and have a solver API page with the docstrings.

Wimmerer:
I'll do at least the sparse docs later this week. The dense ones should be mostly easy to find.

ytdHuang:
I just finished the development of my package, and started to prepare the manuscript for publication.
Temporally, I will use the bibTex @ChrisRackauckas given above.
Let me know if there is any update for the citation of `LinearSolve.jl`.

briochemc:
> We should do a JOSS paper soon.

Is there a JOSS paper? Or any update on how to cite LinearSolve.jl?

ChrisRackauckas:
It hopefully should be done by the end of the month.

briochemc:
Here is a gentle reminder in case this has fallen through the cracks.

rayegun:
Hasn't fallen through, just fallen behind. I will do my best to get a
preprint up early next week.

ytdHuang:
Another gentle reminder (cause I'm writing another paper again).

ChrisRackauckas:
It's not done yet. It's being turned into a much larger review of linear solvers with benchmarks, making it into a more comprehensive paper.

[...]

PR #1216, opened 2026-08-15
Add a citation file

AJ0070:
## Checklist

## Additional context

Addresses the citation half of #211.

- Adds `CITATION.bib` with the entry from that thread, plus a short Citation section in the README so it is visible from the repo front page.
- One change from the entry as written in the issue: the three authors are separated with `and` rather than commas. BibTeX reads commas inside a single braced `author` field as the Last, First, Jr parts of one name, so the comma form collapses to a single mangled author.
- No tests, this adds no code.
- This does not close the issue. The rest of that thread is the JOSS or review paper and the request to enable Zenodo archiving for a DOI, both of which need maintainer action rather than a PR. The entry here is what people in the thread have been citing by hand since 2022, so having it in the repo seems worth doing while the paper is in progress.

AI Disclosure: Used Opus 5

### DIFF
--- a/CITATION.bib
+++ b/CITATION.bib
@@ -1,0 +1,5 @@
+@software{LinearSolve.jl,
+  author = {Will Kimmerer and Vedant Puri and Chris Rackauckas},
+  title = {LinearSolve.jl},
+  url = {https://github.com/SciML/LinearSolve.jl}
+}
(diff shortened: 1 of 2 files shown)


CITATION.bib
```diff
@@ -0,0 +1,5 @@
+@software{LinearSolve.jl,
+  author = {Will Kimmerer and Vedant Puri and Chris Rackauckas},
```
ChrisRackauckas: add yourself too as 3rd

AJ0070: Added, thanks.

Issue #136, opened 2022-05-24
low rank update on `cacheval`


vpuri3:
If you're reusing a LinearCache to solve a slightly perturbed system, it would be useful to do a low-rank update on `cacheval` instead of recomputing the factorization? something like a sherman-morrison formula?

ref
https://infoscience.epfl.ch/record/161468?ln=en
https://en.wikipedia.org/wiki/Woodbury_matrix_identity

ChrisRackauckas:
It could be. We could add it to the interface.

ChrisRackauckas:
Added low rank methods

PR #1220, opened 2026-08-15
Carry a low-rank update on the matrix

AJ0070:
## Checklist

## Additional context

Addresses #136. Rewritten after review: the update is carried by a matrix type rather than stuffed into the algorithm.

- `LowRankUpdatedMatrix(A, U, V; C = I)` is an `AbstractMatrix` holding `A + U C V'` unassembled. Pass it as the matrix of a `LinearProblem`; nothing new goes to `solve`.
- The solve factorizes `A` once and applies the Woodbury identity, so a low-rank change costs a `k x k` factorization rather than a fresh one of the whole matrix.
- This matters when `A` factorizes cheaply and the update does not preserve that structure. On a sparse tridiagonal base with a dense rank-1 update, against assembling and solving the sum: 7.2x at `n = 400` and 36.7x at `n = 2000`, agreeing to 1e-15.

Writing the documentation turned up that the feature was not running at all, which the last three commits fix:

- `LowRankUpdatedMatrix` defined no `Base.copy`, so the `copy(A)` in `__init` fell through to `Base.copymutable` and materialized the wrapper elementwise through `getindex` into a dense `Matrix`. Every solve then took the generic path over exactly the assembled sum the type exists to avoid, and `cacheval` was never a `LowRankUpdatedCache`.
- The tests missed it because they compared only numbers, and the assembled answer is identical. They now assert that the wrapper survives `init`, that `cacheval` is a `LowRankUpdatedCache`, and that a sparse base keeps its `UmfpackLU`.
- Three of the eight algorithms in `_LOWRANK_ALGS` never worked. The correction reaches the base factorization through `do_factorization`, and `GenericLUFactorization`, `UMFPACKFactorization` and `KLUFactorization` route through their own `solve!`, so all three raised a `MethodError` once the path became reachable. They are removed, and a test pins the list against the algorithms proven to solve. A sparse `A` is still covered by `LUFactorization`, which dispatches `lu` to UMFPACK.
- A singular update returned `ReturnCode.Failure` silently, so both failure branches now emit `@SciMLMessage` like the other factorizations.

Documentation, as requested:

- A tutorial at `tutorials/low_rank_updates.md` and a solvers page at `solvers/low_rank_solvers.md`, both registered in `pages.jl`.
- The `@docs` entry moved off `solvers.md` onto the new solvers page so the symbol is documented exactly once, and `solvers.md` keeps a short pointer in the style of the shifted-system section above it.

Two notes on the shape:

- Not `WoodburyMatrices.Woodbury`, which was the obvious type to reuse. It subtypes `Factorization` rather than `AbstractMatrix`, so `LinearProblem` sends it to `SciMLBase.isinplace` and rejects it as a model function.
- The supported factorizations are enumerated rather than dispatched on `AbstractFactorization`. Pairing the matrix type with a general algorithm was ambiguous against every `solve!` that pairs a specific algorithm with a general matrix, 81 of them by Aqua's count. Enumerating keeps each method strictly more specific.

One consequence worth flagging. Now that the wrapper survives `init`, a factorization outside the enumerated list raises rather than silently assembling. Krylov methods are unaffected and consume it matrix free through `mul!`. A friendly `ArgumentError` would need a method on `AbstractFactorization` plus the matrix type, which reintroduces the ambiguity the enumeration avoids, so I left it. Happy to add it another way.

38 tests cover rank-1 and rank-k updates, a middle `C`, the matrix interface, shape validation, reuse across right-hand sides, every enumerated factorization, Krylov through the wrapper, a sparse base keeping its sparse factorization, and a singular update reporting failure.


ChrisRackauckas:
This should use a more complex problem type, rather than stuffing problem data into the algorithm.

AJ0070:
@ChrisRackauckas makes sense. Do you want the matrix to carry it, a `WoodburyMatrices.Woodbury` as the `A` with an extension dispatching on it, or a new problem type in LinearSolve itself?

The first puts `A + U C V'` where the problem data belongs without inventing a problem type, at the cost of a dependency. Happy either way, just want to rewrite it once.

ChrisRackauckas:
Yes, specializing on a matrix type would probably be better.

AJ0070:
@ChrisRackauckas

AJ0070:
@ChrisRackauckas

ChrisRackauckas:
This needs a separate tutorial and solvers page since it's rather specific.

AJ0070:
Both pages are up: `tutorials/low_rank_updates.md` and `solvers/low_rank_solvers.md`, registered in `pages.jl`. The `@docs` entry moved onto the new solvers page so the symbol is documented once, and `solvers.md` keeps a short pointer.

Writing them turned up that the feature was not actually running. `LowRankUpdatedMatrix` had no `Base.copy`, so the `copy(A)` in `__init` densified the wrapper and every solve took the generic path over the assembled sum. That is fixed, three algorithms that never worked are out of `_LOWRANK_ALGS`, and the PR description has the details.

PR #1229, opened 2026-08-17
Hold off the collector across a ParU solve

### LINKED ISSUE #1142: Repeated ParUFactorization solves drive ParU_Analyze/METIS into a GC storm: sub-second solve becomes hours
## Summary

After roughly 100 or more cumulative `ParUFactorization` init+solve cycles in one Julia process, the next symbolic analysis on a large enough matrix effectively hangs: a solve that takes 0.55 s in a fresh process runs for hours at 100% CPU. The time is spent inside Julia's GC, triggered from ParU's METIS ordering.

Found by the SciMLBenchmarks SparseDirect sweep, which wedged reproducibly at the same sweep position across independent runs, and by two multi-hour benchmark-runner CI runs on SciMLBenchmarks PR #1627 (roughly 9 h and 23 h) that were burning the runner in this state before diagnosis.

## Reproduction (deterministic)

```julia
using LinearAlgebra, SparseArrays, LinearSolve, Random
import ParU_jll, SciMLBase

function tridiag(n)
    d = fill(2.0, n); d[1] = 1.0; d[end] = 1.0
    spdiagm(1 => -ones(n-1), 0 => d, -1 => -ones(n-1)) + 0.01I
end

for N in [10 * 2^k for k in 0:12]
    A = tridiag(N); b = rand(MersenneTwister(1), N)
    t = @elapsed for rep in 1:12
        c = SciMLBase.init(LinearProblem(A, b), ParUFactorization())
        solve!(c)
    end
    println("N=$N  12 inits in ", round(t; digits=2), "s"); flush(stdout)
end
```

Every size through N=10240 completes in 0.52 s or less. **N=20480 never completes** (killed after 15 minutes; earlier benchmark runs left it wedged for 7+ hours). A single solve of the same N=20480 system in a fresh process completes in 0.55 s, so the failure needs cumulative in-process state.

Reproduced identically on two unrelated machines:

- Dev laptop: LinearSolve v5.3.0, Julia 1.12.3, linux x86-64, 16 cores / 15 GB. Wedged at N=20480 (7+ h before killed).
- Clean cloud container (Debian slim, fresh depot, stock registry packages): Julia 1.12.6, 4 CPU / 16 GB. Completed every size through N=10240, then wedged at N=20480 (killed by a 20 minute cap).

## Evidence

`kill -USR1` on the wedged process (100% CPU, RSS ~530 MB) shows the main thread in a GC mark loop entered from METIS's allocator inside ParU's analysis:

```
maybe_collect
ijl_gc_counted_malloc
SuiteSparse_metis_gk_malloc          (libcholmod)
ParU_Analyze                          (libparu)
ParU_C_Analyze
solve! at ext/LinearSolveParUExt.jl:215
```

METIS allocates intensively during nested dissection; each allocation goes through Julia's counted-malloc path and its `maybe_collect` check. Under the accumulated state those checks fire actual collections over and over: the "hang" is thousands of full GC marks.

## What it is NOT

- Not a missing finalizer: `LinearSolveParUExt` registers `_paru_free_cache!` (around ext line 58) freeing symbolic/numeric/control objects.
- Not simple GC pressure: inserting `GC.gc(true)` between sizes in the reproducer does **not** prevent the wedge.

## Scale dependence

The ratchet is driven by large analyses, not call count alone: 3,000 factorizations of ~60x60 sparse matrices run at flat per-solve cost (0.11 s per 100 solves, no growth). This is consistent with METIS nested dissection only engaging above CHOLMOD's size cutoff, so small problems never enter the allocation-heavy path.

## Version scope

`ParU_jll` (1.0.0 and 1.0.1) declares `julia = "1.12.0 - 1"` in General, so there is no older-Julia configuration to bisect against. That minimum is itself a lead: a solver JLL pinning a Julia floor typically links libjulia, and the backtrace shows exactly that integration in action (METIS's `gk_malloc` routed into `ijl_gc_counted_malloc`). The counted-allocation wiring appears to be a deliberate feature of the ParU_jll build; the ratchet then looks like an asymmetry bug in that wiring rather than a general Julia GC issue.

## Hypothesis

Since a full collection does not reset whatever drives `maybe_collect`, the accounting itself looks suspect: if METIS/CHOLMOD allocations enter through `jl_gc_counted_malloc` but some frees exit through plain `free` (uncounted), the counted-allocation total ratchets upward with every ParU analysis and never comes back down. Eventually every counted malloc trips the collection heuristic. That would explain (a) the cumulative trigger, (b) `GC.gc` immunity, and (c) the wedge point scaling with prior ParU usage rather than problem size alone. Not yet verified against the SuiteSparse_config malloc/free wiring; input from someone who knows the JLL build would help.

## The damage is process-wide, not ParU-local

After capping ParU at N <= 10240 in the benchmark sweep (so it never again ran a large analysis), the sweep still wedged later, this time inside `SupernodalLUFactorization`'s symbolic analysis (`col_structure`, pure-Julia allocation-heavy code) at a size it completes in milliseconds standalone. ParU's accumulated counted allocations had ratcheted the process-wide GC state and every subsequent allocation-heavy code path became collateral. Removing ParU entirely is the only workaround found that lets the rest of the process run normally.

## Impact

Any workflow doing many ParU factorizations in one process (parameter sweeps, benchmark suites, long-running services) eventually hits this, and the slowdown lands on whatever allocates next, not necessarily on ParU itself, which makes it very hard to attribute in the wild. Two SciMLBenchmarks CI runs wedged this way before diagnosis. The SciMLBenchmarks SparseDirect and SparsePDE documents now exclude ParU with a documented omission pointing at this issue.

AJ0070:
## Checklist

## Additional context

Fixes #1142.

- The diagnosis in the issue holds: the time goes into Julia GC entered from METIS's allocator inside `ParU_Analyze`. Those collections cannot free what METIS holds, so they are pure overhead, and past a certain amount of accumulated analysis state they fire repeatedly and the solve stops progressing.
- `solve!` now runs with the collector disabled and restores the previous state in a `finally`.

Measured on Linux, Julia 1.12.7, 4 CPU, with the reproducer from the issue verbatim:

  | N | released | this branch |
  | --- | --- | --- |
  | 10240 | 1.099 s | 1.304 s |
  | 20480 | no finish in 300 s | 9.889 s |
  | 40960 | not reached | 2.913 s |

- Wrapping only the `ParU_C_Analyze` ccall is not enough. That was the first attempt and the sweep still wedged at N = 20480, so the guard covers the whole solve.
- The tradeoff is that Julia-side garbage is not collected for the duration of a ParU solve. That is bounded by one solve, against a failure mode that has burned multi-hour CI runs.
- No test. The sizes that trigger this take minutes even when healthy, and reaching the wedge needs a long in-process sweep, which does not belong in the suite. The reproducer in the issue is the check.

AI Disclosure: Used Opus 5

### DIFF
--- a/ext/LinearSolveParUExt.jl
+++ b/ext/LinearSolveParUExt.jl
@@ -166,17 +166,36 @@ function SciMLBase.solve!(
         cache::LinearSolve.LinearCache, alg::LinearSolve.ParUFactorization;
         kwargs...
     )
-    # METIS allocates heavily inside ParU's symbolic analysis, and each allocation goes
-    # through Julia's counted-malloc and its `maybe_collect` check. Those collections
-    # cannot free anything METIS holds, so once enough analysis state has accumulated in
-    # the process they fire repeatedly and the solve stops making progress: a 0.5 s solve
-    # has been seen to run for hours. Holding the collector off for the duration of the
-    # solve keeps it at 0.5 s. See SciML/LinearSolve.jl#1142.
-    gc_was_enabled = GC.enable(false)
-    try
-        return _paru_solve!(cache, alg; kwargs...)
+    # METIS allocates heavily inside ParU's symbolic analysis. SuiteSparse routes those
+    # allocations through Julia's counted allocator, which charges them to the GC
+    # heuristic even though the collector can never free them: they belong to METIS. Once
+    # enough of that memory is outstanding, every further allocation trips `maybe_collect`
+    # and the analysis stops making progress, turning a 0.5 s solve into hours.
+    #
+    # Handing SuiteSparse the plain libc allocator for the duration keeps that memory off
+    # Julia's books, which is what the accounting should have said in the first place. The
+    # collector stays on throughout. See SciML/LinearSolve.jl#1142.
+    return _with_uncounted_suitesparse_allocator() do
+        _paru_solve!(cache, alg; kwargs...)
+    end
+end
+
+function _with_uncounted_suitesparse_allocator(f)
+    prev_malloc = LinearSolve.KLU.SuiteSparse_config_malloc_func_get()
+    prev_calloc = LinearSolve.KLU.SuiteSparse_config_calloc_func_get()
+    prev_realloc = LinearSolve.KLU.SuiteSparse_config_realloc_func_get()
+    prev_free = LinearSolve.KLU.SuiteSparse_config_free_func_get()
+    LinearSolve.KLU.SuiteSparse_config_malloc_func_set(cglobal(:malloc))
+    LinearSolve.KLU.SuiteSparse_config_calloc_func_set(cglobal(:calloc))
+    LinearSolve.KLU.SuiteSparse_config_realloc_func_set(cglobal(:realloc))
+    LinearSolve.KLU.SuiteSparse_config_free_func_set(cglobal(:free))
+    return try
+        f()
     finally
-        GC.enable(gc_was_enabled)
+        LinearSolve.KLU.SuiteSparse_config_malloc_func_set(prev_malloc)
+        LinearSolve.KLU.SuiteSparse_config_calloc_func_set(prev_calloc)
+        LinearSolve.KLU.SuiteSparse_config_realloc_func_set(prev_realloc)
+        LinearSolve.KLU.SuiteSparse_config_free_func_set(prev_free)
     end
 end
 
--- a/ext/LinearSolveParUExt.jl
+++ b/ext/LinearSolveParUExt.jl
@@ -166,6 +166,24 @@ function SciMLBase.solve!(
         cache::LinearSolve.LinearCache, alg::LinearSolve.ParUFactorization;
         kwargs...
     )
+    # METIS allocates heavily inside ParU's symbolic analysis, and each allocation goes
+    # through Julia's counted-malloc and its `maybe_collect` check. Those collections
+    # cannot free anything METIS holds, so once enough analysis state has accumulated in
+    # the process they fire repeatedly and the solve stops making progress: a 0.5 s solve
+    # has been seen to run for hours. Holding the collector off for the duration of the
+    # solve keeps it at 0.5 s. See SciML/LinearSolve.jl#1142.
+    gc_was_enabled = GC.enable(false)
+    try
+        return _paru_solve!(cache, alg; kwargs...)
+    finally
+        GC.enable(gc_was_enabled)
+    end
+end
+
+function _paru_solve!(
+        cache::LinearSolve.LinearCache, alg::LinearSolve.ParUFactorization;
+        kwargs...
+    )
     A = cache.A
     A = convert(AbstractMatrix, A)
 


ext/LinearSolveParUExt.jl
```diff
@@ -166,6 +166,24 @@ function SciMLBase.solve!(
...
     )
+    # METIS allocates heavily inside ParU's symbolic analysis, and each allocation goes
+    # through Julia's counted-malloc and its `maybe_collect` check. Those collections
+    # cannot free anything METIS holds, so once enough analysis state has accumulated in
+    # the process they fire repeatedly and the solve stops making progress: a 0.5 s solve
+    # has been seen to run for hours. Holding the collector off for the duration of the
+    # solve keeps it at 0.5 s. See SciML/LinearSolve.jl#1142.
+    gc_was_enabled = GC.enable(false)
+    try
+        return _paru_solve!(cache, alg; kwargs...)
```
ChrisRackauckas: this seems like a hack and not a real fix

AJ0070: Fair, that was the symptom. SuiteSparse routes METIS's allocations through Julia's counted allocator, so memory the collector can never free is charged to the GC heuristic and past a threshold every allocation trips `maybe_collect`. The solve now swaps in the plain libc allocator and restores the pointers in a `finally`, collector left on.

N = 20480 goes from no finish in 300 s to 1.07 s, against 9.9 s for disabling the collector.

ext/LinearSolveParUExt.jl
```diff
@@ -166,6 +166,43 @@ function SciMLBase.solve!(
...
+    # allocations through Julia's counted allocator, which charges them to the GC
+    # heuristic even though the collector can never free them: they belong to METIS. Once
+    # enough of that memory is outstanding, every further allocation trips `maybe_collect`
+    # and the analysis stops making progress, turning a 0.5 s solve into hours.
+    #
+    # Handing SuiteSparse the plain libc allocator for the duration keeps that memory off
+    # Julia's books, which is what the accounting should have said in the first place. The
+    # collector stays on throughout. See SciML/LinearSolve.jl#1142.
+    return _with_uncounted_suitesparse_allocator() do
+        _paru_solve!(cache, alg; kwargs...)
```
ChrisRackauckas, line 179: This seems unsafe if the program fails. Does this have a finalizer that cleans up the memory?

AJ0070: Yes. `ParUCache` registers `finalizer(_paru_free_cache!, cache)` at `init_cacheval`, which calls `ParU_C_FreeNumeric`/`ParU_C_FreeSymbolic`, and that still runs whether or not the solve failed. The allocator pointers themselves are restored in a `finally`, so an error leaves them as it found them.

Freeing after the swap is fine because `jl_gc_counted_malloc` is plain `malloc` plus accounting with no header, so a block allocated under one allocator and freed under the other is a valid pointer either way. I stressed both directions, 20k allocations each, plus a full GC, with no fault.

PR #1234, opened 2026-08-18
Add an AMGX preconditioner

### LINKED ISSUE #512: AMGX.jl support
**Is your feature request related to a problem? Please describe.**

I think it would be nice if we can access [AMGX.jl](https://github.com/JuliaGPU/AMGX.jl) via LinearSolve.jl interface.

**Describe the solution you’d like**

Extension as usual.

AJ0070:
## Checklist

## Additional context

Fixes #512, as the callback route you asked for.

- `AMGXPreconditioner(A)` wraps AMGX's algebraic multigrid as something a Krylov method can apply, so it goes in through `precs`:

  ```julia
  using LinearSolve, CUDA, AMGX
  alg = KrylovJL_GMRES(precs = (A, p) -> (AMGXPreconditioner(A), I))
  sol = solve(LinearProblem(A_gpu_csr, b_gpu), alg)
  ```

- On a 400x400 tridiagonal system on a T4, GMRES goes from 14 iterations to 6 at the same accuracy (4.85e-9 against 5.19e-9). `A` must be a `CuSparseMatrixCSR`; anything else is refused with a message naming the conversion. The AMGX objects are released by a finalizer.

Two things worth knowing about the implementation:

- CUDA.jl's CSR is 1-based and AMGX expects 0-based. `AMGX.upload!` has a `CuSparseMatrixCSR` method but it does not rebase, and AMGX only notices when it builds the hierarchy, reporting `"Internal error"` from `setup!` with nothing pointing at indices. The indices are rebased before upload. This looks like an AMGX.jl bug and I am happy to report it there.
- `print_solve_stats` and `obtain_timings` are not registered as parameter-string keys in AMGX 2.4.0, so the default config omits them.

And two constraints on carrying this:

- The extension triggers on `AMGX` and `CUDA`, not `cuSOLVER`. AMGX 0.2.0 and cuSOLVER cannot be installed together, they disagree on `CUDACore`, the same conflict already noted for CUSOLVERRF and CUDSS in `test/qa/qa.jl`.
- There is no test in the suite, because AMGX.jl ships no library and needs a system AMGX install, which puts this with the Elemental, HYPRE, MUMPS, PETSc and STRUMPACK extensions that `qa.jl` already leaves unscanned. I verified it against AMGX 2.4.0 built from source on a T4; the numbers above are from that run.

AI Disclosure: Used Opus 5


src/extension_algs.jl
```diff
@@ -1516,6 +1516,51 @@ struct BLISLUFactorization <: AbstractFactorization
...
+The AMGX resources are freed by a finalizer, so the preconditioner can be left to
+the garbage collector.
+"""
+mutable struct AMGXPreconditioner
+    config::Any
+    resources::Any
+    matrix::Any
+    solver::Any
+    xvec::Any
+    bvec::Any
```
ChrisRackauckas: why any?

ChrisRackauckas:
Needs to be documented

src/extension_algs.jl
AJ0070: Fixed, it is parametric now. The fields hold AMGX objects, whose types only exist once the extension is loaded, so the struct takes parameters and the extension fills them in. They were untyped because the finalizer nulled them; it sets a `closed` flag instead.

AJ0070:
Documented on the preconditioners page: a GPU algebraic multigrid section with the `precs` usage, the iteration count, and the note that AMGX.jl needs a system AMGX install, plus a `@docs` entry for the type.

PR #1241, opened 2026-08-21
Gate the Block Diagonal GPU test against the direct solution

### LINKED ISSUE #1218: genericlu_naive_ldiv size-aware back-solve test fails on main across Julia 1, lts and pre
`test/Core/genericlu_naive_ldiv.jl:122` fails on unmodified main across Julia 1, lts and pre, reporting 503 passed / 1 failed, so every PR currently shows a red Core check.

The assertion is that at least one of 32 random draws differs bitwise from `ldiv!`:

```julia
@test any(1:32) do _
    bk = rand(n)
    cache.b = bk
    solve!(cache).u != ldiv!(similar(bk), F, copy(bk))
end
```

The comment above it puts bit agreement at 0.13% of draws, so all 32 agreeing is not plausible as bad luck and suggests the naive back-solve path stopped being taken. It passes locally on Windows, so it looks specific to the CI runners' BLAS.

AJ0070:
## Checklist

## Additional context

Fixes #1218.

- The Core failure the issue names (genericlu_naive_ldiv.jl:122) was already fixed by #1187 and #1223; the file passes 504/504 on unmodified main locally.
- The remaining red is the Tests GPU job: the Block Diagonal testset flipped green to red between 6166a400 and d74dcc0e, and neither diff can affect GMRES numerics. That points at test fragility, not a regression.
- The old form compares the blocked and unblocked SimpleGMRES answers to each other at the default Float32 isapprox rtol (about 3.4e-4) on unseeded rand(2, 2) blocks. Measured on a T4: 9/400 draws fail. The failing CI run evaluated a 1.6e-3 deviation, consistent with that.
- No solver bug is involved, checked rather than assumed: against a direct CPU reference both paths are correct in all 400 draws (blocked worst relative error 3.1e-7, unblocked 8.6e-4). Each path converges to within the solver's own Float32 tolerance, so two correct answers can legitimately sit about 1e-3 apart; a solver-vs-solver comparison at the default rtol is structurally flaky on GPU.
- The testset now draws seeded Float32 data with diagonally dominant blocks, gates both solutions against the direct solution, and cross-checks them at an explicit 5e-3, so it still fails on a genuinely wrong answer.
- Tolerances are measured, not guessed: the exact committed draw passes on a T4 with a 481x margin, and the same assertions over 400 different seeds show 0 failures, worst case 1.3e-3.
- Runic clean.

AI Disclosure: Used Fable 5

### DIFF
--- a/test/GPU/cuda.jl
+++ b/test/GPU/cuda.jl
@@ -124,8 +124,18 @@ y = solve(prob1)
 using BlockDiagonals
 
 @testset "Block Diagonal Specialization" begin
-    A = BlockDiagonal([rand(2, 2) for _ in 1:3]) |> cu
-    b = rand(size(A, 1)) |> cu
+    # Seeded, Float32 end-to-end, and gated against the direct solution. The
+    # earlier form compared the blocked and unblocked iterative answers to each
+    # other at the default isapprox rtol (~3.4f-4): on GPU each path converges
+    # to within the solver's own Float32 tolerance, so two correct answers can
+    # legitimately differ by ~1f-3 and the comparison flaked (measured 9/400
+    # draws on a T4; worst observed error vs direct was 8.6f-4).
+    rng_bd = StableRNG(42)
+    A_cpu = BlockDiagonal([rand(rng_bd, Float32, 2, 2) + 2.0f0 * I for _ in 1:3])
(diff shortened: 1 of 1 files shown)


AJ0070:
@ChrisRackauckas

PR #1240, opened 2026-08-20
Fix default algorithm QA regressions

ChrisRackauckas-Claude:
> Ignore this PR until it has been reviewed by @ChrisRackauckas.

## What changed and why

The QA group on `main` was red from a `WOperator`/GPU right-hand-side dispatch ambiguity, an unused `mark_jacobian_updated!` import, and two newly introduced extension-internal qualified accesses. This adds the missing intersection method, removes the stale import, and records the extension-only `_check_matrix_support` and `csrlsvqr!` accesses in the existing strict ExplicitImports inventories.

The regression test uses `JLArray`, a CPU-backed GPU array, to cover the ambiguous dispatch without requiring GPU hardware.

### Failing before

At [`6166a400bd393af4653ea752033a6ab792517408`](https://github.com/SciML/LinearSolve.jl/commit/6166a400bd393af4653ea752033a6ab792517408), the new regression case failed:

```julia
julia +release --project=.tmp-gpu-rhs-before -e 'using LinearSolve, JLArrays, LinearAlgebra; using SciMLOperators: WOperator; W = WOperator{true}(I, 0.1, Matrix{Float64}(I, 2, 2), zeros(2)); alg = LinearSolve.defaultalg(W, JLArray(ones(2)), LinearSolve.OperatorAssumptions(true)); @assert alg.alg === LinearSolve.DefaultAlgorithmChoice.KrylovJL_GMRES'
```

```text
ERROR: MethodError: defaultalg(::WOperator{...}, ::JLArray{Float64, 1}, ::OperatorAssumptions{Bool}) is ambiguous.

Candidates:
  defaultalg(A::WOperator, b, assump::OperatorAssumptions{Bool})
  defaultalg(A::SciMLOperators.AbstractSciMLOperator, b::GPUArraysCore.AnyGPUArray, assump::OperatorAssumptions{Bool})
```

The clean base QA run also reproduced all three defects:

```sh
GROUP=QA julia +release --project -e 'using Pkg; Pkg.test()'
```

```text
Quality Assurance | 46 pass, 1 fail, 2 errors, 49 total
```

The first bad commits were bisected to:

- [`47cdfb2eb81757346bb06cd8a3053a9736d15291`](https://github.com/SciML/LinearSolve.jl/commit/47cdfb2eb81757346bb06cd8a3053a9736d15291): ambiguity and stale import
[...]

PR #1242, opened 2026-08-21
Keep a GPU WOperator off the host-only LHL reduction

AJ0070:
## Checklist

## Additional context

Follows #1240, which resolved the `defaultalg` ambiguity between `WOperator` and a GPU right hand side by forwarding to the `WOperator` method. That method can answer `LHLFactorization`, whose reduction lives in LHLFactorization.jl and runs on the host.

- A GPU array still satisfies `DenseMatrix`, since `CuArray <: DenseArray`, so `_lhl_defaultable` accepts a device Jacobian at any `n` from `LHL_DEFAULT_MIN_SIZE` up. The default picked a host factorization for a device operator and the solve failed.
- This forwards to the operator default instead. A host `WOperator` is untouched and still reaches `LHLFactorization`.

Measured on a T4, on `main`:

```
GPU  WOperator default = LHLFactorization
J isa DenseMatrix on GPU: true
SOLVE: FAILED  Scalar indexing is disallowed.
```

and with this change:

```
GPU  WOperator default = KrylovJL_GMRES
host WOperator default = LHLFactorization
SOLVE: ok  retcode=Success  relerr=1.26e-9
```

The test in `test/GPU/cuda.jl` covers the algorithm choice, that the solve now succeeds and agrees with the dense answer, and that a host `WOperator` still takes the LHL path.

This PR originally carried the three QA fixes on `main`. #1240 landed the same three while it was open, so those are dropped and only the GPU routing remains.

### DIFF
--- a/src/default.jl
+++ b/src/default.jl
@@ -252,10 +252,18 @@ function defaultalg(A::WOperator, b, assump::OperatorAssumptions{Bool})
     return @invoke defaultalg(A::SciMLOperators.AbstractSciMLOperator, b, assump)
 end
 
+# Routed to the operator default rather than to the `WOperator` method above, because
+# that one can answer `LHLFactorization`. The reduction behind it lives in
+# LHLFactorization.jl and runs on the host, while a GPU array still satisfies
+# `DenseMatrix` (`CuArray <: DenseArray`), so `_lhl_defaultable` would accept a device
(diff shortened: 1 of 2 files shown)


AJ0070:
@ChrisRackauckas

PR #857, opened 2025-12-16
Fix GPU Cholesky cache initialization for non-square matrices

ChrisRackauckas-Claude:
## Summary

Fixes https://github.com/SciML/NonlinearSolve.jl/issues/746

When using `DefaultLinearSolver` with non-square GPU matrices (e.g., for least squares problems in NonlinearSolve.jl), the `init_cacheval` function for `CholeskyFactorization` would fail with `DimensionMismatch: matrix is not square` because it tried to compute `cholesky(A)` on a non-square matrix.

### Root Cause

The `@generated` `init_cacheval` for `DefaultLinearSolver` initializes caches for **all** possible algorithms upfront, including `CholeskyFactorization`. For non-square GPU matrices, this fails even though `CholeskyFactorization` would never actually be used (the `defaultalg` function correctly selects `QRFactorization` for non-square matrices).

### Fix

The fix checks `assumptions.issq` before attempting Cholesky factorization in `init_cacheval(::CholeskyFactorization, ::GPUArraysCore.AnyGPUArray, ...)` and returns `nothing` for non-square matrices, allowing the cache initialization to succeed.

## Changes

- **`src/factorization.jl`**: Added square matrix check in `init_cacheval` for `CholeskyFactorization` with GPU arrays
- **`test/gpu/cuda.jl`**: Added tests for non-square GPU matrices (overdetermined and underdetermined systems)

## Test plan

- [x] LinearSolve package loads correctly
- [ ] CI tests pass
- [ ] New CUDA tests for non-square matrices pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1250, opened 2026-08-24
Solve a non-square GPU system instead of throwing

AJ0070:
## Checklist

## Additional context

Revives #857, which is on another fork and conflicts. Solving a non-square GPU system threw `DimensionMismatch`, in two places:

- The default solver initializes every algorithm slot up front, and the `CholeskyFactorization` and `NormalCholeskyFactorization` slots both call `cholesky` on `A` itself. Both now skip a non-square `A`.
- A wide `A` then failed in the QR solve, since a GPU `qr` factors it but `ldiv!` builds `UpperTriangular` on a non-square `R`. Wide GPU input now goes through the existing `MinNormQR`, as the banded extensions already do, which also gives the minimum-norm solution.

Checked on a T4: tall, wide and square all solve, and the wide case matches LAPACK's `‖x‖`. #857's guard alone is no longer enough, since the `NormalCholeskyFactorization` slot moved to `sparsearrays.jl` and fails first.

AI Disclosure: Used Opus 5

### DIFF
--- a/src/factorization.jl
+++ b/src/factorization.jl
@@ -1054,7 +1054,13 @@ function do_factorization(alg::QRFactorization, A, b, u)
         # `alg.pivot` so the return type is determined by the static
         # `QRFactorization{P}` parameter (otherwise this branch returns
         # `Union{QRCompactWY, QRPivoted}` depending on `alg.inplace`).
-        if A isa GPUArraysCore.AnyGPUArray || is_cusparse(A) || issparsematrixcsc(A)
+        if A isa GPUArraysCore.AnyGPUArray && !is_cusparse(A) && is_underdetermined(A)
+            # A GPU `qr` factors a wide matrix happily, but solving with the result
+            # builds `UpperTriangular(R)` on an `R` that is not square and throws.
+            # Going through `Aᵀ` turns it back into a triangular solve and gives the
+            # minimum-norm solution, which is what dense `\` returns on the CPU.
+            fact = MinNormQR(qr(copy(transpose(A))))
+        elseif A isa GPUArraysCore.AnyGPUArray || is_cusparse(A) || issparsematrixcsc(A)
             fact = qr(A)
         elseif alg.inplace
             if A isa Symmetric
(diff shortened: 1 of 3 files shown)


src/factorization.jl
```diff
@@ -1192,6 +1204,12 @@ function init_cacheval(
         alg::CholeskyFactorization, A::GPUArraysCore.AnyGPUArray, b, u, Pl,
         Pr, maxiters::Int, abstol, reltol, verbose::Union{LinearVerbosity, Bool}, assumptions::OperatorAssumptions
     )
+    # `cholesky` needs a square matrix. The default solver initializes every slot it
+    # holds, this one included, before it knows which it will use, so a non-square `A`
+    # would fail here rather than reaching the least-squares algorithm that will actually
+    # serve it. Leave the slot empty instead.
+    # See https://github.com/SciML/NonlinearSolve.jl/issues/746
+    assumptions.issq || return nothing
     return cholesky(A; check = false)
```
ChrisRackauckas: this isn't type-stable? It should probably construct an empty cholesky

AJ0070: Good catch. `issq` is a field, not a type parameter, so that branch made the return a `Union`. Now instances an empty `cholesky(similar(A, 0, 0); check = false)`, which comes back as the same `Cholesky` type as the real one on a T4.

The `NormalCholeskyFactorization` slot in `sparsearrays.jl` still returns `nothing`, since it already does on its two `is_cusparse` branches. Happy to make that one consistent too if you prefer.

src/sparsearrays.jl
```diff
@@ -1133,6 +1133,13 @@ function LinearSolve.init_cacheval(
...
         nothing
+    elseif A isa LinearSolve.GPUArraysCore.AnyGPUArray && !assumptions.issq
+        # A non-square GPU `A` cannot be instanced here: `cholesky_instance` needs a
+        # square matrix, and the default solver reaches this slot for every `A` before
+        # it knows which algorithm it will use. Sparse CPU input is deliberately not
+        # covered by this branch, since `cholesky_instance` handles a non-square sparse
+        # `A` and `solve!` stores a real factorization into the slot afterwards.
+        nothing
     else
         ArrayInterface.cholesky_instance(convert(AbstractMatrix, A))
```
ChrisRackauckas: Why is this not just a cholesky instance? That would keep it type-stable and 0x0 is square.

AJ0070:
@ChrisRackauckas

Issue #1236, opened 2026-08-18
Default polyalgorithm cannot init cachevals for a matrix-free WOperator


ChrisRackauckas-Claude:
Reproduces on `main` at a449aab, independent of any LHL work.

A `WOperator` whose Jacobian is matrix-free solves fine with an explicit Krylov algorithm, but throws when the algorithm is left to the default. `_init_default_cacheval` builds a cacheval for **every** slot, and the `PureKLUFactorization` slot calls `convert(AbstractMatrix, W)`, which a matrix-free `J` cannot satisfy.

## Reproducer

```julia
using LinearSolve, LinearAlgebra, SciMLOperators, Random
using SciMLOperators: WOperator, FunctionOperator

n = 40
J = randn(MersenneTwister(1), n, n)
b = randn(MersenneTwister(2), n)
fj(v, u, p, t) = J * v
fj(w, v, u, p, t) = mul!(w, J, v)
W = WOperator{true}(I, 0.1, FunctionOperator(fj, zeros(n), zeros(n); islinear = true), zeros(n))

solve(LinearProblem(W, b), KrylovJL_GMRES())   # fine
solve(LinearProblem(W, b))                     # throws
```

## Observed

```
MethodError: Cannot `convert` an object of type var"#fj#fj##0"{Matrix{Float64}} to an object of type AbstractMatrix

 [1] convert(::Type, L::FunctionOperator{...})
   @ SciMLOperators .../src/func.jl:729
 [2] convert(::Type, W::WOperator{...})
   @ SciMLOperators .../src/woperator.jl:256
 [3] init_cacheval(alg::PureKLUFactorization, A::WOperator{...})
   @ LinearSolve .../src/sparsearrays.jl:747
 [4] macro expansion
   @ .../src/default.jl:663 [inlined]
 [5] _init_default_cacheval(alg::LinearSolve.DefaultLinearSolver, A::WOperator{...})
   @ LinearSolve .../src/default.jl:628
```

The explicit-Krylov solve on the same operator returns a correct answer (`relres = 1.56e-8`), so the operator itself is well-formed; only the eager cacheval construction fails.

## Suggested fix

The sparse-factorization slots should decline to build a cacheval for an `A` that is not convertible to a matrix, the same way the Krylov slots already skip `A isa DenseMatrix || issparsematrixcsc(A)`. A predicate along the lines of `SciMLOperators.isconvertible(A)` guarding those slots would cover it.

[...]

PR #1255, opened 2026-08-26
Test that a matrix-free WOperator reaches the default solver

AJ0070:
## Checklist

## Additional context

For #1236. Fixed at the source in SciML/SciMLOperators.jl#427, so all that is left here is the regression test.

`has_concretization(::AbstractWOperator)` was an unconditional `true`, so a `WOperator` over a matrix-free Jacobian claimed a concrete form it cannot produce, and the slots that guard on the trait threw a `MethodError` out of `init`. The workaround I had here is gone.

Blocked until SciMLOperators tags a release with that fix and the compat bound moves, so the test will be red until then.

AI Disclosure: Used Opus 5


src/LinearSolve.jl
```diff
@@ -302,6 +302,14 @@ needs_concrete_A(alg::AbstractKrylovSubspaceMethod) = false
 needs_concrete_A(alg::AbstractSolveFunction) = false
 
 # Util
+# `SciMLOperators.has_concretization(::AbstractWOperator)` is an unconditional `true`, so a
+# `WOperator` over a matrix-free Jacobian says it can be concretized and then throws from
+# `convert`. The default solver builds a cacheval for every slot it holds, several of which
+# concretize, so that lands as a `MethodError` from `init` before any algorithm runs. Ask
+# the Jacobian instead. See https://github.com/SciML/LinearSolve.jl/issues/1236.
+_has_concretization(A) = SciMLOperators.has_concretization(A)
```
ChrisRackauckas: Fix that at the source, don't pepper over it here.

PR #1270, opened 2026-08-29
Add a CKTSO sparse solver extension

### LINKED ISSUE #392: Add support for cktso
https://github.com/chenxm1986/cktso

In my preliminary comparison, it is much faster than KLU on solving multi-vector rhs.

AJ0070:
## Checklist

## Additional context

Closes #392. Adds `CKTSOFactorization` as an extension over [CKTSO.jl](https://github.com/AJ0070/CKTSO.jl), which I wrote and registered for this. CKTSO keeps its symbolic analysis and refactorizes the same pattern with new values, so the cache holds the factorization and only re-analyzes when the pattern actually changes.

The library is not redistributable and has no JLL, so the tests skip unless `CKTSO_LIBRARY` points at a copy. I checked solve, cached solve, refactorize, new rhs and pattern change against the real library on Linux, all around 5e-16, with singular matrices coming back as `ReturnCode.Failure` rather than throwing.

Happy to move CKTSO.jl to SciML or JuliaLinearAlgebra as you asked, I just need access to the org.

AI Disclosure: Used Opus 5


AJ0070:
@ChrisRackauckas

PR #1277, opened 2026-08-31
Feature/trilinos amesos2

Mxyank24:
Draft: Setup basic CxxWrap bindings for Amesos2

@ChrisRackauckas Here is the minimal CxxWrap skeleton successfully compiling and hooking into SciMLBase.solve!. If this architecture looks good and avoids the JLL machinery concerns, I will start mapping the CSR matrix data to Tpetra inside the C++ wrapper next.


ChrisRackauckas:
It wouldn't be in the main project here...

Mxyank24:
Should I spin up a standalone Amesos2.jl (or Trilinos.jl) package for the C++ bindings first and then just add a lightweight extension here once it's working?

ChrisRackauckas:
This probably needs a whole separate package for the bindings

Mxyank24:
Already on it pulled the bindings into their own package: github.com/Mxyank24/Amesos2.jl

CxxWrap binding builds and links against Trilinos_jll on macOS, and it's building a real Tpetra::CrsMatrix from CSR arrays now. Still need to wire up the actual Amesos2 solve call, working on that next.

PR #1287, opened 2026-09-09
Disable BLIS extension on i686 (no blis_jll product)

ChrisRackauckas-Claude:
## Summary
- `blis_jll` has no i686 product, so `LinearSolveBLISExt` fails to precompile on 32-bit with `UndefVarError: blis not defined`.
- Guard the extension with `@static if` so it loads disabled when `blis_jll.blis` is absent (WORD_SIZE != 64 or product missing).

## Test plan
- [ ] Core x86 CI green
- [ ] Core x64 still exercises BLIS when available

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.github/workflows/Downstream.yml
+++ b/.github/workflows/Downstream.yml
@@ -15,8 +15,7 @@ jobs:
           - {user: SciML, repo: OrdinaryDiffEq.jl, group: InterfaceII}
           - {user: SciML, repo: ModelingToolkit.jl, group: All}
           - {user: SciML, repo: SciMLSensitivity.jl, group: Core1}
-          # Temporarily disabled: MCI≤0.2 vs LatticeRules 0.0.2 (needs MCI 0.3 registry)
-          # - {user: SciML, repo: BoundaryValueDiffEq.jl, group: All}
+          - {user: SciML, repo: BoundaryValueDiffEq.jl, group: All}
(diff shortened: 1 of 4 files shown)


ChrisRackauckas-Claude:
Also fixed unpivoted QR singularity detection: use the same relative R-diagonal threshold as `_qr_rank_deficient` so x86 OpenBLAS near-zeros are caught (was the remaining Core x86 failure in `retcodes.jl`).

.github/workflows/Downstream.yml
```diff
@@ -15,7 +15,8 @@ jobs:
           - {user: SciML, repo: OrdinaryDiffEq.jl, group: InterfaceII}
           - {user: SciML, repo: ModelingToolkit.jl, group: All}
           - {user: SciML, repo: SciMLSensitivity.jl, group: Core1}
-          - {user: SciML, repo: BoundaryValueDiffEq.jl, group: All}
+          # Temporarily disabled: MCI≤0.2 vs LatticeRules 0.0.2 (needs MCI 0.3 registry)
+          # - {user: SciML, repo: BoundaryValueDiffEq.jl, group: All}
```
ChrisRackauckas:
```suggestion
          - {user: SciML, repo: BoundaryValueDiffEq.jl, group: All}
```

PR #1307, opened 2026-09-14
Back up A for the extension LU branches before the QR fallback

### LINKED ISSUE #1306: Extension LU algorithms never snapshot `A`, so the QR safety fallback restores a stale matrix
## Description

This is #887 again, for the LU algorithms that live in extensions.

#887 was fixed by having the in-place LU algorithms snapshot `cache.A` into the cacheval's
`A_backup` before factorizing, through `_copy_A_for_safety`, so that `_do_qr_fallback` has
the original matrix to restore. `LUFactorization`, `GenericLUFactorization`,
`MKLLUFactorization`, `AppleAccelerateLUFactorization`, `OpenBLASLUFactorization` and the
generic `AbstractFactorization` path all call it. The four supplied by extensions never
do:

```
$ grep -c _copy_A_for_safety ext/LinearSolve{RecursiveFactorization,BLIS,CUDA,Metal}Ext.jl
0
0
0
0
```

`a_backup_allocated` therefore never flips for them, so `A_backup` goes on aliasing the `A`
the cache was built from for the life of the cache. The copy in `setproperty!` is gated on
that same flag, so it never fires either and the two guards mask each other.

The consequence splits in two:

- `RFLUFactorization` and `BLISLUFactorization` factorize `cache.A` in place
  (`RecursiveFactorization.lu!(A, ipiv, ...)`, `getrf!(A_work, ...)`), exactly as
  `MKLLUFactorization` did in #887. The fallback restores a matrix from an earlier
  refactorization and returns its solution with `ReturnCode.Success`.
- `CudaOffloadLUFactorization` and `MetalLUFactorization` copy to the device and leave
  `cache.A` intact, so the fallback instead `copyto!`s the stale backup *over* a matrix
  that needed no restoring. Narrower, since neither passes a retcode to
  `build_linear_solution` and so never reports `Failure`; only the non-finite check can
  reach the fallback.

The first solve is fine either way, because `A_backup` still aliases the matrix the cache
was built from. It goes wrong from the first refactorization onward, which is the
`cache.A = J; solve!(cache)` loop of a Newton or implicit ODE solve.

## Where

- `src/factorization.jl:43`, `_copy_A_for_safety`, and its call sites: `factorization.jl`
  L602 / L744 / L2684, `mkl.jl:329`, `appleaccelerate.jl:343`, `openblas.jl:372`. No
  extension is among them.
- `src/default.jl:787`, the only consumer: `copyto!(cache.A, cache.cacheval.A_backup)`.
- `src/common.jl:336-346`, the `setproperty!` copy, gated on `a_backup_allocated`.

## Reproduction

LinearSolve 5.17.4 (`main` at eb0bd76), Julia 1.13.0, with RecursiveFactorization loaded.

```julia
using LinearSolve, LinearAlgebra, RecursiveFactorization, StableRNGs
rng = StableRNG(11); n = 20

A0 = rand(rng, n, n) + n * I
b = rand(rng, n)
cache = init(LinearProblem(copy(A0), copy(b)),
             LinearSolve.DefaultLinearSolver(LinearSolve.DefaultAlgorithmChoice.RFLUFactorization))
solve!(cache)

A1 = rand(rng, n, n) + n * I
A1[:, 2] .= 0.0                       # singular, so LU fails and the fallback fires
cache.A = copy(A1)
u = copy(solve!(cache).u)

@show u ≈ qr(A0, ColumnNorm()) \ b    # the previous matrix
@show u ≈ qr(A1, ColumnNorm()) \ b    # the one actually asked for
```

```
u ≈ qr(A0, ColumnNorm()) \ b = true
u ≈ qr(A1, ColumnNorm()) \ b = false
```

with `retcode = ReturnCode.Success`. Stepping through the flags:

```
after init   : allocated=false  A_backup === prob.A : true   cache.A === prob.A : false
after solve1 : allocated=false  A_backup === prob.A : true   A_backup == A0 : true
after A=A1   : allocated=false  A_backup still == A0 : true  (should be A1)
after solve2 : retcode=Success  fell_back_to_qr=true
```

Running the same refactor-to-singular sequence across the default's dense LU choices, with
both a fresh `A` object and the in-place `copyto!(Aw, A1); cache.A = Aw` pattern, gives
identical results for each style:

| default choice | `\|A1 u - b\|` | best possible | verdict |
| --- | --- | --- | --- |
| LUFactorization | 0.62251 | 0.62251 | ok |
| GenericLUFactorization | 0.62251 | 0.62251 | ok |
| AppleAccelerateLUFactorization | 0.62251 | 0.62251 | ok |
| QRFactorization | 0.62251 | 0.62251 | ok |
| RFLUFactorization | 0.66296 | 0.62251 | solves the previous matrix |

BLIS, CUDA and Metal are read from the source rather than run; none was available here.

## Suggested fix

Take the snapshot in the default solver's `solve!` for the four branches whose algorithms
do not take it themselves, rather than asking each extension to opt in:

```julia
function _snapshot_A_for_fallback!(cache::LinearCache, alg::DefaultLinearSolver)
    (alg.safetyfallback && cache.isfresh) && _copy_A_for_safety(cache)
    return nothing
end
```

called at the top of the `RFLUFactorization`, `BLISLUFactorization`,
`CudaOffloadLUFactorization` and `MetalLUFactorization` branches.

That also makes the `setproperty!` copy redundant for every path that can reach
`_do_qr_fallback`, so it can go. It is a second O(n^2) pass per refactorization on top of
the one `_copy_A_for_safety` already does, measured at 3.1% to 4.5% of a refactorize and
solve for n between 20 and 400. `a_backup_synced` is written in three places and read in
none, and can go with it.

AJ0070:
## Checklist

## Additional context

Fixes #1306, which is #887 again for the LU algorithms that live in extensions. #887 was fixed by having the in-place LU algorithms snapshot `cache.A` into `A_backup` through `_copy_A_for_safety` so the QR fallback has the original matrix to restore; the four supplied by extensions never call it, so `a_backup_allocated` never flips and `A_backup` goes on aliasing the `A` the cache was built from. The fallback then restores a matrix from an earlier refactorization and returns its solution with `ReturnCode.Success`. The first solve is fine, because the alias is still the right matrix; it goes wrong from the first refactorization onward, which is the `cache.A = J; solve!(cache)` loop of a Newton or implicit ODE solve.

The snapshot is taken in the default solver's `solve!` for those four branches rather than in each extension, so a backend that adds an in-place LU later is covered without having to know about the contract. That also makes the copy in `setproperty!` redundant for every path that can reach `_do_qr_fallback`, so it goes; it was a second O(n^2) pass per refactorization on top of the one `_copy_A_for_safety` already does. `a_backup_synced` was written in three places and read in none, and goes with it.

Refactorize and solve on the dense default, min of 3000 samples at one BLAS thread, is 3.4% to 5.4% faster for n between 20 and 400 as a result.

RFLU is the only one of the four this machine can run; BLIS, CUDA and Metal are covered by reading their `solve!` implementations. They split into two shapes: RFLU and BLIS factor `cache.A` in place and so restore a stale matrix, while CUDA and Metal copy to the device and leave `cache.A` intact, so the stale restore would instead overwrite a matrix that needed no restoring. Both are fixed by the same snapshot, and the three unexercised branches sit behind the `useblis`/`usecuda`/`usemetal` guards that already gate them.

Checked with the source reverted, where the new testset fails on both reassignment styles. Core, AD, DefaultsLoading and Preferences all pass, and the allocation QA still holds, which matters because the removed copy is on the refactorization hot path.

AI Disclosure: Used Opus 5


src/default.jl
```diff
@@ -731,6 +729,20 @@ function _is_gpu_sparse(A)
...
+    _snapshot_A_for_fallback!(cache::LinearCache, alg::DefaultLinearSolver)
+
+Save `cache.A` into the cacheval's `A_backup` before an in-place factorization, so
+[`_do_qr_fallback`](@ref) has the original matrix to restore.
+
+The in-tree LU algorithms do this themselves through `_copy_A_for_safety`; the ones
+supplied by extensions do not, so the default solver takes the snapshot for them.
+"""
+function _snapshot_A_for_fallback!(cache::LinearCache, alg::DefaultLinearSolver)
+    (alg.safetyfallback && cache.isfresh) && _copy_A_for_safety(cache)
```
ChrisRackauckas, line 742: this allocates each time.

AJ0070:
It reuses the buffer. `_copy_A_for_safety` allocates once per cache, or on a size change, and `copyto!`s after that. Six fresh solves at n=300 allocate `[0, 0, 0, 0, 0, 0]` bytes.

The copy is about 2% of the LU (2.7% at n=300, 2.0% at n=2000) and it is what makes the fallback correct. Refactorizing with a singular matrix returns the previous system's solution on main, with `ReturnCode.Success`:

```
main:     [0.333333, 0.666667, 1.0, 1.333333, 1.666667, 2.0]   (= A1 \ b, the old matrix)
patched:  [2.0, 2.0, 0.0, 4.0, 5.0, 6.0]                       (= pivoted QR of A2)
```

Same gating the in-tree LU branches already use, and `safetyfallback` is on by default, so this is the cost those branches pay today.

I can drop it from the Metal and CUDA branches if you prefer. Neither touches the host matrix, and both throw on a singular factorization instead of returning `Failure`, so the fallback never runs there.

PR #1311, opened 2026-09-14
Add the non-square pullback term the reverse rules were missing

### LINKED ISSUE #1309: Reverse-mode `∂A` drops the residual term for non-square solves, disagreeing with forward mode
## Description

All three reverse-mode rules compute `∂A` with the square-system formula `∂A = -λxᴴ`. For a
non-square `A` the solve is `x = A⁺b`, whose pullback carries a second term coming from the
residual:

```
∂A = -λxᴴ + r ((AᴴA)⁻¹ ∂x)ᴴ ,    r = b - Ax ,   λ = Aᴴ \ ∂x
```

`λ` is still right, so `∂b` is correct; only `∂A` is wrong, and it is wrong silently. The
ForwardDiff extension derives and includes the residual term for the overdetermined case, so
forward and reverse mode disagree with each other on the same problem.

Reachable through `QRFactorization`, `SVDFactorization`, `NormalCholeskyFactorization`,
`KrylovJL_LSMR`, and the default solver on any non-square `A`.

## Where

The same formula in three places:

- `ext/LinearSolveChainRulesCoreExt.jl:73-74` — `tu = adjoint(sol.u)`; `∂A = .-(λ .* tu)`
- `ext/LinearSolveMooncakeExt.jl:113-114`
- `ext/LinearSolveEnzymeExt.jl:849` — `_sparse_outer_sub!(dA, z, y)`

against `ext/LinearSolveForwardDiffExt.jl:87-146`, whose `!issquare(A)` branch already
derives the extra term.

## Reproduction

LinearSolve 5.17.4 (`main` at eb0bd76), Julia 1.13.0. Reference is `ForwardDiff` on a plain
`A \ b`, cross-checked against `FiniteDiff`.

```julia
using LinearSolve, LinearAlgebra, Zygote, FiniteDiff, ForwardDiff, Random

Random.seed!(9)
m, n = 7, 4
A0 = randn(m, n); b0 = randn(m)
fref(Av) = sum(reshape(Av, m, n) \ b0)
gfd = ForwardDiff.gradient(fref, vec(A0))

for alg in (QRFactorization(), SVDFactorization())
    g(Av) = sum(solve(LinearProblem(reshape(Av, m, n), b0), alg).u)
    @show norm(Zygote.gradient(g, vec(A0))[1] - gfd) / norm(gfd)
    @show norm(ForwardDiff.gradient(g, vec(A0))  - gfd) / norm(gfd)
end

dx = ones(n); x = A0 \ b0; r = b0 - A0 * x; lam = A0' \ dx
@show norm(vec(-lam * x')                          - gfd) / norm(gfd)
@show norm(vec(-lam * x' + r * ((A0'A0) \ dx)')    - gfd) / norm(gfd)
```

```
FiniteDiff vs ForwardDiff on the reference: relerr=1.774e-10
QRFactorization    Zygote relerr=0.3289      ForwardDiff relerr=8.556e-16
SVDFactorization   Zygote relerr=0.3289      ForwardDiff relerr=1.34e-15
naive -lam*x'  relerr=0.3289
corrected      relerr=8.555e-16
```

The naive formula reproduces Zygote's error exactly, and adding the residual term closes it
to machine precision. Enzyme and Mooncake give the same 0.3289.

`test/AD/` has no reverse-mode gradient check against a non-square `A`.

## A second, distinct defect in the same area

`defaultalg_adjoint_eval`'s `NormalCholeskyFactorization` branch (`src/default.jl:1290`)
evaluates `cacheval' \ dy`. The cached object is `cholesky(AᴴA)`, not a factorization of
`A`, so that yields `(AᴴA)⁻¹∂x` where the adjoint solve needs `A (AᴴA)⁻¹∂x` — the wrong
value and the wrong length. `src/adjoint_factorization.jl:181-186` gets this right for the
explicit algorithm via `_NormalAdjointFactorizationReuse`; the default-solver dispatch lumps
NormalCholesky in with the plain-factorization group.

`defaultalg` selects it exactly when `!issquare(A) && condition === WellConditioned`:

```julia
using LinearSolve, LinearAlgebra, Random
Random.seed!(11); A0 = randn(7,4); b0 = randn(7); dy = randn(4)
as = LinearSolve.OperatorAssumptions(false;
        condition = LinearSolve.OperatorCondition.WellConditioned)
c = init(LinearProblem(A0, b0); assumptions = as); solve!(c)
length(LinearSolve._adjoint_solve(c, dy))
```

```
default solver choice : NormalCholeskyFactorization
default  _adjoint_solve length=4 (correct 7)
explicit _adjoint_solve length=7  relerr=3.719e-16
```

## Suggested fix

For the pullback, add the residual term when `A` is not square. `r` and `x` are both to
hand, and the `(AᴴA)⁻¹` apply can reuse the same factorization the forward solve built for
the least-squares algorithms. Since the expression is identical in all three extensions, a
shared helper in `src/adjoint_factorization.jl` alongside `_adjoint_solve` would keep them
from drifting apart again.

For the NormalCholesky branch, route it to the `A * (factorization \ dy)` form that
`_NormalAdjointFactorizationReuse` already implements rather than the plain
`factorization' \ dy` group.

Underdetermined (`m < n`) systems are wrong in both modes, since the ForwardDiff extension
only derives the overdetermined formula — worth deciding whether to fix or to reject at
`init`.

AJ0070:
## Checklist

## Additional context

Fixes #1309. All three reverse rules computed `∂A` with the square-system formula `-λ xᴴ`. A non-square solve returns `x = A⁺b`, whose pullback carries a second term, so `∂b` was right and `∂A` was wrong by roughly 90% with nothing to indicate it. The ForwardDiff extension already derives the overdetermined form, so forward and reverse disagreed with each other.

The term differs by orientation, and for a full-rank `A` only one of them is ever nonzero: the residual `r = b - A x` drives it when `m > n`, and the null space of `A` when `m < n`. Both are now in one helper next to `_adjoint_solve` so the three extensions cannot drift apart again. They are written as solves with `A` rather than through `AᴴA`/`AAᴴ`, which keeps the normal equations and their squared condition number out of it; the two forms were checked against each other and agree to machine precision on tall and wide shapes.

The underdetermined branch was not derived by hand. Candidate expressions were fitted against a ForwardDiff reference until one matched to 8.9e-16, then confirmed on further shapes.

A non-square `A` is now copied unconditionally in the ChainRules rule. The existing preservation skips the copy exactly when the factorization can be reused, which left `A_` aliasing the matrix `qr!` overwrites, and the correction solves with `A` itself. Mooncake already restores pristine operands after the solve, so it needs no copy; Enzyme takes one above the primal solve, so square problems keep paying nothing, and the accumulation is guarded to dense `dA` so the sparsity-preserving path is untouched.

`defaultalg_adjoint_eval` also had `NormalCholeskyFactorization` in the group that evaluates `F' \ dy`. Its cacheval is `cholesky(AᴴA)`, not a factorization of `A`, so that returned the wrong value and the wrong length; it now uses the `A * (F \ dy)` form that `_NormalAdjointFactorizationReuse` already applies to the explicit algorithm, and the two now agree to 3.7e-16.

Checked on a 7x4 system against a ForwardDiff reference cross-checked with FiniteDiff: Zygote, Enzyme and Mooncake all move from about 0.93 relative error to 7.785e-16, and the square path is unchanged at 1e-16. Two cases stay wrong and are unchanged from `main`: a wide `NormalCholeskyFactorization`, which does not return the minimum-norm solution so no pullback formula applies, and wide QR and the wide default, which throw `DimensionMismatch` in the forward solve. Core and AD both pass, and the new testset fails on the unpatched source.

AI Disclosure: Used Opus 5


ext/LinearSolveEnzymeExt.jl
```diff
@@ -847,6 +853,10 @@ function EnzymeRules.reverse(
 
         # Use sparse-safe outer product subtraction to preserve sparsity pattern
         _sparse_outer_sub!(dA, z, y)
+        if A_ns !== nothing && dA isa StridedMatrix
+            extra = LinearSolve._nonsquare_pullback_term(A_ns, b_ns, y, z, dy)
+            extra === nothing || (dA .+= extra)
```
ChrisRackauckas: not a necessary allocation.

AJ0070: resolved it.

PR #1316, opened 2026-09-14
Re-derive the dual cache's primal A and partials when isfresh is set

### LINKED ISSUE #1315: ForwardDiff extension: in-place updates to `cache.A` leave the primal matrix and partials stale, so `isfresh` solves the previous system
## Description

`isfresh` is documented as `false => cacheval is set wrt A, true => update cacheval wrt A`
(`src/common.jl:288`), and for a plain `LinearCache` that is enough: filling `cache.A` in
place and setting `cache.isfresh = true` refactorizes and gives the right answer. A
`DualLinearCache` silently diverges, because it derives two further pieces of state from
`A` at assignment time and neither is reachable that way.

`setA!` (`ext/LinearSolveForwardDiffExt.jl:676`) maintains three things:

```julia
prop = nodual_value!(getproperty(dc.linear_cache, :A), A)  # primal copy, a separate array
setproperty!(dc.linear_cache, :A, prop)
setfield!(dc, :dual_A, A)
partial_vals!(getfield(dc, :partials_A), A)                # only when A carries Duals
setfield!(dc, :A_partials_valid, false)
```

`getproperty(dc, :A)` returns `dc.dual_A` (`:785`), so `copyto!(cache.A, A2)` reaches only
the dual matrix. The primal array that actually gets factorized is never updated, and
`A_partials_valid` stays `true`, so the partials of the first solve are reused at `:301`
and `:327`.

The result is not a stale factorization of the right matrix, it is the solution of a
different system, and both the values and the partials are wrong. Nothing warns.

## Repro

```julia
using LinearSolve, ForwardDiff, LinearAlgebra

const T = ForwardDiff.Tag{typeof(identity), Float64}
const DU = ForwardDiff.Dual{T, Float64, 1}
d(v, p) = DU(v, ForwardDiff.Partials((p,)))

A1 = [d(2.0, 1.0) d(1.0, 0.0); d(1.0, 0.0) d(3.0, 0.0)]
A2 = [d(5.0, 7.0) d(1.0, 0.0); d(1.0, 0.0) d(4.0, 0.0)]
b  = [d(1.0, 0.0), d(2.0, 0.0)]

c = init(LinearProblem(copy(A1), copy(b)))
copyto!(c.A, A1); c.isfresh = true; solve!(c)
copyto!(c.A, A2); c.isfresh = true
u = solve!(c).u
```

```
copyto!(cache.A, A2) + isfresh    value = [0.111111, 0.777778]   partials = [-0.061728, 0.012346]
cache.A = A2                      value = [0.105263, 0.473684]   partials = [-0.155125, 0.038781]
A2 \ b                            value = [0.105263, 0.473684]   partials = [-0.155125, 0.038781]
```

Inspecting the cache after the in-place update shows why:

```
after copyto!(c.A, A2):
  dual_A           = [5.0 1.0; 1.0 4.0]   <- updated
  inner primal A   = [2.0 1.0; 0.5 2.5]   <- not updated, still the LU factors of A1
  A_partials_valid = true                 <- partials never recomputed
```

The same pattern on a `Float64` cache is correct, which is what makes it a trap:

```julia
F1 = [2.0 1.0; 1.0 3.0]; F2 = [5.0 1.0; 1.0 4.0]; fb = [1.0, 2.0]
c3 = init(LinearProblem(copy(F1), copy(fb)))
copyto!(c3.A, F1); c3.isfresh = true; solve!(c3)
copyto!(c3.A, F2); c3.isfresh = true
solve!(c3).u   # [0.105263, 0.473684], matches F2 \ fb
```

## Suggestion

Assignment is the documented update path and it works, so this may be a documentation
matter rather than a code one. But the failure is silent, the base cache makes the pattern
look supported, and `isfresh`'s own comment invites it. Either having `isfresh = true` also
clear `A_partials_valid` and refresh the primal copy, or documenting on `DualLinearCache`
that `cache.A` must be assigned rather than written through, would close it.

Found while reusing a cache across solves under `ForwardDiff` in
JuliaComputing/StateSelection.jl#160, where it silently produced wrong Jacobians on every
call after the first.

Related, in that they are all about state derived from `A` outliving the `A` it came from:
 #1306, #1313.

## Versions

LinearSolve 5.17.5, ForwardDiff 1.4.6, Julia 1.13.0, macOS aarch64.

AJ0070:
## Checklist

## Additional context

Fixes #1315. `setA!` maintains three things from `A`: the primal copy the solve actually factorizes, the extracted partials, and `dual_A`. `getproperty(dc, :A)` returns `dual_A`, so filling `cache.A` in place reaches only the last of them, and `cache.isfresh = true` forwarded straight to the inner cache without touching the other two. The solve then factorized the previous matrix and reused the previous partials, so both the values and the derivatives were wrong with nothing to indicate it.

Setting `isfresh = true` now re-derives from whatever `dual_A` currently holds, by calling `setA!` with it. That reuses the existing bookkeeping rather than restating it, so the primal copy, the partials and `A_partials_valid` stay consistent by construction.

The only cost lands on `cache.A = A2; cache.isfresh = true`, where `setA!` now runs twice, measured at about 20% of a refactorize and solve at n = 120. That idiom is already redundant, since the assignment sets `isfresh` itself; writing through `cache.A` and then setting `isfresh`, which is the case this fixes, is unchanged.

Checked against `A2 \ b` for the default, `LUFactorization`, `GenericLUFactorization` (the direct-dual path), `QRFactorization` and `KrylovJL_GMRES`, on values and partials. The new testset fails on the unpatched source on both. Core and AD both pass.

One nearby case is deliberately left alone: `b` has the same split between `dual_b` and the primal copy, so `copyto!(cache.b, b2)` is stale in the same way, but it has no freshness flag to hang this on and would need re-deriving on every solve instead. Different enough in shape that it wants its own issue.

AI Disclosure: Used Opus 5


ext/LinearSolveForwardDiffExt.jl
```diff
@@ -773,6 +773,15 @@ function Base.setproperty!(dc::DualLinearCache, sym::Symbol, val)
         setu!(dc, val)
     elseif sym === :p
         setproperty!(dc.linear_cache, :p, nodual_value(val))
+    elseif sym === :isfresh
+        # `cache.isfresh = true` is the documented way to say that `A` changed, and on a
+        # plain `LinearCache` filling `cache.A` in place is enough for that to be true.
+        # Here `cache.A` is `dual_A`, while the solve factorizes a separate primal copy and
+        # reads cached partials, so both have to be re-derived from whatever `dual_A` now
+        # holds. `setA!` with the current `dual_A` does exactly that.
```
ChrisRackauckas, line 781: this seems wrong, just set isfresh correctly.

PR #1330, opened 2026-09-20
Refactor PureKLU with the pivots reused unless growth says otherwise

ChrisRackauckas-Claude:

Depends on PureKLU v1.6.0 (SciML/PureKLU.jl#87, registered as JuliaRegistries/General#168872), which adds the growth estimate this uses. Compat is bumped to `PureKLU = "1.6"`.

## What changed and why

#1301 made every `PureKLUFactorization` refactorization pass `reuse_pivots = false`, which is a full `klu_factor!` rather than a `klu_refactor!`. That fixed a real defect — reused pivots can go unstable on new values and return `ReturnCode.Success` with a wrong answer — but it pays for new pivots on *every* solve whether or not the old ones have actually gone bad, and it lands on the default sparse path, so it is a regression users hit without opting in.

Reported by a user benchmarking a chromatography Newton solve under ForwardDiff duals, who bisected it to 5.17.3 → 5.17.5 and identified `reuse_pivots = false` as the cause before filing. #1301 is the only change in that range that touches the sparse KLU path.

The reason it was unconditional is stated in #1301's own message: *"PureKLU exposes no pivot-growth estimate, so the refresh here is unconditional."* PureKLU 1.6 now exposes one, so this takes the same shape the SuiteSparse `KLUFactorization` path has had since #1299 — refactor with the pivots reused, then redo the factorization only when the status or the growth says they have become unstable.

## Cost

Block-banded operand (chromatography-like Newton matrix, `nc` species per node with upwind coupling), Julia 1.11:

| | refactor | + rgrowth | total | full factor (current `main`) |
|---|---|---|---|---|
| n=8000, nnz=47992 | 0.641 ms | 0.117 ms | **0.758 ms** | 1.547 ms |
| n=20000, nnz=119992 | 1.618 ms | 0.293 ms | **1.911 ms** | 3.894 ms |
| n=16000, nnz=159984 | 2.915 ms | 0.364 ms | **3.279 ms** | 5.881 ms |

End to end on the reporter's shape — 8 cached solves through one cache, `Dual` with 2 partials, so each solve is one factorization plus three backsolves:

```
n=8000   23.11 ms -> 17.45 ms
n=20000  59.14 ms -> 45.45 ms
```

For reference, `KLUFactorization()` (SuiteSparse, which already had the conditional re-pivot) measures 16.78 / 44.73 ms on the same problems — so this brings the pure-Julia path back alongside it.

## Verification

The new testset alone, without the `src/sparsearrays.jl` change:

```
PureKLU refactorization keeps pivots that are still stable | Pass  Fail  Total
                                                          |    3     1      4
ERROR: LoadError: Some tests did not pass: 3 passed, 1 failed, 0 errored, 0 broken.
```

With it: `4 4`. The discriminator is the pivot order itself — `1e-4` on the diagonal is below KLU's `tol`, so a full factorization moves the pivot from `[1, 2]` to `[2, 1]`, while the growth stays at `3.0e-4`, far above `sqrt(eps(Float64)) = 1.49e-8`. The existing `pureklu_refactor_accuracy.jl` testset (which is what #1301 added, and covers the case where pivots genuinely must be refreshed) still passes unchanged at 132/132.

Full Core group on this branch, Julia 1.11, resolving PureKLU v1.6.0 from the registry:

```
$ GROUP=Core julia --project -e 'using Pkg; Pkg.test()'
Test Summary:                    | Pass  Total     Time
PureKLU refactorization accuracy |  132    132  1m24.2s
... 40 testsets, no failures, no errors ...
```

Runic (`--check`) and `typos` both clean.

## Not verified

- **The docs build did not run on this machine.** `LinearSolvePyAMG` fails to precompile from an unrelated CondaPkg/pixi problem (`Failed to update PyPI packages ... LookupError: no codec search functions registered`), which happens before Documenter starts and is independent of this diff. A local build with that one dependency dropped is reported separately in a comment below; CI's docs job covers the real thing. The edited page and docstring add prose only, with no new cross-references.
- GPU and downstream groups were not run locally.
- Only Julia 1.11 on x86_64-linux.

## For a reviewer

- The growth threshold `sqrt(eps(Float64))` is copied from the SuiteSparse branch rather than chosen here, so the two KLU paths agree on when a refactorization is untrustworthy. If that threshold is wrong it is wrong in both places.
- `PureKLU.rgrowth` is verified against SuiteSparse's `klu_rgrowth` in PureKLU's own test suite, including the row-scaled cases. This PR trusts that agreement rather than re-testing it.
- This is a behavior change for anyone who came to depend on 5.17.4/5.17.5 always re-pivoting. It is still the case that an unstable refactorization is refreshed; what changes is that a stable one is not.

🤖 Generated with [Claude Code](https://claude.com/claude-code) (model: claude-opus-5[1m])

https://claude.ai/code/session_01MfiS7BpgL82gQaHa5C5VaE


ChrisRackauckas-Claude:
Docs build result, as promised in the description.

The real `docs/make.jl` cannot run on the machine this was developed on: `LinearSolvePyAMG` fails to precompile because CondaPkg's `pixi` cannot start its bundled Python —

```
Error:   × Failed to update PyPI packages for environment 'default'
      LookupError: no codec search functions registered: can't find encoding
ERROR: LoadError: InitError: failed process: Process(`.../bin/pixi install --manifest-path docs/.CondaPkg/pixi.toml`)
```

That reproduces after deleting and rebuilding `docs/.CondaPkg` from scratch, and it happens during `Pkg.instantiate`, before Documenter is entered — so it is a local environment problem, not something in this diff.

To get coverage of the page and the docstring this PR edits, I ran the build in a scratch worktree with `LinearSolvePyAMG` dropped from `docs/Project.toml`, `docs/make.jl` and its `@docs` block in `solvers.md`. That build reaches rendering with no Documenter errors:

```
[ Info: SetupBuildDirectory: setting up build directory.
[ Info: Doctest: running doctests.
[ Info: ExpandTemplates: expanding markdown templates.
[ Info: CrossReferences: building cross-references.
[ Info: CheckDocument: running document checks.
[ Info: Populate: populating indices.
[ Info: RenderDocument: rendering document.
[ Info: HTMLWriter: rendering HTML pages.
```

The only remaining diagnostics are pre-existing and unrelated: `Metal.jl is only supported on Apple Silicon` (a load-time message on non-Apple hardware, not a Documenter error), a 301 redirect on the CKTSO.jl link, the `size_threshold_warn` on `solvers/solvers.md`, and the deployment skip.

The intermediate run — same worktree, PyAMG dropped from `modules` but its `@docs` block still in place — produced exactly four errors, all of them `unable to get the binding for 'LinearSolvePyAMG.…'`, i.e. only the ones I caused. Nothing came from `PureKLUFactorization`'s docstring or from `tutorials/caching_interface.md`.

None of these workaround edits are in the branch; CI's docs job builds the real thing.

Issue #1333, opened 2026-09-20
Wrapped sparse matrices are treated as dense, so transpose and adjoint fail at init


AJ0070:
## Description

Dispatch does not see through `transpose`, `adjoint`, `Symmetric` and `Hermitian` to the
sparse matrix underneath, so a wrapped sparse matrix is treated as dense.

`init` itself throws for `transpose` and `adjoint`, before any solve is attempted. The
default cache builds a column-pivoted QR slot for every problem, and the
`Adjoint`/`Transpose` method that serves it calls `qr(A, alg.pivot)`, which SPQR rejects
for a sparse matrix. `Symmetric` and `Hermitian` survive `init`, but every sparse
algorithm named explicitly on one fails, because their `init_cacheval` methods dispatch on
`AbstractSparseArray` and a wrapper is not one, so the cacheval is `nothing`.

Base solves all four of these.

## Where

- `src/factorization.jl:1063` matches `A::Union{<:Adjoint, <:Transpose}` with no
  constraint on the parent, and `src/factorization.jl:1068` then calls `qr(A, alg.pivot)`.
  Reached from `_init_default_cacheval` at `src/default.jl:646`.
- The sparse `init_cacheval` methods dispatch on `AbstractSparseArray{T, <:Integer}` or
  `SparseMatrixCSC`, for example `src/sparsearrays.jl:984` and `:992` for
  `SparseColumnPivotedQRFactorization`. A wrapper misses them and falls to the catch-all
  that returns `nothing`. KLU and UMFPACK then read `.nzval` off that `nothing`, and
  `SparseColumnPivotedQRFactorization` tries to store a factorization into a slot typed
  `Nothing`.
- `defaultalg` misses the same way: `transpose` of a sparse matrix is routed to
  `KrylovJL_GMRES`.

## Reproduction

LinearSolve 5.17.5, Julia 1.13.0. `A` is sparse and positive definite throughout, so
nothing here is a numerical failure.

`init` with the default algorithm:

```
  bare       -> ok
  Symmetric  -> ok
  Hermitian  -> ok
  transpose  -> THREW ErrorException: Pivoting Strategies are not supported for `SparseMatrixCSC`s
  adjoint    -> THREW ErrorException: Pivoting Strategies are not supported for `SparseMatrixCSC`s
```

A named sparse algorithm, eltype `Float64`:

```
  KLU             bare=ok Sym=FieldError Herm=FieldError transp=FieldError adj=FieldError
[...]

PR #1334, opened 2026-09-20
Give adjoint and transpose of sparse matrices their own QR cacheval

AJ0070:
## Checklist

## Additional context

Part of #1333, and not a full fix for it, so I have left it without a closing keyword.

`init(LinearProblem(transpose(A), b))` threw for a sparse `A`, before any solve was attempted. The default cache builds a column-pivoted QR slot for every problem, and the `init_cacheval` that serves `Union{<:Adjoint, <:Transpose}` puts no constraint on the parent, so a wrapped sparse matrix took the dense path and asked SPQR for `qr(A, ColumnNorm())`, which it rejects. Base solves `transpose(A) \ b` fine.

The new method mirrors the `Symmetric{<:Number, <:SparseMatrixCSC}` one directly above it, which is there for the same reason. `transpose` and `adjoint` of a sparse matrix now reach `init` and solve.

What is left in #1333 is not a dispatch problem. The sparse `solve!` bodies read `nonzeros(A)`, which throws a `MethodError` on a wrapper, so naming KLU, UMFPACK or `SparseColumnPivotedQRFactorization` on a wrapped sparse matrix still fails, and routing `transpose` to a sparse direct solver would move the failure from `init` into `solve` rather than fix it. Supporting wrappers there means materializing a `SparseMatrixCSC`, which is a memory decision rather than a signature change. For now `transpose` goes through `KrylovJL_GMRES`, correct but at 3.95e-9 against 3.62e-16 for a direct solve.

AI Disclosure: Used Opus 5


ChrisRackauckas:
conflicts

PR #1338, opened 2026-09-21
Bump version

ErikQQY:
Release the new version for AlmostBandedMatrices.jl boundary block dense storage. Should release https://github.com/SciML/FastAlmostBandedMatrices.jl/pull/95 first. @oscardssmith


oscardssmith:
> Should release https://github.com/SciML/FastAlmostBandedMatrices.jl/pull/95 first

why? Does this need a higher lower bound of FastAlmostBandedMatrices?

ErikQQY:
No, but the fix only implemented in the latest releases of FastAlmostBandedMatrices.jl

oscardssmith:
why do we need a bump here then?

